PR #38, opened 2018-02-19
Reading multiple filters per observation and parsing commissioning APT templates

laurenmarietta:


bhilbert4:
Still have a few diffs to go through. Two small things I've noticed so far:

yaml_generator.py, line 555, and apt_inputs.py, line 370, you left in the name of your local file.

bhilbert4:
Line 233 in obs_generator.py, you added the line:
print("Observation generation complete.")
It might be helpful to add "Linearized" at the start of that line, just because for the case where the linearized and raw outputs are created, the raw output is created after the linearized version, so it's a little misleading to say that the observation generation is complete and then move on to create the raw version.

Other than that, everything looks great!

laurenmarietta:
I updated the filepaths to the subarray definition files, and moved the "Observation generation complete." notice to after the raw outputs are created!

bhilbert4:
Looks good! I'm going to merge into the mult-filters-per-observation branch to allow for further testing.

Issue #67, opened 2018-03-27
Uniformize usage of lower/uppercase instrument names in configuration files


Johannes-Sahlmann:
For instance, these two files have inconsistent nomenclature:
config/NIRCam_subarray_definitions.list 
config/niriss_subarrays.list

This may partially be superseded when the dependence on pysiaf will be introduced.

PR #69, opened 2018-03-28
No distortion mode

bhilbert4:
I think I have the distortionless mode working. My first test for a series of point sources separated by a constant RA across the detector produced data with the sources separated by a constant number of pixels. I still want to test using a non-zero telescope roll, to make sure that my fake WCS is correct.

To get distortionless output, set the distortion reference file input field to None. This brings up a related topic: the names in the input yaml file are confusing. Currently the `astrometric` entry contains the name of the distortion reference file (and should be set to None for distortionless mode), while the `distortion_coeffs` entry contains the name of the SIAF, which is still needed. These names should be changed, but this will break earlier versions of the code, and necessitate changes in the simulator code and yaml generator function. Maybe these changes should be saved for a separate pull request, in order to get distortionless mode implemented in the `master` branch now?

### DIFF
--- a/nircam_simulator/scripts/catalog_seed_image.py
+++ b/nircam_simulator/scripts/catalog_seed_image.py
@@ -16,7 +16,7 @@ import math
 from photutils import detect_threshold, detect_sources
 from astropy.io import fits, ascii
 from astropy.table import Table
-from astropy.modeling.models import Sersic2D
+from astropy.modeling.models import Shift, Sersic2D, Polynomial2D, Mapping
 from astropy.convolution import convolve
 from asdf import AsdfFile
 import yaml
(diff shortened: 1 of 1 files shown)


Johannes-Sahlmann:
It is true that the naming is a little confusing but I suggest to hold off breaking changes for 'cosmetic' reasons until a later time. There are other similar issues (e.g. #67, and other items that are coming as we implement NIRISS and FGS support) which we could possible clean up alltogether.

bhilbert4:
Makes sense to me. I'll open a separate issue about the yaml entry names. My distortionless test using a non-zero telescope roll looks ok as well. I think this is now ok to merge.

PR #72, opened 2018-03-28
WIP: Apply PAM to all sources except point sources

bhilbert4:
Code is in place such that the pixel area map is applied only to extended targets and background signals. What needs to happen for point sources? Is there a different correction for those? @KevinVolkSTScI is probably the best person to answer this.

Note that this branch needs to be rebased before it is merged.

### DIFF
--- a/nircam_simulator/scripts/catalog_seed_image.py
+++ b/nircam_simulator/scripts/catalog_seed_image.py
@@ -131,6 +131,10 @@ class Catalog_seed():
         # calculate the exposure time of a single frame, based on the size of the subarray
         self.calcFrameTime()
 
+        # Read in the pixel area map, which will be needed for certain
+        # sources in the seed image
+        self.prepare_PAM()
+        
         # For imaging mode, generate the countrate image using the catalogs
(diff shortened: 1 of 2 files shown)


KevinVolkSTScI:
For point sources my opinion is that we should just apply the WebbPSF image directly with no corrections.  If the physical pixel area on the detector is varying to create a distortion of the images then this would be the wrong thing to do, but if the distortions are in the optics rather than on the detector the pixel area effect should not change the point source PSF because all the photons are coming from one point.  For NIRISS all the distortion values are derived from raytracing and therefore only deal with the optics, and there is no contribution from the physical pixel sizes which are assumed to be perfectly square and uniform across the array.

bhilbert4:
Thanks. I believe the NIRCam distortion model is similar to NIRISS in that it comes from the optics and does not include variable pixel areas. I'll run a couple of test cases using this branch to make sure the outputs make sense, and then merge.

bhilbert4:
Test results look ok. Pixel area map applied to all sources/backgrounds other than point sources. Merging.

Issue #73, opened 2018-03-28
Implement APT interface for NIRISS


Johannes-Sahlmann:
Enhance the yaml_generator script (and dependents) to convert the content of an NIRISS APT file into simulator input files.

bhilbert4:
@Johannes-Sahlmann, I just tried running the `yaml_generator` on a NIRISS APT file and it crashed. It looks like the only NIRISS APT template that or works for is `NIRISSExternalCalibration`. Do I need to do something special in APT to access this template? Or is it out of date? All I see in APT now are templates for Imaging, WFSS, SOSS, and AMI.

Johannes-Sahlmann:
I think I only tested this for a `NIRISSExternalCalibration` APT program. You have to set the proposal category to ENG to have access to it.

bhilbert4:
Ok thanks. I was able to get to the `NIRISSExternalCalibration` template. It looks to be more complicated than I need for my test. At some point we should add at least the `Imaging` template, although when that happens depends I guess on how badly people need it. I'm just running a single simple test of full frame vs subarray, so maybe that's not enough of a driver to make this a high priority.

Issue #64, opened 2018-03-26
Develop unit tests for different APT files/templates


laurenmarietta:
Create one APT file for each instrument that can be simulated (for now, NIRCam and NIRISS), with all necessary templates in each file.

Ultimately, compare `.yaml` output with some kind of reference `.yaml`.

Use the STScI package template to model test structure for continuous integration:
https://github.com/spacetelescope/stsci-package-template

Johannes-Sahlmann:
Functional example for the template's CI via Travis: https://github.com/spacetelescope/nrm_analysis

laurenmarietta:
Though unit testing does still need to be more robustly developed, this particular system test for APT templates has been developed.

PR #74, opened 2018-03-28
NIRISS APT interface

Johannes-Sahlmann:
This is the first functional version that supports processing a basic NIRISS APT file (#73).
This needs careful review because many changes were made and I did not have an easy way to verify that they don't break the use for NIRCam.

Because NIRISS and NIRCam reference files are located in different locations and have different names, changes to the SimInput __init__ method were made, which imply that mirage can only work on the observations of one instrument at the time (instrument can be 'WFSC', 'NIRCam', ' NIRISS').

If needed I can make the example driver script and the APT file available (eventually we want to have this as a test case).

### DIFF
--- a/nircam_simulator/scripts/catalog_seed_image.py
+++ b/nircam_simulator/scripts/catalog_seed_image.py
@@ -1637,6 +1637,10 @@ class Catalog_seed():
             if bstr == "0.0":
                 bstr = "0.00"
 
+            if self.params['simSignals']['psfbasename'].lower() == 'niriss':
+                astr = '{:1.2f}'.format(np.float(astr))
+                bstr = '{:1.2f}'.format(np.float(bstr))
+
             #generate the psf file name based on the center of the point source
             #in units of fraction of a pixel
             frag = astr + '_' + bstr
@@ -1659,8 +1663,7 @@ class Catalog_seed():
                     else:
                         raise FileNotFoundError("PSF file {} not found.".format(psffn))
                 except:
-                    print("ERROR: Could not load PSF file {} from library".format(psffn))
-                    sys.exit()
+                    raise RuntimeError("ERROR: Could not load PSF file {} from library".format(psffn))
 
             # Normalize the total signal in the PSF as read in
             totalsignal = np.sum(webbpsfimage)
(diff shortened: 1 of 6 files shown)


bhilbert4:
I think it's good to get NIRISS working here, and then later we can think about APT files that include multiple instruments. For now, if there are multiple instruments, then all inputs should be generated by running the APT file through the code once for each instrument, right? 

I can try running a NIRCam APT file using this branch and make sure the outputs look ok. I don't think I've ever generated an APT file with prime and parallel instruments, but assuming I can do that, I can test that case as well.

Johannes-Sahlmann:
The answer to your question is yes.

It would be great if you test this branch in the way you suggest, which bridges the gap to the tests that Lauren is setting up (#64)

Johannes-Sahlmann:
Yes, please send me that NIRCam APT file and I will fix that.

Johannes-Sahlmann:
The NIRCam example files that Bryan gave me now pass successfully. Note that the fix does not yet include any NIRCam template other than the basic imaging one. I deferred additional changes until we decide if/how to restructure read_apt_xml. 

@laurenmarietta if you want to keep the current structure of read_apt_xml for the time being, I am happy to implement the fix in all the read_template functions. However, this will be a blind change, as I don't have a way to test.

Johannes-Sahlmann:
Even if currently only very basic modes of NIRISS are supported, I think this is ready to merge (from my side). 
More advanced and complex use cases will be much easier to develop in new branches ones we have the test infrastructure set up.

nircam_simulator/scripts/catalog_seed_image.py
```diff
@@ -1637,6 +1637,10 @@ def makePointSourceImage(self, pointSources):
             if bstr == "0.0":
                 bstr = "0.00"
 
+            if self.params['simSignals']['psfbasename'].lower() == 'niriss':
```
bhilbert4: I think you may need to rebase this from `master`. I merged some changes on Thursday that solved the PSF filename issue, such that two digits are always kept from the sub-pixel position.

bhilbert4: Rebasing may also fix the conflicts below.

Johannes-Sahlmann:
Rebasing seemed overkill given the minor merge conflict. I resolved it by using the master branch code. I believe your changes to the master also solved the problem with identifying the WebbPSF library file for NIRISS. I will test this after the merge.

PR #81, opened 2018-04-02
Unit Test Infrastructure & NIRCam APT Test

laurenmarietta:
- Added `tests/` directory
- Wrote & provided files for testing NIRCam APT templates
    - Wrote preliminary `get_catalog.py` script that queries WISE and 2MASS for lists of NIRCam F212N and F480M sources
    - Created SIAF config json file that contains the directory for the NIRCam SIAF, and wrote utils function to read it
- Implemented fixes that resulted from running the test


tests/test_apt.py
```diff
@@ -0,0 +1,57 @@
...
+from nircam_simulator.scripts import write_observationlist, yaml_generator, utils
+from nircam_simulator.scripts.get_catalog import get_all_catalogs
+
+INSTRUMENTS = ['NIRCam']
+PROPOSAL_ID = '1111'
+
+TESTS_DIR = os.path.realpath(os.path.join(os.getcwd(), os.path.dirname(__file__)))
+
+
+def test_APT():
```
Johannes-Sahlmann: Can you please modify the name of the test function to be a bit more descriptive?

This test is missing an assert statement. I am actually not sure what happens if you run pytest on this?
Please add appropriate assert statements.

tests/test_apt.py
```diff
@@ -0,0 +1,57 @@
...
+    for instrument in INSTRUMENTS:
+        # Define .pointing and .xml file locations
+        pointing_file = os.path.join(TESTS_DIR, 'test_data', instrument + 'Test.pointing')
+        xml_file = os.path.join(TESTS_DIR, 'test_data', instrument + 'Test.xml')
+
+        # Point to appropriate output directory
+        out_dir = os.path.join(TESTS_DIR, 'test_data', 'APT_{}_out'.format(instrument))
+
+        # Create/locate catalogs for target(s)
+        sw_cats, lw_cats = get_all_catalogs(pointing_file, PROPOSAL_ID)
```
Johannes-Sahlmann: Are the catalogs already present? If not this will try to access vizier which may fail if the machine running the test cannot access vizier. 
Tests should be as basic as possible, so here I would suggest to create the catalogs in advance and include them in the test_data directory.
You could write an independent test for the automatic catalog generation, but I am not sure how/if this will work once we run this on Travis or similar.

tests/test_apt.py
```diff
@@ -0,0 +1,57 @@
...
+        yam.pointing_file = pointing_file
+        yam.siaf = utils.get_siaf()['NIRCam']
+        yam.output_dir = out_dir
+        yam.simdata_output_dir = out_dir
+        yam.observation_table = observationlist_file
+        yam.use_JWST_pipeline = False
+        yam.use_linearized_darks = True
+        yam.datatype = 'linear'
+        yam.reffile_setup()
+        yam.create_inputs()
```
Johannes-Sahlmann: Since this creates a lot of temporary files that are only used for the test, one clean solution is to use the pytest yield function, which allows you to automatically clean up temporary files afterwards. This is just a suggestion and can find an example in the tests I wrote for jwql.permissions (https://github.com/spacetelescope/jwql/blob/master/jwql/tests/test_permissions.py).

nircam_simulator/scripts/get_catalog.py
```diff
@@ -0,0 +1,219 @@
...
+WISE for short-wave and long-wave catalogs, respectively, and writes
+new .list files.
+
+Authors
+-------
+    - Lauren Chambers
+
+Use
+---
+    This module can be executed in a Python shell as such:
```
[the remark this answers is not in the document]
laurenmarietta: I agree; I think having all of the python files in one "scripts" directory has gotten somewhat unwieldy as the tool has grown in complexity. I'm not particularly sure what modules make sense to create, though ('tools' sounds like a good one). I would agree that this particular file more accurately belongs in the `catalogs/` sub-directory, so I'll move it!

nircam_simulator/scripts/get_catalog.py
```diff
@@ -0,0 +1,219 @@
...
+            continue
+
+        # If not, query shortwave sources from the 2MASS catalog from Vizier
+        v = Vizier(catalog='II/246/out', columns=['_RAJ2000', '_DEJ2000', 'Kmag'])
+        v.ROW_LIMIT = -1
+        result = v.query_region(t, radius="1deg")
+        queried_catalog_sw = result['II/246/out']
+
+        print('Queried {} 2MASS objects within {} of RA, Dec ({:.2f}, {:.2f}).'.
+              format(len(queried_catalog_sw),'1 degree', t.ra.deg, t.dec.deg))
```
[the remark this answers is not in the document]
laurenmarietta: Done.

nircam_simulator/scripts/get_catalog.py
```diff
@@ -0,0 +1,219 @@
...
+        asc.write(queried_catalog_lw, output=catalog_filename_lw, overwrite=True)
+        print('Successfully saved longwave catalog to {}.'.format(catalog_filename_lw))
+
+        catalog_filenames_lw.append(catalog_filename_lw)
+
+    return catalog_filenames_lw
+
+
+def get_all_catalogs(pointing_file, prop_id):
+    target_coords = get_target_coords(pointing_file, prop_id)
```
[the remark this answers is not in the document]
laurenmarietta: Whoops, missed this one. Done!

nircam_simulator/scripts/get_psf.py
```diff
@@ -0,0 +1,4 @@
+import webbpsf
+
+def create_commissioning_psf(params):
```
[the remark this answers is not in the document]
laurenmarietta: Added the docstring; I will add an issue after I talk more with Marshall about what he envisions this function will look like.

tests/test_apt.py
laurenmarietta: Up until this point, I've really just been running pytest to ensure that all of the code was executed without throwing any errors. I'm not entirely sure what kinds of assert statements are useful here (please let me know if you have any ideas?).

For now at least, I added in a case that will compare the observationlist.yaml file that is created by the simulator to a reference file, if one is present in the output directory.

tests/test_apt.py
laurenmarietta: Ah, I hadn't thought about machines that can't get to Vizier. I'll modify this to just pull the catalogs from the test_data directory, and I'll make an issue for writing a test for catalog creation.

tests/test_apt.py
laurenmarietta: That sounds like a great idea, and is definitely something I would like to implement. However, from looking over your example and trying to find other examples online, I am kind of confused as to how I would do so in this case. (Decorators make things so much harder to understand, and I haven't been able to find a helpful explanation of what `yield` does...) If you're up for it, I would be happy to meet at some point and go over how this could be implemented?

laurenmarietta:
I think I addressed all of your comments; except when it comes to the concerns about placing modules in the scripts directory, I thought it might be better to have a more intentional discussion about that. 

From what I know, none of the files in the "scripts" directory are really being used as scripts, anyway (at least I am not running anything from the command line). So we might want to consider re-naming that whole folder to "modules"?

Or maybe re-organizing the files currently in "scripts" into sub-packages? I'm not too sure what the logical breakdown for a sub-package structure would be here, but perhaps having one for each step of the simulator (creating the seed image, generating the dark, and then generating the observation and output yaml files) might be a good place to start. I'm curious to hear what you think.

Issue #90, opened 2018-05-14
The "stmag" option in the source catalogue files does not seem to be working


KevinVolkAlt:
Testing of the "stmag" option for a point source list showed that this option did not work for me on the witserv3 machine.  I found that the condition used to test whether to switch from abmag to stmag was never returning True.  In fixing this I also added an option to use the "vegamag" string to enable use of the vegamag column in the NIRCam_zeropoints.list or niriss_zeropoints.list file.  Hence the addition of the normal magnitudes should work for both instruments.  I have only tested the change with NIRISS.

KevinVolkAlt:
I posted the code in branch magnitude-correction (I hope...it seemed to work from my computer).  The branch also includes the fix for the frame time.

KevinVolkSTScI:
I believe that this is fixed now and so I am closing this issue.

Issue #89, opened 2018-05-14
Frame time calculation gives the wrong values for NIRISS/Guider


KevinVolkSTScI:
Testing of the simulation code showed that the frame time calculated in the code for NIRISS is not correct.  The full frame time was given as 10.742 seconds.  This is off by 0.05% but the discrepancy was worse for subarrays such as SUB80.
   To fix this the calcFrameTime subroutine needs to be revised.  This routine is in both obs_generator.py and catalog_seed_image.py.  It is not clear why there are two versions of the same code differing only in a print statement in the routine in obs_generator.py.  It would be better to have only one calcFrameTime routine.
  The fix leaves the NIRCam frame time calculation as it was.  I note however that the JDox page ttps://jwst-docs.stsci.edu/display/JTI/NIRCam+Detector+Readout gives the same frame time for NIRCam as for NIRISS in full frame read-out, 10.73677 seconds.
  I have coded a fix and tested it for a few NIRISS sub-arrays and the values are as expected.  I have not been able to upload the changes (I get an error about my username or password).  Since I am using the password that works in the browser to get this issue up, I do not know what the problem is with pulling the changes back.

bhilbert4:
When I originally wrote the frame time correction functions, they followed the exposure time given on that page. Seems that number must have been updated since that point. NIRCam subarray exposure times follow the equations given on the [detector subarray page](https://jwst-docs.stsci.edu/display/JTI/NIRCam+Detector+Readout+Patterns).

Originally the simulator was split into three separate repos: seed image generator, dark prep, and observation generator. In that case each had it's own frame time calculation. When the three were combined into a single repo, I never got around to removing repeated functions. This is on my to do list. I think the best solution is to make a `utils` directory that contains functions needed in multiple places.

PR #92, opened 2018-05-15
Magnitude correction

KevinVolkSTScI:
This version has the frame time correction and the stmag/vegamag changes.
This includes issues #89, #90, and #92.


KevinVolkSTScI:
This pull request needs to be closed because I had to resync my local copy of the master to the up to date version.  Pull request #94 puts in the same changes as were originally in this request.

PR #94, opened 2018-05-15
WIP: Changes to (a) make the "stmag" option work and (b) add the "vegamag"…

KevinVolkSTScI:
… option.  This is for issue #90.  The change should work for NIRCam and NIRISS as the vegamag option uses the VEGAMAG column in the zeropoints.list files.  The FGS zero points need to be added as I do not see such a file in the config directory.  I have the values for both FGS detectors.

### DIFF
--- a/nircam_simulator/scripts/catalog_seed_image.py
+++ b/nircam_simulator/scripts/catalog_seed_image.py
@@ -387,6 +387,12 @@ class Catalog_seed():
                 #sys.exit()
                 raise ValueError(("AB mag to countrate conversion failed."
                                   "magnitude = {}, photfnu = {}".format(mag, photfnu)))
+        if magsys.lower() == 'vegamag':
+            try:
+                return 10**((self.vegazeropoint - mag) / 2.5) 
+            except:
+                raise ValueError(("Vega mag to countrate conversion failed."
+                                  "magnitude = {}".format(mag)))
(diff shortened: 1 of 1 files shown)


Johannes-Sahlmann:
@KevinVolkSTScI please include FGS magnitude zero point files.

KevinVolkSTScI:
In response to Bryan's comment, there is a distinction between vega magnitudes as used by HST and pysynphot wherein the Vega template spectrum is defined to have zero magnitude in all filters, and the traditional system where the average A0V star colours are zero between different filters.  Vega has V magnitude 0.03 so there is an offset for say HST F555W with respect to V from this magnitude discrepancy.  The issue is that people at large take the two systems to be the same when they are not.

KevinVolkSTScI:
This is approved, but I do not know how to get the changes into the master branch.  The workflow instructions have no description of that part of things.  It just says "once the pull request has been accepted and merged you can delete your local branch".  I need some help with this.

bhilbert4:
Can you see the green "Merge pull request" button below the comments and the "Changes approved" and "This branch has no conflicts with the base branch"? If it's not there or greyed out, maybe there is some permission setting that needs to be changed.

bhilbert4:
I sent an invite to collaborate. I think once you accept this, you should have permission to merge code.

Issue #91, opened 2018-05-14
When simulating a subarray the stars were not put in the right place


KevinVolkAlt:
The code does not generally locate stars correctly from a point source list when a subarray is being simulated, at least for NIRISS (but probably for NIRCam as well).  This was found to be due to having point source positions relative to the subarray in the code, whereas the RADectoXY routine produces return values of (x,y) for the pixel in the full frame coordinates.  One needs to subtract the self.subarray_bounds[0] and self.subarray_bounds[1] values to get pixel positions relative to the sub-array.  Related to this, the XYtoRADec routine needs input pixel values in the full frame but the code is passing values relative to the subarray so the RA/Dec output values are incorrect.  One in that case needs to add self.subarray_bounds[0] and self.subarray_bounds[1] to the pixel values that are being passed into XYtoRADec.  I have corrections for these problems ready to post when I get the account issues straightened out.

The fix needs to be put into catalog_seed_image.py in three places: for point sources, for moving targets, and for galaxies/extended sources.

bhilbert4:
Hmm, interesting. I know I've checked full frame vs subarray simulations before using the same RA, Dec values and had sources show up where I expected. Maybe I was checking some special case or something. One of the things I want to work on when I get back is to add a proper testing module to the repo that uses Jenkins to automatically run tests each time the code is updated.

KevinVolkAlt:
The only case I can think of that would work is a subarray where self.subarray_bounds[0] and self.subarray_bounds[1] are equal to zero.  It happens that none of the NIRISS subarrays are of that sort in the DMS coordinates, although several of them are that way in the detector coordinates.

Johannes-Sahlmann:
@bhilbert4 do you have a strong preference for Jenkins compared to Travis? I am using travis for https://github.com/spacetelescope/pysiaf and I could try to set it up for mirage.
Let me know.

bhilbert4:
I don't have any preference between Travis/Jenkins. I think output file sizes may be a problem for testing on github, although I'm not positive.

bhilbert4:
I just did a quick check for NIRISS and NIRCam. For NIRCam I made a full frame and a SUB160 exposure, both with RA, Dec set to (4., 45.), and both using a point source catalog with a single source at (4, 45). So I expect the source to show up at the reference location in each of the exposures, which it did.

[...]

PR #97, opened 2018-05-18
I have changed the code to add/subtract the subarray bound values as …

KevinVolkSTScI:
…needed to get sources into the right place when doing a subarray simulation.  The XYToRADec and RADecToXY routines are for pixel positions given with respect to the full frame whereas the code is assuming that pixelx, pixely are with respect to the subarray, hence whenever pixel positions are input for sources one has to use pixelx + self.subrray_bounds[0] and pixely + self.subarray_bounds[1] to get the correct RA/Dec strings, and when RA/Dec input positions are converted to pixelx, pixely positions using RADecToXY_astrometric (or RADecToXY_manual as the case may be) one has to subtract self.subarray_bounds[0] and and self.subarray_bounds[1] from pixelx and pixely respectively.  The changes have been made wherever I see the XYToRADec or RADecToXY routines being called.

### DIFF
--- a/nircam_simulator/scripts/catalog_seed_image.py
+++ b/nircam_simulator/scripts/catalog_seed_image.py
@@ -1113,6 +1113,8 @@ class Catalog_seed():
             else:
                 # No distortion at all - "manual mode"
                 pixelx, pixely = self.RADecToXY_manual(ra, dec)
+            pixelx = pixelx- self.subarray_bounds[0]
+            pixely = pixely- self.subarray_bounds[1]
 
         else:
             # Case where the point source list entry locations are given in units of pixels
@@ -1122,7 +1124,7 @@ class Catalog_seed():
             pixelx = entry0
             pixely = entry1
 
-            ra, dec, ra_str, dec_str = self.XYToRADec(pixelx, pixely, matrix, transform)
+            ra, dec, ra_str, dec_str = self.XYToRADec(pixelx+self.subarray_bounds[0], pixely+self.subarray_bounds[1], matrix, transform)
         return pixelx, pixely, ra, dec, ra_str, dec_str
 
     def nonsidereal_CRImage(self, file):
@@ -1619,6 +1621,8 @@ class Catalog_seed():
                     #else:
                     #    # No distortion at all - "manual mode"
                     #    pixelx, pixely = self.RADecToXY_manual(ra, dec)
+                    pixelx = pixelx-self.subarray_bounds[0]
+                    pixely = pixely-self.subarray_bounds[1]
 
                 else:
                     # Case where the point source list entry locations are given in units of pixels
@@ -1634,7 +1638,7 @@ class Catalog_seed():
                     pixelx = entry0
                     pixely = entry1
 
-                    ra, dec, ra_str, dec_str = self.XYToRADec(pixelx, pixely, attitude_matrix,
+                    ra, dec, ra_str, dec_str = self.XYToRADec(pixelx+self.subarray_bounds[0], pixely+self.subarray_bounds[1], attitude_matrix,
                                                               coord_transform)
 
                 # Get the input magnitude of the point source
@@ -2249,6 +2253,8 @@ class Catalog_seed():
                 # Call is the same regardless of whether distortion reference file
                 # is given or not
                 pixelx, pixely = self.RADecToXY_astrometric(ra, dec, attitude_matrix, coord_transform)
+                pixelx = pixelx - self.subarray_bounds[0]
+                pixely = pixely - self.subarray_bounds[1]
 
                 #else:
                 #    # No distortion. Fall back to "manual" calculations
@@ -2268,7 +2274,7 @@ class Catalog_seed():
                 pixelx = entry0
                 pixely = entry1
 
-                ra, dec, ra_str, dec_str = self.XYToRADec(pixelx, pixely, attitude_matrix, coord_transform)
+                ra, dec, ra_str, dec_str = self.XYToRADec(pixelx+self.subarray_bounds[0], pixely+self.subarray_bounds[1], attitude_matrix, coord_transform)
 
             # only keep the source if the peak will fall within the subarray
             if pixely > outminy and pixely < outmaxy and pixelx > outminx and pixelx < outmaxx:
@@ -2635,6 +2641,8 @@ class Catalog_seed():
                     #if self.runStep['astrometric']:
                     # Same function call regardless of whether distortion file is provided or not
                     pixelx, pixely = self.RADecToXY_astrometric(ra, dec, attitude_matrix, coord_transform)
+                    pixelx = pixelx - self.subarray_bounds[0]
+                    pixely = pixely - self.subarray_bounds[1]
                     #else:
                     #    # No distortion at all - "manual mode"
                     #    pixelx, pixely = self.RADecToXY_manual(ra, dec)
@@ -2653,7 +2661,7 @@ class Catalog_seed():
                     pixelx = entry0
                     pixely = entry1
 
-                    ra, dec, ra_str, dec_str = self.XYToRADec(pixelx, pixely, attitude_matrix,
+                    ra, dec, ra_str, dec_str = self.XYToRADec(pixelx+self.subarray_bounds[0], pixely+self.subarray_bounds[1], attitude_matrix,
                                                               coord_transform)
 
                 # Get the input magnitude


KevinVolkSTScI:
This is the fix for issue #91 that I posted earlier this week.  The unresolved question is whether changing this for NIRISS will break the code for NIRCam.  One would hope that the XYToRADec and RADecToXY routines work the same way for NIRCam and NIRISS, but it needs to be checked.

bhilbert4:
I'll run some tests using NIRCam simulations and make sure that everything looks ok.

KevinVolkSTScI:
Thank you Bryan.  I have to sort out how to test the code myself.  The issue is that I need to run on witserv3 or one of the other witserv machines to be able to get to /ifs/jwst/wit/nircam efficiently when I am not at STScI, and there are problems with getting to github when I am remote.  So I have been developing code and testing it on witserv but doing the github stuff from my laptop machine just copying over lines in the code by hand.  That is not a good long term way to work on things.  I may be able to do it all on witserv, if using firefox on witserv remotely is not too slow to be useful.  I need to experiment with this a bit.

bhilbert4:
Are you using VPN? When I'm at home I typically do everything on my laptop, and VPN allows me to see the needed directories. Running code on one of the witserv machines is nice because they are faster than the laptop. But I gave up on using firefox on witserv because it was so slow. Also, the version of git that is on the witserv machines is old enough that it caused problems for me during the normal workflow.

bhilbert4:
I tried a quick test this evening. Since your code updates are on your own branch, I didn't have access to them. I think I know how to clone your branch and keep it separate from the master branch, but I didn't have time to try that. Instead I ran a subarray test using the current master branch that does not have your fix in it. I created a NIRCam B5 full frame image as well as a B5 SUB160 image, and the point sources matched up and looked correct in both of them. I also wanted to try using a subarray that is offset from the center of the detector, but I found a bug in the yaml_generator that prevented me from setting up that test. 

What were your inputs in the case where your output subarray sources did not appear where they should have? Was your subarray centered on the full frame detector? If not, did you adjust the RA and Dec values in the subarray input file to compensate for the different locations of the full frame and subarray reference locations?

bhilbert4:
The code is agnostic when it comes to the location of the subarrays within the full detector (although this may be only for the case where sources are given in RA, Dec as opposed to x,y, given what you've found?). It relies on the RA and Dec entries in the input yaml file corresponding to the RA and Dec of the subarray (or full frame) reference location. So if your subarrays are not centered on the detector and use the same pixel as the reference location, then then RA and Dec entries in the input will have to be adjusted. I just happened to use subarrays centered on the detector last night because I already had the input file for that case. It's also the easiest case to test since the RA and Dec don't have to change. But for other NIRCam subarrays that are not centered on the detector, the situation is the same as it is for NIRISS. 

I'll try testing a NIRCam case again where the inputs are in units of x,y rather than RA, Dec. I'm assuming I'll see the same problem you were seeing. In that case then I believe your fix is what we need, assuming it doesn't affect the situation where the inputs are given in units of RA, Dec.

bhilbert4:
My test confirms what I said above. When the catalog inputs are in units of pixels, then I do see the same problem that you found. When the inputs are in units of RA, Dec, the positions look correct.

KevinVolkSTScI:
My understanding of things is that the (RA,Dec) position specified for a NIRISS observation should correspond to the reference pixel position within the subarray at the given orientation.  That is how I was doing things in rampsim.py.  Now I may have been mistaken about this, but I do not think we need to adjust (RA,Dec) for a subarray position.  The [V2,V3] reference position gets matched to the (RA,Dec) position specified directly.  When for example one acquires a star for either the AMI or SOSS modes of NIRISS the star is put directly at the defined reference position and the (RA,Dec) used are the ones for the star itself, not some offset position at the center of the detector.  But NIRCam may do things differently because of having multiple detectors.
  The fix for the pixel inputs is just to add self.subarray_bounds[0] and self.subarray_bounds[1] to the pixelx and pixely values in the call to XYToRADec so it does not affect anything else in the code.

bhilbert4:
I should have been more explicit. In the situation where you have a source that you want to place at pixel x, y (say (1747, 1747)) in a full frame exposure, and keep that source on the same pixel in a subarray exposure (subarray pixel (100,100) in a 400x400 subarray in the upper right corner of the detector), then the RA, Dec values in the subarray input must be different than those in the full frame input file.  This is because the reference location, and therefore V2,V3 reference values of the subarray differ from those of the full frame,

bhilbert4:
I've been looking at subarray simulations this morning. So far, what I've found is that the only time the source fails to appear where it should is when using NIRISS and with the point source inputs in units of RA, Dec. If the units are pixels, the source appears where it should. Also, when using NIRCam, the source always appears in the correct place, for inputs of RA, Dec and pixel values. Does this agree with what you were seeing?

I'm still trying to figure out what lines of code NIRSS simulations might be using that NIRCam is not. My first though was that maybe there was a typo in the SIAF. I double checked the NIRISS SIAF just to be sure, and the v2_ref, v3_ref entries look reasonable.

KevinVolkSTScI:
What I was seeing was problems with both the pixel positions and the RA/Dec inputs, **if** one assumes that the (x,y) pixel positions are with respect to the sub-array and not with respect to the full frame image.  If one assumes that the pixel positions are given for the full frame then the code works for pixel inputs...but I would have assumed that the pixel values should be with respect to the sub-array.  Of course if one has a set of pixel positions for sources in full frame imaging and wants the sources in the sub-array to have the same listed pixel position as in the full image then this works.

Johannes-Sahlmann:
Although I am not clear about the origin of the issues discussed here, my feeling is that they could be fixed by implementing the relevant transformations using pysiaf, which will treat all instruments in a consistent and relatively well-tested way. That would also easily allow to define pixel coordinates in either `detector` (full-frame) or `science` (subarray) pixels and apply the corresponding pysiaf transformations (the only detail is that we would have to agree and account for the pixel indexing scheme: 0 or 1-indexed).

bhilbert4:
I agree that `pysiaf` is the way to go in the long term. It seems like this shouldn't be a difficult fix in the short term though, until I get some time to play with `pysiaf`.  @KevinVolkSTScI, when using pixel values in the input catalogs, they are assumed to be with respect to the subarray coordinate system. For example, a source at (80, 80) will place the source at (80, 80) within whatever aperture you are simulating. I did get this part to work for NIRISS this morning. Did you set the `position_pixel` flag in the catalog to alert `Mirage` that the units were not RA, Dec?

KevinVolkSTScI:
When using pixel positions and the position_pixel flag for input to mirage, I was assuming for example that in both the NIRISS SUB64 and SUBTAAMI sub-arrays, which are 64x64 pixels but not at the same position on the full detector, that if one asks for a source at pixel position (32,32) then it would be at the center of the sub-array.  In such a case one only needs to know the dimensions of the sub-array to define the pixel positions one wants, and does not have to worry about where it is in the full detector.  Or one can use the reference (RA,Dec), the roll angle, and the reference pixel values listed in the SIAF file  to get relative pixel positions for stars (ignoring the small distortion effects), and again not have to worry about where the array is within the full frame image.

The changes in the code in this branch explicitly arrange for this to be the case by using the listed lower left corner pixel values from the niriss_subarrays.list file along with the (relative) sub-array pixel positions to get the RA/Dec.

bhilbert4:
That's all correct. I'm just worried that your changes will cause sources on NIRCam subarrays to be pushed to incorrect locations, since those seem to be working at the moment. I'm also worried about NIRISS subarrays where the inputs are in x,y, since those also are working (for me).

I've just been playing with `pysiaf` to see how much work it would take to integrate into `Mirage`. It shouldn't be much work, but I have some questions that I left for Johannes. In the meantime, I'll try testing your changes with NIRCam and NIRISS inputs.

bhilbert4:
Ok, I think I found the problem. In the NIRISS subarray definition file, the refpix_x and refpix_y columns list the reference location in full frame pixels. If you look in the NIRCam subarray definition file, you'll see that these should be in subarray coordinates. So for the NIRCam SUB160 subarray, the reference location is listed as (79.5, 79.5), which is the center of the array.

KevinVolkSTScI:
We do not currently have the position reference positions for the NIRISS sub-arrays defined in sub-array pixels because in one or two cases the reference positions are outside the subarray (i.e. SUBAMPCAL for example).  All sub-array reference positions for NIRISS as listed in the SIAF file are given in terms of the full array.
Maybe the code will be fine for this case as well with suitable negative reference position values in some cases.  One can always try it.  In making the NIRISS subarrays file I just copied the reference pixel positions directly from the SIAF file.

KevinVolkSTScI:
Further on the last comment, it is unlikely that we would ever use SUBSTRIP96 and SUBSTRIP256 for imaging as these are intended for SOSS mode only.  The code takes all the entries in the SIAF file except NIS_CEN_OSS to make the new file.  I could just exclude these sub-arrays from the output niriss_subarrays_new.list file on the basis that they will not be used in mirage.  However, I am worried that the bad values coming up for SUBSTRIP96 and SUBSTRIP256 imply that there is a problem with the other values as well....

bhilbert4:
Sounds good. It's probably best to wait until you're sure the new file is correct. Once it's ready, we can bring it in to the repository through its own separate branch. At that point we can remove this pull request since no change to the code was necessary to solve the issue.

KevinVolkSTScI:
I am going to test the revised file this week.  I was going to start today but I got an urgent SOCCER review request so I have to do that first.

Johannes-Sahlmann:
It appears that the problem is related to the xstart ... yend columns.
@bhilbert4 : can you please confirm that these columns are still used by mirage?
Also, can you please specify in which coordinate system these values are defined? The documentation reads 

> In addition, the file must list the x,y values of the lower left and upper right corners of the subarray,

KevinVolkSTScI:
There was an issue in the niriss_subarrays_new.list file, as I found out when I tried testing it.  I have made a correction and the first few tests look OK.  I am going to close the branch.

PR #93, opened 2018-05-15
Frame time correction

KevinVolkSTScI:
Correction for NIRISS frame time, #89.  Note the NIRCam frame time has not been updated here.

PR #101, opened 2018-06-25
Bugfix so multiple integrations with raw dark input is functional

bhilbert4:
Fixed a small bug in dark_prep.py that was causing the simulator to crash when requesting an output with (N > 1) integrations and using as input a raw dark exposure with fewer than N integrations. Updated documentation for the affected functions.

Also fixed two missing parentheses in frame time calculations that were causing code to crash. These must have been leftover from the fixes merged in #93?

### DIFF
--- a/nircam_simulator/scripts/dark_prep.py
+++ b/nircam_simulator/scripts/dark_prep.py
@@ -720,10 +720,22 @@ class DarkPrep():
         return dark, sbzero
 
     def darkints(self):
-        # Check the number of integrations in the dark
-        # current file and compare with the requested
-        # number of integrations. Add/remove integrations
-        # if necessary
+        """Check the number of integrations in the dark
+        current file and compare with the requested
+        number of integrations. Add/remove integrations
+        if necessary
+
+        Parameters
+        ----------
+
+        None
+
+
+        Returns
+        -------
+
+        None
+        """
         ndarkints, ngroups, ny, nx = self.dark.data.shape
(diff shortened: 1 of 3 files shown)


nircam_simulator/scripts/dark_prep.py
```diff
@@ -775,18 +803,36 @@ def integration_copy(self, req, darkint):
...
         if ngroup * (nskip + nframe) > inputframes:
-            print("WARNING: Not enough frames in the input integration to create the requested number of output groups. Input has {} frames. Requested output is {} groups each created from {} frames plus skipping {} frames between groups.".format(inputframes, ngroup, nframe, nskip))
-            print("for a total of {} frames.".format(ngroup * (nframe + nskip)))
-            print("Making copies of {} dark current frames and adding them to the end of the dark current integration.".format(ngroup * (nskip + nframe) - inputframes))
+            print(("WARNING: Not enough frames in the input integration to "
+                   "create the requested number of output groups. Input has "
+                   "{} frames. Requested output is {} groups each created from "
+                   "{} frames plus skipping {} frames between groups for a total "
+                   "of {} frames."
+                   .format(inputframes, ngroup, nframe, nskip, ngroup * (nframe + nskip))))
```
laurenmarietta: I had no idea that you could format multiline strings like this! Awesome!

bhilbert4:
Thanks for the quick review!!

PR #102, opened 2018-06-25
Added function to deal with NIRCam 'TIGHT' dither patterns

bhilbert4:
NIRCam has a new option when using the `FULL` dither pattern. Observers can specify the number of dither positions, as before, but now they also have the option to specify `3TIGHT`, which is a special case with 3 dither positions with smaller dithers in between. When `3TIGHT` is used, the number of primary dither positions as listed in the xml file from APT is "3TIGHT". This was throwing off apt_inputs.py, which previously converted (e.g. "3") to an integer. 

This fix introduces a small function that tests for the presence of the `3TIGHT` entires and if they are there, strip "TIGHT" from the strings.

### DIFF
--- a/nircam_simulator/scripts/apt_inputs.py
+++ b/nircam_simulator/scripts/apt_inputs.py
@@ -92,6 +92,14 @@ class AptInput:
         readxml_obj = read_apt_xml.ReadAPTXML()
         tab = readxml_obj.read_xml(self.input_xml)
 
+        # If the number of dithers is set to '3TIGHT'
+        # (currently only used in NIRCam)
+        # remove 'TIGHT' from the entries and leave
+        # only the number behind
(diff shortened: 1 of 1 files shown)


bhilbert4:
What would be PEP8 compliant in this case then? Extending the comment to be 72 characters before moving to the next line? I've always interpreted PEP 8 as giving a maximum line length, but not saying anything about shorter lines.

laurenmarietta:
Hmmm, I had never thought of it that way! I was interpreting it such that comments should extend to 72 characters before creating a new line, but you're right that there's no minimum line length requirement. I think I will post in the #python slack channel and see if any python gurus have thoughts.
 
(But if you want to merge the pull request aside from this, please go ahead.)

Issue #65, opened 2018-03-26
Add example for OTE Commissioning simulation


laurenmarietta:
Create iPython notebook detailing how to simulate NIRCam (and other instruments?) images during OTE Commissioning

laurenmarietta:
Closed via #103

PR #103, opened 2018-06-26
Add OTE Commissioning Example Notebook

laurenmarietta:
From #65 

Modifying my existing notebook to function as an example use notebook for using MIRAGE to simulate OTE Commissioning data. Added example APT files; made some small fixes.


laurenmarietta:
Thanks for the comments! I went ahead and changed everything you mentioned: changing the name of the package in the note, making the SIAF paths match, and adding a note about the `imaging_simulator` module. And thanks for the catch on the ITAR path...

I'll go ahead and merge it, then!

PR #105, opened 2018-06-26
WIP: Change to mirage

bhilbert4:
This PR encompasses all of the code changes necessary to support the repo name change from `nircam_simulator` to `mirage`.  Unfortunately, since the directory names also went from `nircam_simulator` to `mirage`, git considers all files in those directories to be new.

Actual substantive changes were only in setup.py, README, MANIFEST, gitignore, and then import statements in the various scripts and notebooks. I ran a NIRCam imaging test case with the updated code, and it ran successfully.

I did not yet make any changes to the `nircam_simulator_data` references. My thought is that those will probably be reorganized relatively soon now that I think we have all the NIRISS and FGS files. Once those files are reorganized, it should be possible to significantly simplify the input yaml files. We just need a function that knows the data file directory structure and can choose the appropriate files. All of that work needs its own branch and PR.

@laurenmarietta, @KevinVolkSTScI, @Johannes-Sahlmann, do you see anything obvious that I missed?


laurenmarietta:
This is kind of odd - I just recently changed the name of a package, and I didn't have the same problem with all the files being marked as new; they were just marked as having been moved.... and when I downloaded your branch, the `nircam_simulator` directory is still there, with all the old files in it. I'm not sure whether this is something I did wrong, or what could be going on...

When you renamed the directory, did you copy the `nircam_simulator` into the main repo and then rename it `mirage`? Or did you directly rename the `nircam_simulator` directory? I could imagine the first case leading to what I'm seeing on this branch...

bhilbert4:
That is odd. I didn't make any copies. I just renamed the directories. When I made the commit in git, it listed all the old `nircam_simulator` files as deleted. And then I essentially added all the files back under the `mirage` directory. 

I couldn't find any special instructions for what to do after renaming a repo, so I was winging it. Maybe I need to do some better googling.

laurenmarietta:
Odd indeed! Yeah, at least from my end, it seems like the deletion of the `nircam_simulator` files somehow wasn't included in the commit. Does the `nircam_simulator` directory still exist in your local copy?

This makes me a little nervous, because I don't think the file history is retained if the files are considered copied and deleted, rather than moved. For instance, it now looks like all of the past changes to `nircam_simulator/obs_generator.py` would not be retained in the new `mirage/obs_generator.py`: https://github.com/bhilbert4/mirage/commits/7a768efe9371e3357569310bedba053a65ec4d1b/mirage/scripts/obs_generator.py

I will also google around and let you know if I find anything helpful.

bhilbert4:
I think I solved the problem of the residual `nircam_simulator` files. They were deleted from my directory, but I needed to do a `git rm` to remove them from the repo. Fixing the lack of history I haven't yet figured out. I have found lots of complaints that history isn't tracked when a file is renamed, but no good solution yet. Still looking.

laurenmarietta:
Okay, I'm running the `test_imaging_simulator.py` script now, but I've run into a number of errors. Most of them were path definition problems that I was able to fix, but I just got this KeyError that I haven't seen before:
```
tests/test_imaging_simulator.py:14: in <module>
    m.create()
mirage/scripts/imaging_simulator.py:60: in create
    cat.make_seed()
mirage/scripts/catalog_seed_image.py:115: in make_seed
    self.checkParams()
mirage/scripts/catalog_seed_image.py:2940: in checkParams
    self.params['Telescope']['tracking'] = self.params['Telescope']['tracking'].lower()
E   KeyError: 'tracking'
```
It looks like there is supposed to be a "tracking" parameter in the input yaml file that isn't there. Have you seen this before?

laurenmarietta:
Okay, I got around the previous error by just manually adding a "tracking" parameter to the yaml (not sure why it wasn't already in there), but now it's hanging on the `obs_generator`. This is the last line I get before it seizes up:
```
Inverted IPC kernel saved to /ifs/jwst/wit/nircam/nircam_simulator_data/reference_files/saturation/Kernel_to_add_IPC_effects_from_NRCB3_17011_WellDepthADU_2016-03-10_ssbsaturation_wfact_DMSorient.fits for future simulator runs.
```
I will continue to investigate!

bhilbert4:
Hmmm. It just occurred to me that I may have gone about this all wrong. I still have an `add-fgs` branch that has yet to be merged into `master`. But with deleting and re-adding files for this name change work, I doubt that I'd be able to merge the `add-fgs` after the name change because the old files will be gone. And similarly, if I were to merge `add-fgs` first, I'm not sure those changes would be able to flow into the `mirage` branch, because again there's no connection between new files and old. I think I screwed up by renaming the directories with just `mv` rather than `git mv`.

bhilbert4:
Closing this PR. I need to redo the way I renamed directories and files in order to get everything to work right.

PR #107, opened 2018-06-28
Add FGS

bhilbert4:
This PR contains code to enable FGS within the simulator. I've done a simple test creating a 5 group integration, and the code ran successfully. The output looks as expected. I haven't done a close inspection of the output countrates compared with expectations.

Currently the throughput curves used for the guider include photon yield. That's correct, right @KevinVolkSTScI ? 

My thought is that it would be good to get this merged into `master` and have FGS functional but perhaps not perfect before moving on to address changes that need to be made for the name change to `mirage`. 

I have a yaml file and simple function that runs my FGS test. Let me know if you'd like that as a starting point for any testing you want to do.


KevinVolkSTScI:
Yes the FGS throughputs include the photon yield.  Ditto for the conversion factors from count rate to (photon-weighted mean) flux density and magnitude.

KevinVolkSTScI:
I also do not see any reference to guider_subarrays.list in the code.  However this may not be needed given that all calibration activities with the Guider use full frame imaging, as far as I know.

bhilbert4:
I added guider_zeropoints.list to the repo. I also updated the name of the NIRSS file to be niriss_zeropoints.list, in order to be consistent with the other instruments. 

I haven't investigated the `jwst_backgrounds` issue, but it is possible to test the code without using that module. Set the `bkgdrate` entry of the yaml file to a number (e/pixel/s). `jwst_backgrounds` is only used when `bkgdrate` is set to one of `high, medium, low`.

KevinVolkSTScI:
My test case has just a number in the background field, but the code fails in trying to import jwst_backgrounds so it was not possible to test it.

bhilbert4:
I'm working on testing why that is. If the background value is a number, then `jwst_backgrounds` should never be imported. In the course of testing for that, I found a small bug where one of the FGS changes to the code affects NIRCam and NIRISS. I'm working on fixing that, and then will try testing.

Also, I heard from someone else who was having trouble installing `jwst_backgrounds` that he was able to fix the problem by reinstalling `healpy`. That might be worth trying. https://healpy.readthedocs.io/en/latest/install.html

KevinVolkSTScI:
The issue I had trying to install jwst_backgrounds was that healpy was not compiling.  I will take a look at the page and see if this gives some work-around.

bhilbert4:
I just pushed a few updates to the code. I added `Detector` to the zeropoint table when using NIRCam or NIRISS, in order to be consistent with what we have for FGS. This is also anticipating detector-dependent zeropoints in the future.

I tested the updates using all three instruments, and all three ran successfully. I also confirmed that when `bkgdrate` is set to a number, the `jwst_backgrounds` module is never imported.

KevinVolkSTScI:
I will take a look at it in the next day or two.

KevinVolkSTScI:
The Guider zero point values in the file in the config directory do not correspond to the count rate values that I provided for the Guiders.  The values are too high by a factor of about 3.3.  The current file in config is

Detector Filter Module VEGAMAG ABMAG STMAG PHOTFLAM PHOTFNU Pivot_wave
GUIDER1 NA G 30.430280577747304 31.56893591813441 34.867472098580926 4.10214644$
GUIDER2 NA G 30.486673614112945 31.378557246458875 34.75434067260171 4.55264441$

while the input values that I calculated are

# Filter Module    Mag       ABMag     STMag     PHOTFLAM      PHOTFNU    Pivot Wave.
Guider1    G1    29.12446  30.31398  33.61256  1.303172e-22  2.719165e-32  2.5011
Guider2    G2    29.05481  30.29085  33.66668  1.239808e-22  2.777709e-32  2.5917

Where do the values currently in the file come from?  My values come from the same simulations that are used for NIRISS.  The pivot wavelengths cannot be longer than 4 microns for the Guiders.  The pivot wavelength is always near the mean wavelength.  For the Guiders if one has roughly flat response from 0.6 to 5.5 microns the mean wavelength will be about 3 microns.  In fact the pivot wavelength is smaller because the response (with the photon yield) is higher at the longer wavelengths.

If the Guider group is providing the values in the file, then we have to determine why these are not consistent with what I calculate.

bhilbert4:
The guider zeropoints file that you originally created was in different units, so it looks like I must have screwed up in translating. Here is the original file:

zero magnitude values, based on the Sirius 2014 spectrum

Filter       Count Rate   Mean F_lambda  Mean F_nu  Pivot Wavelength
              electron/sec  W/m^2/micron       Jy         Microns
Guider1       7.196973e+11  5.818174e-10  1.214005e+03   2.501078
Guider2       6.749752e+11  5.191312e-10  1.163080e+03   2.591652

I'll update the file to be the version that you showed above.

Issue #115, opened 2018-07-17
Add continuous integration service


Johannes-Sahlmann:
I suggest to use Travis, simply because I already know how to do it. 
If you object and lean strongly towards Jenkins, please comment this issue.

laurenmarietta:
Closed via #120

PR #120, opened 2018-07-17
Enable travis ci

Johannes-Sahlmann:
Addresses #115

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,68 @@
+# We set the language to c because python isn't supported on the MacOS X nodes
+# on Travis. However, the language ends up being irrelevant anyway, since we
+# install Python ourselves using conda.
+language: c
+
+compiler: gcc
+
+os:
+    - linux
+
+# Setting sudo to false opts in to Travis-CI container-based builds.
+sudo: false
+
+# The apt packages below are needed for sphinx builds, which can no longer
+# be installed with sudo apt-get.
+#addons:
+#    apt:
+#        packages:
+#            - graphviz
+#            - texlive-latex-extra
+#            - dvipng
+
+env:
+    global:
+        - MAIN_CMD='python setup.py'
+        - CONDA_CHANNELS='http://ssb.stsci.edu/astroconda'
+        - CONDA_DEPENDENCIES='pytest sphinx'
+        - PIP_DEPENDENCIES=''
+        - PYTHON_VERSION=3.6
+        - NUMPY_VERSION=stable
+        - ASTROPY_VERSION=stable
+        - EVENT_TYPE='push pull_request'
+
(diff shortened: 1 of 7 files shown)


Johannes-Sahlmann:
OK. The infrastructure is there and Travis will now automatically build the project and run one trivial test for any pull request. We should add more useful tests as we go.
Please have a look at the dependencies in `setup.py` and the way it installs the `jwst` package. It is likely that there are dependencies still missing because the tests do not cover much of mirage's functionality/imports.

bhilbert4:
I've been looking through the code. Here are my questions, some of which are based on my ignorance of how Travis/pytest work together:

Dependencies needed: jwst_backgrounds, NIRCAM_Gsim

I see where test_apt.py will be run by pytest, where you set `self.test_args = ['tests/test_apt.py']` in setup.py. Will the `test_xxx_imaging.py` functions also be run? I don't see where those are called, but maybe this is a case where pytest runs all functions that start with `test`?

And the way the code is set up now, every time a branch is merged into master, Travis will first install mirage and its dependencies, then run `RunAllAPTTemplates`, and also the `test_xxx_imaging.py` tests? Is that correct?

What's the best way to test this? Do we need to merge this into master, and then test by merging a branch with some trivial change? Or can we merge a branch with a trivial change into `add-travis-ci` and have it work?

Johannes-Sahlmann:
I had to limit the test scripts that are run because of some strange import error in the `test_xxx_imaging.py` scripts. Eventually, we will set `self.test_args = ['tests']` and pytest will run all test scripts that start with `test`. You can also see that I allowed most tests to fail (`@pytest.mark.xfail` decorator) because we have to find a public location for `$MIRAGE_DATA` first.

Currently, Travis will first install mirage and its dependencies, then run all tests in `test_apt.py`, allow the failures and find that `test_trivial()` passes.

You can make a trivial (or non-trivial) change to my branch in the pull request and Travis will automatically start after you pushed to my branch.

Johannes-Sahlmann:
For the missing dependencies, we should add tests that use those dependencies.

bhilbert4:
My thought is to do for `jwst_backgrounds` what you did for `jwst` in setup.py. Check if the package is installed, and if not, to try. That makes more sense to me than to create a pytest function, since the values returned by `jwst_backgrounds` may change over time through no fault of `mirage`. 

I'll find something more clever for NIRCAM_Gsim. Maybe create a pytest function that disperses a single pixel to compare to a known result, or something along those lines. 

Does that all make sense? I can try to add that if so.

bhilbert4:
Is it only me, or is subprocess.check_call really slow? I'm trying it using ["import","jwst"] and it takes >60 seconds.

PR #122, opened 2018-07-23
WIP: Use FittableImageModel (later EPSFModel) for psfs

bhilbert4:
This work is meant as preparation for a switch from using a full webbpsf-generated library containing a grid of sub-pixel centered PSF files to a situation where a single oversampled PSF will be stored in a photutils.psf.EPSFModel instance, which will do the sub-pixel interpolation. Additionally, PSF variations across the detector will be accounted for by interpolating over a grid of PSFs.

For this work, I have replaced the old method of reading in PSF files and working with them as numpy arrays with the use of FittableImageModel instances (this is because EPSFModel is not yet in photutils). I have also changed the code to prepare for detector position dependence by having a new PSF selected for each source. Previously, all galaxies and extended objects were convolved with self.centerpsf. Hopefully this will make it easier to swap in the detector position-dependent PSF later.

At the moment, the seed image produced by catalog_seed_image.py looks good except for a small (~<1 pixel) shift in galaxy and extended object locations from where they were previously. I tracked the shift to strange behavior from the output of the FittableImageModel `evaluate` method. I have asked Larry Bradley about it and am waiting on a response. Once that issue is resolved, we should be able to merge this branch and push forward with the other aspects of this PSF work.

### DIFF
--- a/mirage/scripts/catalog_seed_image.py
+++ b/mirage/scripts/catalog_seed_image.py
@@ -735,6 +735,27 @@ class Catalog_seed():
 
         return mtlist, pixelflag, pixelvelflag, msys.lower()
 
+    def minimal_psf_evaluation(self, model):
+        """
+        Create a PSF by evaluating a FittableImageModel instance. Return an array only just
+        big enough to contain the PSF data.
+
+        Parameters:
+        -----------
+        model : obj
+            FittableImageModel instance containing the PSF model
+
+        Returns:
+        --------
+        eval_psf : ndarray
+            2D numpy array containing the evaluated PSF, with normalized signal
+        """
+        eval_xshape = np.int(np.ceil(model.shape[1] / model.oversampling))
+        eval_yshape = np.int(np.ceil(model.shape[0] / model.oversampling))
+        y, x = np.mgrid[0:eval_yshape, 0:eval_xshape]
+        eval_psf = model.evaluate(x=x, y=y, flux=1., x_0=eval_xshape//2, y_0=eval_yshape//2)
+        return eval_psf
+
     def movingTargetInputs(self, file, input_type, MT_tracking=False,
                            tracking_ra_vel=None, tracking_dec_vel=None,
                            trackingPixVelFlag=False):
@@ -868,6 +889,18 @@ class Catalog_seed():
                 x_frames = pixelx + (entry['x_or_RA_velocity'] / 3600.) * frameexptimes
                 y_frames = pixely + (entry['y_or_Dec_velocity'] / 3600.) * frameexptimes
 
+            # Determine the filename for the most appropriate PSF
+            psf_file = self.find_subpix_psf_filename(pixelx, pixely)
+
+            try:
+                psf_exists = self.file_exists(psf_file)
+                if psf_exists:
+                    psf_model = self.populate_epsfmodel(psf_file, oversample=1)
+                else:
+                    raise FileNotFoundError("PSF file {} not found.".format(psf_file))
+            except:
+                raise RuntimeError("ERROR: Could not load PSF file {} from library".format(psf_file))
+            
             # If we have a point source, we can easily determine whether
             # it completely misses the detector, since we know the size
             # of the stamp already. For galaxies and extended sources,
@@ -875,7 +908,7 @@ class Catalog_seed():
             # the stamp lands on the detector.
             status = 'on'
             if input_type == 'pointSource':
-                status = self.on_detector(x_frames, y_frames, self.centerpsf.shape,
+                status = self.on_detector(x_frames, y_frames, psf_model.shape,
                                           (newdimsx, newdimsy))
             if status == 'off':
                 continue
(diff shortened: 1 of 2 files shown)


laurenmarietta:
I just skimmed through this, and it is looking nice! I just have one thought on organization, which you can take or leave.

It might be helpful to isolate the parts of catalog_seed_image that are being used to get the PSF and add it to the data in their own separate module - something like `get_psf.py`. This module could then have a class in it with all of the functions you've written up here as methods: `minimal_psf_evaluation`, `find_subpix_psf_filename`, `file_exists`, `populate_epsfmodel`, maybe all of `makePointSourceImage` and `makeExtendedSourceImage`, future functions for interpolation and such. The class could even take an instance of `Catalog_seed` as an argument, so all the information you're working with in `catalog_seed_image.py` would still be available. In this setup, there would be some main function where you input the `Catalog_seed` argument, and the output of that function would be the array with all the PSFs added appropriately.

I think this might help make `catalog_seed_image` a little easier to understand (and a little shorter!), and might also reduce redundancy, since there are some parts of the code that are repeated for each kind of object (point source v. extended v. moving target).

Of course, I'm getting this impression off of a quick review of what you did, so this might not be as helpful for the code base as I am feeling. What do you think?

bhilbert4:
I think it's a good idea. Shortening and simplifying `catalog_seed_image.py` is a good idea. And you're right that it will make the next steps easier. I'll try rearranging things tomorrow.

bhilbert4:
PSF handling has been updated to use a new PSF class (in psf_generator.py). I also wrapped up some of the seed image versus PSF/stamp image overlap coordinate calculations into a separate function, rather than having them repeated for each source type.

@laurenmarietta, what do you think?

laurenmarietta:
This is lovely! I especially love that `cropped_coords` function; it went a long way in making things feel cleaner.

My only question: is there a reason you switched back from using relative imports (`from . import rotations`) to absolute imports (`from mirage.scripts import rotations`)?

KevinVolkSTScI:
I installed this branch and made an initial trial run.  The following error came up

  File "./testfull1.py", line 5, in <module>
    t1.create()
  File "/grp/jwst/nis1/kevin/anaconda3/envs/pipeline7.1.3/lib/python3.5/site-packages/mirage-1.1-py3.5.egg/mirage/scripts/imaging_simulator.py", line 56, in create
    cat.make_seed()
  File "/grp/jwst/nis1/kevin/anaconda3/envs/pipeline7.1.3/lib/python3.5/site-packages/mirage-1.1-py3.5.egg/mirage/scripts/catalog_seed_image.py", line 135, in make_seed
    self.seedimage, self.seed_segmap = self.addedSignals()
  File "/grp/jwst/nis1/kevin/anaconda3/envs/pipeline7.1.3/lib/python3.5/site-packages/mirage-1.1-py3.5.egg/mirage/scripts/catalog_seed_image.py", line 1323, in addedSignals
    psfimage, ptsrc_segmap = self.makePointSourceImage(pslist)
  File "/grp/jwst/nis1/kevin/anaconda3/envs/pipeline7.1.3/lib/python3.5/site-packages/mirage-1.1-py3.5.egg/mirage/scripts/catalog_seed_image.py", line 1766, in makePointSourceImage
    interval=self.params['simSignals']['psfpixfrac'], oversampling=1)
  File "/grp/jwst/nis1/kevin/anaconda3/envs/pipeline7.1.3/lib/python3.5/site-packages/mirage-1.1-py3.5.egg/mirage/scripts/psf_generator.py", line 52, in __init__
    raise FileNotFoundError("PSF file {} not found.".format(psf_filename))
FileNotFoundError: PSF file /ifs/jwst/wit/niriss/nircam_ramp_simulation_files/webbpsf_files/NIS/f200w/predicted/niriss_NIS_x1024_y1024_f200w_predicted_0_0p20_m1p20.fits not found.

The PSF pixel position for the star is being calculated as x = +0.2, y = -1.2 for some source, which should not be the case.  I cannot tell which source it is because there are many sources in the input image for this test (~31000 in the NIRISS field of view).  Running the same test with the default version of the code does not produce any PSF position that is out of range.

A test with just one source at the reference position shows the star on pixel offset in x from the reference position, the same as for the default version of the code.

I will see about inserting print statements to tell me what star position is causing this error tomorrow.

KevinVolkSTScI:
I am not entirely sure what the plan is for the change in the PSF generation, but I am wondering if it will actually save any space for us.  Tests indicate that the NIRISS PSF needs to be sampled at least with 11x11 oversampling to get an accurate PSF in the wings (20x20 oversampling would be better still, but it is very expensive to use).  Since I also have 11x11 sampling of the star positions in the PSF grid, one oversampled image is roughly the same size as the full set of sub-pixel position PSF files that are made for a given filter and optical model.  If we require really fine oversampling for astrometric field simulations, the size of the oversampled images will become quite large.

Another comment is that I have been running new PSF files for NiRISS with WebbPSF 0.7.0 and including distortion.  The run has been slow, but it looks like including the distortion has little effect on the output PSF.  Comparing the distorted and undistorted PSFs, one can see tiny changes in the pixel values, but overall the effect is probably too small to worry about given the other uncertainties in the calculations.

Issue #123, opened 2018-07-24
When *_pointsources.list is empty, mirage hangs indefinitely 


Johannes-Sahlmann:
When exercising the FGS mode of mirage I found that if there are no valid sources (e.g. the telescope pointing does not match the coordinates of the input source list) mirage hangs.

bhilbert4:
Fixed via a source distance filter, in #125

Issue #111, opened 2018-07-13
When a source position is too far from the specified field center position, mirage hangs.


KevinVolkSTScI:
I have found that mirage hangs in catalog_seed_image.py if the input source sky coordinates are too far from the field position specified.  By mistake I had coordinates for stars in a sky field near the M4 globular cluster at position (245.924, -26.505) as the input star list, but put the RA/Dec positions in the .yaml file for a comparison field at the same galactic latitude away from the globular cluster at position (228.083, -40.352).  The code gets to the call 

pixelx, pixely = self.RADecToXY_astrometric(ra, dec, attitude_matrix, coord_transform)

in catalogue_seed_image.py for the first object on the list, but it then hangs up or at least takes a very long time.  I killed the python run after 1 hour.  The code had still not gotten the pixel (x,y) values for the first target.

bhilbert4:
Fixed via a source distance filter, in #125

PR #125, opened 2018-07-24
WIP: Filter out sources far from nominal pointing

bhilbert4:
Added functionality to make a quick check of input source positions, and remove any that are very far from the nominal detector pointing. Previously, sources which were too far away off the detector were causing `Mirage` to hang.


bhilbert4:
I tested with point sources and the filtering seemed to work fine for me. Galaxies and extended sources already had their own more involved filtering. I double checked those with very far away sources, and those both completed successfully as well.

bhilbert4:
This addresses #111, and possibly #123?
@Johannes-Sahlmann I'm interested to hear if this fixes the problem you saw this morning in #123.

PR #128, opened 2018-07-27
WIP: Frametime calculation update

bhilbert4:
These updates have cleaned up the frame time calculation function. NIRCam had an error in the calculated full frame time. That error was fixed. The function was also cleaned up such that only one equation is used for all instruments/apertures and it was moved into `utils.py` since it is called within both `catalog_seed_image.py` and `obs_generator.py`. My tests have shown correct exposure times calculated for NIRCam and NIRISS full frame and subarrays. FGS exposure times match those given on the relevant JDox page: https://jwst-docs.stsci.edu/display/JTI/Fine+Guidance+Sensor%2C+FGS

### DIFF
--- a/mirage/scripts/catalog_seed_image.py
+++ b/mirage/scripts/catalog_seed_image.py
@@ -29,6 +29,7 @@ from . import read_siaf_table
 from . import set_telescope_pointing_separated as set_telescope_pointing
 from . import moving_targets
 from . import segmentation_map as segmap
+from .utils import calc_frame_time
 
 INST_LIST = ['nircam', 'niriss', 'fgs']
 MODES = {'nircam': ["imaging", "ts_imaging", "wfss", "ts_wfss"],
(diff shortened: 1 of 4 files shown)


KevinVolkSTScI:
When I look at the code changes listed as associated with this branch in firefox, it appears to be OK.  FGS should have the same frame time as NIRISS for a full frame image.  However when I install the code according to the instructions I get a different version of utils.py and the frame time for FGS is given as 10.61383 seconds.  I cannot tell for certain whether git is installing the wrong code (as it seems) or whether the code record associated with the branch is wrong somehow.  Either of these options should not happen.  I am perplexed.

I cannot approve the changes until this problem is resolved.  I suspect that Bryan has the right code and git is simply giving me problems once again.

bhilbert4:
According to the FGS JDox page, 10.6138 is the correct exposure time for a full frame FGS frame. The difference between FGS and NIRCam/NIRISS is because the colpad term is 6 for FGS and 12 for the other two instruments. The only time colpad is 12 for FGS is for the ACQ1 subarray. I'm basing all of this on what I read on the FGS page. https://jwst-docs.stsci.edu/display/JTI/Fine+Guidance+Sensor%2C+FGS

KevinVolkSTScI:
OK I will look at the page just to be sure.  I had thought that the frame time was the same for NIRISS and the Guider but it need not be so.

KevinVolkSTScI:
I am OK with the change but I do not see any button that lets me approve the pull request.

bhilbert4:
Usually the button appears close to the top of this page. I can't see it since it only appears for you, but if I remember correctly the button may be in a banner just under the title of the pull request, or along the right side, still close to the top.

KevinVolkSTScI:
There does not seem to be any button for reviewing the changes or approving them or not approving them.  There is a notice that the branch has conflicts that must be resolved.  It that why I am not able to approve the changes?

bhilbert4:
Ah, maybe that's it. I thought that would come after approval. Now that my git troubles for today are (mostly) solved, let me see if I can resolve conflicts.

bhilbert4:
Ok, conflicts resolved

Issue #129, opened 2018-07-27
Re-organization of scripts


laurenmarietta:
It might be wise to discuss how all the `mirage` scripts are currently organized within the repo, and whether there is enough reason to re-organize them. At the moment, every script is within the `scripts` directory, but designing more structure into the repo could make it easier to understand the scripts' functions.

Here's one particular option I came up with for an alternate directory structure:
```
mirage/
    examples/
    tests/
    mirage/
        catalogs/
        config/
        apt/
            __init__.py
            apt_inputs.py
            read_apt_xml.py
        seed_image/
            __init__.py
            catalog_seed_image.py
            get_psf.py
            moving_targets.py
            segmentation_map.py
        dark/
            __init__.py
            dark_prep.py
        ramp_generator/
            __init__.py
            obs_generator.py
            unlinearize.py
        setup/
            __init__.py
            get_catalog.py
            write_observationlist.py
            yaml_generator.py
        utils/
            __init__.py
            polynomial.py
            read_fits.py
            read_siaf_table.py
            rotations.py
            set_telescope_pointing_separated.py
            utils.py
            yaml_update.py
        __init__.py
        imaging_simulator.py
        wfss_simulator.py
    README.md
    setup.py
```

In this case, something like `apt_inputs` would then be imported with a new statement: `from mirage.apt import apt_inputs`.

I am not sure where `blot_image.py`, `coord_transform_code.py`, `crop_and_blot.py`, `crop_mosaic.py`, `fits_seed_image.py` would be placed, since I have never used these. 

I am curious what others' opinions are!

bhilbert4:
I think in general it's a good idea. Your layout would make things more clear to people writing and using the code. 

That being said, I'm a little gun shy at the moment due to all the trouble I had when trying to rename things from `nircam_simulator` to `mirage`. It may be unrelated, but I've been having problems with `git` on my machine ever since then.

laurenmarietta:
That is certainly fair - I realized while thinking it through that this would mean another major overhaul to really the entire repo and every single import statement. Definitely a daunting task, and not a particularly urgent one. 

I just thought I might document the idea here so we could discuss, and maybe decide to make an update sometime in the future.

Johannes-Sahlmann:
Looks good to me. Only suggestion is not to create a directory names `setup` to avoid confusion with `setup.py`. Could be named `initialize/`.

[...]

PR #130, opened 2018-07-30
Organize directory structure

laurenmarietta:
As described in #129, modifying directory structure to implement a more descriptive sub-package naming scheme.

Now within the `mirage` package there are the following sub-packages: `apt`, `catalogs`, `dark`, `ramp_generator`, `seed_image`, `utils`, and `yaml`.

I updated the import structures within all the scripts and example notebooks (updating for PEP8 compliance along the way).


bhilbert4:
Looks good! I like having the `imaging_simulator` and `wfss_simulator` at the top level, since those are what most people are going to import. 

Just a couple of suggestions. 

1. Update the version number in setup.py. I've been bad about doing this up until the last couple of PRs. In speaking with the NIRCam team, who are excited to have this go public, they requested a version number that will be tied to the handbook, even if that version number is incremented the next day. It's good practice to keep the version number updated anyway, so I've started trying to do this with each PR.

2. I wonder if the example catalogs that are currently in the `catalogs` directory should be moved over into the `examples` directory?

laurenmarietta:
I am happy to make both of those changes! Is there a place where the version number is being documented? I realized I'm not sure what the current number is...

bhilbert4:
Right now the version number is in `setup.py`. I looked around a little bit a while ago to see how it's most commonly done, but I never found any really great examples other than the number in `setup.py`.

laurenmarietta:
Ah, got it - so would you like me to increment that to 1.1.3 then?

I am in the process of moving the catalog files around and making sure the example notebooks still work - the WFSS notebook is going to take a while, so I will work on that more tomorrow.

bhilbert4:
@Johannes-Sahlmann I was trying at one point last week to figure out how `pysaif` did its versioning. I didn't realize that `relic` was a space telescope package. If setting that up can help with version numbering without us having to update numbers manually each time, then I'm all for it. I read through the example code for the relic package, although it's still not clear to me how it goes about updating the version number. I guess that's a discussion for a different PR though.

And I agree that with the reorganization in this PR, mirage should go to version 2.0.0.

PR #134, opened 2018-08-01
Implement Gridded PSF Libraries

bhilbert4:
In order to solve my `git` problems I had to delete and reclone my fork, which means that the previous pull request for this EPSFModel work was closed. This is just a new copy of the PR. Below I'll recreate the conversation from the previous PR.


bhilbert4:
from @bhilbert4:

This work is meant as preparation for a switch from using a full webbpsf-generated library containing a grid of sub-pixel centered PSF files to a situation where a single oversampled PSF will be stored in a photutils.psf.EPSFModel instance, which will do the sub-pixel interpolation. Additionally, PSF variations across the detector will be accounted for by interpolating over a grid of PSFs.
For this work, I have replaced the old method of reading in PSF files and working with them as numpy arrays with the use of FittableImageModel instances (this is because EPSFModel is not yet in photutils). I have also changed the code to prepare for detector position dependence by having a new PSF selected for each source. Previously, all galaxies and extended objects were convolved with self.centerpsf. Hopefully this will make it easier to swap in the detector position-dependent PSF later.
At the moment, the seed image produced by catalog_seed_image.py looks good except for a small (~<1 pixel) shift in galaxy and extended object locations from where they were previously. I tracked the shift to strange behavior from the output of the FittableImageModel evaluate method. I have asked Larry Bradley about it and am waiting on a response. Once that issue is resolved, we should be able to merge this branch and push forward with the other aspects of this PSF work.

bhilbert4:
from @laurenmarietta:

I just skimmed through this, and it is looking nice! I just have one thought on organization, which you can take or leave.
It might be helpful to isolate the parts of catalog_seed_image that are being used to get the PSF and add it to the data in their own separate module - something like get_psf.py. This module could then have a class in it with all of the functions you've written up here as methods: minimal_psf_evaluation, find_subpix_psf_filename, file_exists, populate_epsfmodel, maybe all of makePointSourceImageand makeExtendedSourceImage, future functions for interpolation and such. The class could even take an instance of Catalog_seed as an argument, so all the information you're working with in catalog_seed_image.py would still be available. In this setup, there would be some main function where you input the Catalog_seed argument, and the output of that function would be the array with all the PSFs added appropriately.
I think this might help make catalog_seed_image a little easier to understand (and a little shorter!), and might also reduce redundancy, since there are some parts of the code that are repeated for each kind of object (point source v. extended v. moving target).
Of course, I'm getting this impression off of a quick review of what you did, so this might not be as helpful for the code base as I am feeling. What do you think?

bhilbert4:
from @bhilbert4:

I think it's a good idea. Shortening and simplifying catalog_seed_image.py is a good idea. And you're right that it will make the next steps easier. I'll try rearranging things tomorrow.

bhilbert4:
from @bhilbert4:

PSF handling has been updated to use a new PSF class (in psf_generator.py). I also wrapped up some of the seed image versus PSF/stamp image overlap coordinate calculations into a separate function, rather than having them repeated for each source type.
@laurenmarietta, what do you think?

bhilbert4:
from @laurenmarietta:

This is lovely! I especially love that cropped_coords function; it went a long way in making things feel cleaner.
My only question: is there a reason you switched back from using relative imports (from . import rotations) to absolute imports (from mirage.scripts import rotations)?
Bhilbert4
I thought I read somewhere recently that absolute imports keep more with preferred style guides. Although it could be that I'm misremembering. Sometimes what the powers-that-be declare make the code more readable don't make sense to me, and this may be one of those cases.

bhilbert4:
from @KevinVolkSTScI:

I installed this branch and made an initial trial run. The following error came up
File "./testfull1.py", line 5, in 
t1.create()
File "/grp/jwst/nis1/kevin/anaconda3/envs/pipeline7.1.3/lib/python3.5/site-packages/mirage-1.1-py3.5.egg/mirage/scripts/imaging_simulator.py", line 56, in create
cat.make_seed()
File "/grp/jwst/nis1/kevin/anaconda3/envs/pipeline7.1.3/lib/python3.5/site-packages/mirage-1.1-py3.5.egg/mirage/scripts/catalog_seed_image.py", line 135, in make_seed
self.seedimage, self.seed_segmap = self.addedSignals()
File "/grp/jwst/nis1/kevin/anaconda3/envs/pipeline7.1.3/lib/python3.5/site-packages/mirage-1.1-py3.5.egg/mirage/scripts/catalog_seed_image.py", line 1323, in addedSignals
psfimage, ptsrc_segmap = self.makePointSourceImage(pslist)
File "/grp/jwst/nis1/kevin/anaconda3/envs/pipeline7.1.3/lib/python3.5/site-packages/mirage-1.1-py3.5.egg/mirage/scripts/catalog_seed_image.py", line 1766, in makePointSourceImage
interval=self.params['simSignals']['psfpixfrac'], oversampling=1)
File "/grp/jwst/nis1/kevin/anaconda3/envs/pipeline7.1.3/lib/python3.5/site-packages/mirage-1.1-py3.5.egg/mirage/scripts/psf_generator.py", line 52, in init
raise FileNotFoundError("PSF file {} not found.".format(psf_filename))
FileNotFoundError: PSF file /ifs/jwst/wit/niriss/nircam_ramp_simulation_files/webbpsf_files/NIS/f200w/predicted/niriss_NIS_x1024_y1024_f200w_predicted_0_0p20_m1p20.fits not found.
The PSF pixel position for the star is being calculated as x = +0.2, y = -1.2 for some source, which should not be the case. I cannot tell which source it is because there are many sources in the input image for this test (~31000 in the NIRISS field of view). Running the same test with the default version of the code does not produce any PSF position that is out of range.
A test with just one source at the reference position shows the star one pixel offset in x from the reference position, the same as for the default version of the code.
I will see about inserting print statements to tell me what star position is causing this error tomorrow.

bhilbert4:
from @KevinVolkSTScI:

I am not entirely sure what the plan is for the change in the PSF generation, but I am wondering if it will actually save any space for us. Tests indicate that the NIRISS PSF needs to be sampled at least with 11x11 oversampling to get an accurate PSF in the wings (20x20 oversampling would be better still, but it is very expensive to use). Since I also have 11x11 sampling of the star positions in the PSF grid, one oversampled image is roughly the same size as the full set of sub-pixel position PSF files that are made for a given filter and optical model. If we require really fine oversampling for astrometric field simulations, the size of the oversampled images will become quite large.
Another comment is that I have been running new PSF files for NiRISS with WebbPSF 0.7.0 and including distortion. The run has been slow, but it looks like including the distortion has little effect on the output PSF. Comparing the distorted and undistorted PSFs, one can see tiny changes in the pixel values, but overall the effect is probably too small to worry about given the other uncertainties in the calculations. Attempts to use the position dependent PSF option have so far failed to produce any changes in the results. A PSF calculated for the corner of the NIRISS array comes out exactly the same as in the center. Things may be different for NIRCam as far as using that option is concerned.

mperrin:
In reply to @KevinVolkSTScI's comment from a couple days ago:

First, as I see it the point of this change is not to save disk space, it's to get more accurate results by allowing smooth interpolation of PSFs to subpixel positions. The EPSF framework allows interpolation of PSFs as continuous functions to very fine precision (hundredths of a pixel), without having to have PSFs sampled that finely in a huge lookup table.

I'd be curious to understand more how you arrived at the conclusion that you need 11x11 sampling to get an accurate PSF. What criteria are you basing that on? That seems much finer than should be necessary, in particular based on the Nyquist sampling theorem which says once you're well sampled for your shortest wavelength (0.8 microns at the blue end of F090W) then 0.5&lambda;/D = 0.012 arcsec, so 6x6 subsampling should be sufficient to model all the spatial information in the PSF. I suspect a likely factor here may be webbpsf's simple approach of just summing together the subsampled data values to get the detector sampling. The approach implemented here for the ePSF model by @shanosborne includes convolving by a smoothing function based on the pixel size. That should make the PSF results converge more rapidly without needing excessively high sampling factors. 

Secondly, I'm not surprised to hear the distortion has a relatively small effect. NIRISS isn't too bad in that regard. I _am_ surprised to hear you're not getting any variations from the position dependence. That should easily be seen. Can you file an issue report on the webbpsf repo with some example code for what you're doing?

mirage/seed_image/catalog_seed_image.py
```diff
@@ -885,22 +893,25 @@ def movingTargetInputs(self, file, input_type, MT_tracking=False,
...
+            # Evaluate the PSF model. Place in stamp image just large enough to contain it
+            eval_psf = psf_obj.minimal_psf_evaluation()
+            
             if input_type == 'pointSource':
-                stamp = self.centerpsf
+                stamp = eval_psf
 
             elif input_type == 'extended':
-                stamp, header = self.basic_get_image(entry['filename'])
+                stamp, header = self.basicGetImage(entry['filename'])
```
mperrin: minor comment: `basicGetImage` is not a PEP8 compliant name.

mirage/seed_image/catalog_seed_image.py
```diff
@@ -1928,25 +1920,7 @@ def cropPSF(self, psf):
...
-            astropy Table containing catalog
-
-        pflag : bool
-            Flag indicating units of source locations. True for detector
-            pixels, False for RA, Dec
-
-        msys : str
-            Magnitude system of the source brightnesses (e.g. 'abmag')
-        """
+        # Read in the point source list
```
mperrin: This looks accidental; PR should be updated to not delete the docstring of this function.

mirage/seed_image/catalog_seed_image.py
```diff
@@ -122,15 +123,12 @@ def make_seed(self):
...
-        self.frametime = utils.calc_frame_time(self.params['Inst']['instrument'], self.params['Readout']['array_name'],
-                                         self.nominal_dims[0], self.nominal_dims[1], self.params['Readout']['namp'])
-        print("Frametime is {}".format(self.frametime))
+        self.calcFrameTime()
 
         # Read in the pixel area map, which will be needed for certain
         # sources in the seed image
         self.prepare_PAM()
-
+
```
mperrin: This PR has a lot of trailing whitespace changes, which I expect are probably accidental. Perhaps some cleanup commit in master that hasn't been merged into this PR yet?

mirage/seed_image/psf_generator.py
```diff
@@ -0,0 +1,160 @@
...
+        y_position : float
+            Row number on the detector where the PSF will be located
+
+        psf_base : str
+            Base of the PSF filename (e.g. "nircam"
+
+        interval : float
+            Sub-pixel centering resolution of the PSF library (e.g. 0.25)
+
+        oversampling : int
```
mperrin: Could `oversampling` and `interval` be read from the PSF file header, rather than explicitly provided as parameters?

bhilbert4:
Hmm, there's something strange going on here. It's proposing to take out some of the recent changes that have been put in (e.g `get_basic_image` is the updated version, `getBasicImage` is older and was swapped out). Maybe I need to rebase or something. This PR isn't ready to merge yet anyway. I've been working on the new interpolation function in this branch.

mperrin:
Yeah I expected it was something along those lines. And yes I know this isn't ready to merge - I just figured I should at least start to read through so I have a better idea myself of how this functionality is coming together!  I think I'm going to need to understand some of the details to help figure out how to handle some of the weird cases relevant for OTE commissioning...

mirage/seed_image/psf_generator.py
bhilbert4: `interval` is not in the file header. It's essentially the resolution of the files in the library. This function was written as a quick fix under the assumption that we would merge the branch and use `fittableImageModel` with the current PSF libraries. But now that we have @shanosborne's tool to create new libraries, I was thinking that development in this branch could push forward to use those libraries and get the position-dependent PSFs implemented. In that case, this class will have to be updated.

KevinVolkSTScI:
The 11x11 oversampling for NIRISS was a result of some testing that Shannon did to see how much oversampling was needed to get the far wings of the PSF to be stable.  This was in response to the comment by John Stansberry that for NIRCam one needed significant oversampling to get the far wings of the PSF correct.  Previously I had been using 5x5 oversampling for the NIRISS PSF library.  I do not know exactly what criteria Shannon used to judge when the far wings were stable, but it was something like less than 1% deviation from the profile at very high sub-sampling in the wings at 150 pixels from the PSF center, since we use 301x301 pixel PSF images for NIRISS and the Guider.

KevinVolkSTScI:
I did not get back to see what star is causing the error that I reported above where some star is assigned position x = +0.2 y = -1.2.  I have lost track of which conda environment has this branch in it, so it may take some time to figure this out and see why the value is out of range.

mirage/seed_image/catalog_seed_image.py
laurenmarietta: The text editor that I use - Sublime - includes an option to delete all trailing whitespace every time you save, which has been very helpful. I'm not sure what editor you're using, Bryan, but you might have a similar option!

KevinVolkSTScI: You can have emacs take out trailing spaces.  See https://www.emacswiki.org/emacs/IndentingPython .

bhilbert4:
@shanosborne, in your PSF library generator, you describe the shape of the output file as `[SCA, i, j, x, y] where (i,j) is the PSF position on the detector grid (integer positions) and (x,y) is the 2D PSF.`. Does this mean you have essentially switched the x and y dimensions from the way that they are used in `numpy`? For example, if I have a PSF library open and pull out the PSF at `i=2, j=1` (where DET_IJ9 = (2,1) and DET_XY9 = (1365, 682)) this means I should do `my_array[sca, 2, 1, :, :]`? And this will give me the PSF at x=1365, y=682 on the detector?  Similarly, once I have that 2D PSF, do I need to transpose it assuming that I will be placing it in a simulated image which is in a numpy array where I have the detector y axis aligned with the numpy y axis (i.e. (y,x))? I'm confusing myself a little, and just want to make sure that my code for extracting a PSF is doing the right thing.

mperrin:
I believe Shannon is still out on vacation until Thurs. 

My guess/hope is that that’s just a typo in the description; it should probably be just `y,x` instead, with no need to transpose.

shanosborne:
Responding to @KevinVolkSTScI from a week ago: Sorry for the delayed response, I was out on vacation. If you’re referring to the analysis done for the NIRISS Predicted PSF JDox page, we had actually decided on a 9x9 oversampling. This value came from trying to achieve values accurate to 1% for both FWHM and encircled energy for the PSFs across all filters. This does seem higher than one would really need- and at the time we thought it might mean that something is off in the measurement code, that would make the pixel scale more dependent than it should be when calculating the FWHM and encircled energy. So at the time we decided to go with a larger oversample to make sure our PSFs were as accurate as possible for the PSF plots/tables for the JDox page. 
In the current case of NIRISS oversampling for this PSF library, I agree with @mperrin that something like a 6x6 oversampling looks like it will work well (FWHM and encircled energy values are accurate within 1% for the smoothed PSFs).

shanosborne:
Also I just wanted to add to this thread that @bhilbert4 ’s comments on the x,y notation in the psf_library script was in fact a typo in the documentation and that has since been updated with pull request #149

KevinVolkSTScI:
I stand corrected about 11x11 over-sampling.  I am not certain that 6x6 oversampling is enough, but that can be checked later and if for some simulation one really needs more oversamping it can be done explicitly for a given simulation or set of simulations.

laurenmarietta:
Hi Bryan, any thoughts as to what needs to happen for this to be merge-able?

I am in the process of testing a preliminary version of MIRAGE with commissioning PSFs using a branch that includes these commits, but I've run into a funny problem (`error: (mx>kx) failed for hidden mx: regrid_smth:mx=1`) that I seems to be due to the interpolation in `psf_selection.py`. As I'm pretty unfamiliar with this PR, I'm not sure if it's a problem with my input or if it's because the PR is still incomplete.

mirage/seed_image/psf_generator.py
```diff
@@ -0,0 +1,160 @@
+#! /suyr/bin/env python
+
+"""
+Class to generate PSFs to be used with Mirage
```
laurenmarietta: A small thing - would this file be better located in mirage/psf rather than mirage/seed_image?

bhilbert4:
Hard to say where that error could be coming from. Does it trace all the way back to photutils, or is it in mirage itself? My impression is that the sub-pixel interpolation done in photutils works but may not be extensively tested. But in the (more likely) case that the error is in the detector position interpolator that I wrote, I haven't seen it before and am interested in digging into it.

There is a hack day scheduled for photutils/ePSF stuff on Oct 8 that I should be going to. I believe that interpolation is on the list of things to discuss, but I don't know any other details at the moment.

eteq:
A couple things to note here: the EPSF builder is now in photutils so hopefully you can switch to that.  *however*, the default smoothing kernel used to generate the EPSF is optimized for 4x4 oversampling, so you may need to build a custom smoothing kernel to make it work correctly (the existing one *might* work, but hasn't been tested for very different oversamplings).

laurenmarietta:
@bhilbert4 The problem isn't in photutils; it's with `scipy.interp2d`. (I'll attach the full traceback as a txt file here.) I haven't had time to read through the code to get a sense of what's happening yet, but my gut feels that it might be because the case I am using only has 1 fiducial PSF (rather than 4x4 or 9x9 or something). That is to say, there doesn't actually need to be any interpolation between different PSFs to approximate the PSF at the specific pixel on the detector, because there is only 1 PSF provided for each detector. 

[psf_selection_error.txt](https://github.com/spacetelescope/mirage/files/2397278/psf_selection_error.txt)

@eteq Great to hear that EPSFBuilder is in photutils now! However, I don't think we are using it in this particular case (we are generating ePSFs directly from WebbPSF rather than extracting them from an image).

bhilbert4:
Thanks for the traceback. I'll have a look at it later today. You're probably right that it's due to having a single fiducial PSF.

mirage/seed_image/psf_generator.py
bhilbert4: Before this branch is merged, I'm pretty sure that file should be deleted. It relies on the older-style PSF filenames (rather than files created by Shannon's functions), and loads them into a FittableImageModel instance.

bhilbert4:
The latest commit has `psf_selection.py` working in my test case of a PSF library with a single fiducial PSF. However, there is a slight offset in the location of the PSF on the detector, and the signal level is too low. I'm still trying to figure out what's going on there.

laurenmarietta:
Quick note: the environment files will need to be updated to `photutils>0.6`. (Which might cause many other conflicts, I am currently troubleshooting.)

bhilbert4:
Good catch. I'm thinking that it would probably be a good idea to follow `jwql`'s lead and use `pyup` to try and help keep the environment file up to date.

laurenmarietta:
Ooh, pyup is a great idea! 

I tried to just update `photutils` in my `mirage` environment, but this *broke Jupyter notebook* because it updated `tornado` but didn't update `nbconvert`. 🤦‍♀️ Thank goodness for StackOverflow!

bhilbert4:
Interesting. I don't think I had that issue when updating `photutils`. Maybe I haven't tried using it in a notebook since I updated though.

bhilbert4:
I think this PR is finally ready for testing! Last week I made an some initial gridded PSF libraries for NIRISS. I also have libraries for just a couple NIRCam filters. My checks with both have shown sources being placed in the correct locations. For NIRCam, the 5X oversampled libraries resulted in sources being placed in the seed image within several hundredths of a pixel of the requested position (when using input catalogs specifying pixels rather than RA, Dec). This was much more accurate than the ~0.1-0.2 pixel accuracies I was seeing for the older libraries with NIRCam, where we have a 4x4 grid of sub-pixel PSF locations.

The initial NIRISS libraries I created are also oversampled by a factor of 5. It seems like this may end up being less accurate than the current 10x10 sub-pixel grid of PSFs for NIRISS, but I figured it would be good enough for this initial testing. There is a small bug in webbpsf that needs to be resolved before we create the final gridded libraries anyway. For this testing, I just want to make sure that Mirage doesn't crash, and sources with the correct signals are placed in the correct locations on the detector.
Also note that the current gridded PSF libraries for NIRISS use the `predicted` wavefront error, and `realization` 0. These are both the defaults used by the `yaml_generator`. 

I have updated the `yaml_generator` to point to the new gridded PSF libraries by default.

@Johannes-Sahlmann @KevinVolkSTScI do either of you have some quick test cases for this? For NIRCam I was working with several well-separated point sources and I calculated the centroid and did aperture photometry (with very large radii) on the sources in the seed images (as opposed to the raw ramps output by the `observation_generator'.)

I haven't yet worked on creating libraries for FGS, since they will be complicated by spectral type.
Also note that I'm waiting for webbpsf to add the PUPIL value to the header of the gridded libraries that it outputs. So running this branch you'll see a reminder to that effect printed to the screen.

bhilbert4:
Timing tests with a new gridded PSF library are giving disappointing results. Simulating part of the LMC on one detector is showing that just over 3 sources are added to the seed image per second. For my timing test, where there are 8900 sources in the FOV, this translates to 45 minutes needed to place the sources in the seed image. Compare this to the old library, where the code needed only 1 minute to add all the sources.

Does this mean we abandon the griddedPSF library? Perhaps we can re-make the files in the original PSF library such that they include distortion? Except that would still not include PSF variations across the field of view, which was the main motivation for switching to the griddedPSF library in the first place.

KevinVolkSTScI:
I am  of the opinion that we need to keep both possibilities, maybe in separate branches, because for some purposes the old astrometric accuracy is sufficient and for other purposes we need the better PSF simulation that the gridded library provides.  Whether we can make the code able to use either method with some sort of swtich in the parameter file is not clear to me.

bhilbert4:
I'm sure it would be possible to set up the code to use one library or another based on a switch, but it would take a lot of work. The changes I made to switch to the new libraries were pretty invasive. I reorganized a lot of the code for creating the seed images to make it more clear and easier to maintain. I'll have to think about how feasible this is, and I don't think I'd be able to get to it until later in the summer.   

I'm not in favor of keeping the two versions in different branches though because that means two versions of the code to maintain going forward. For example, next on my list is to add TSO support before the summer student arrives in early June to start creating TSO datasets. I wouldn't want to have to do that to two separate branches.

KevinVolkSTScI:
Well maybe the work needed to have both possibilities is excessive.  If the gridded PSF version is too slow for general use it may still be needed for special cases where one really needs PSF variations over the field and better astrometric accuracy in the placing the stars.  One would not need to keep supporting new features in the gridded PSF version if it is only practical for such specialized use where one just lives with the relatively slow generation of the seed images.  I am not sure what is best.  It seems a shame to entirely abandon the gridded PSF version when you have put so much work into it.

mperrin:
Hold on - let's not jump to any drastic solutions too fast. I don't think anyone has done any significant work on benchmarking or optimizing the PSF grid code yet. There may well be ways to optimize and speed that up. In my opinion we should investigate this avenue and understand where the computational cost is coming from, before just assuming it cannot be faster than it currently is.

bhilbert4:
True. My gut feeling is that the bottleneck is in the `evaluate` function of the library, but I haven't looked into it in detail.

As @arminrest pointed out, interpolating a 1500x1500 pixel grid (301x301 pixels, subsampled by a factor of 5) will take time. His suggestion is to try using multiple gridded libraries in Mirage, each with different dimensions. Then for a point source of a given magnitude, the most appropriate library is chosen and interpolated. That way for dim sources which cover only a small area, you would only have to interpolate a (e.g.) 50x50 pixel image, which would presumably save a lot of time compared to using the full 1500x1500 pixel library.

bhilbert4:
Part of my frustration here is that I need to enable TSO simulations in the next month. I was hoping to branch off the gridded PSF branch for that in order to make use of the reorganized seed image code. But perhaps it's best to branch off of master and keep to the old libraries for that, and then reorganize that bit when the gridded PSF libraries are brought in.

mperrin:
1500x1500 sounds fairly daunting! I am not surprised that's slow. Definitely not optimized in terms of the choice of that size, I think.  Is that the same size as the PSFs you were using before? 

I very much agree that something like 50x50 pixels is likely sufficient for a majority of use cases (that corresponds to ~ 95% encircled energy for NIRCam). Furthermore, the spatial variation of wavefront error is primarily on low and mid spatial frequencies, which corresponds to changes in the inner regions of the PSF. Based on current information I do not think we should expect large variations in the outer parts of the PSF - so one possible approach could be to compute a field-dependent small inner core of the PSF, then (for the brighter sources that need it) paste that onto a single realization of the outer extent of the PSF with no field dependence.

bhilbert4:
After discussing with @mperrin, we think that the best way forward is to use gridded PSF libraries that are smaller than those I've been using up until now. Looking at F070W in NIRCam, the vast majority of the differences in the PSF across the field are in the inner 50x50 pixels or so. So the idea would be to use a library of this size in order to accurately capture the PSF variations, and then paste this into a static image of the PSF wings and place that into the seed image. I tested the interpolation of a 50x50 library and found that Mirage was placing 200 sources per second into the seed image that way (compared to just over 3 sources per second with the 300x300 library). This brings execution time back to about where it is with the older style library.

I'll work on adding the infrastructure to paste the results into an image of the wings.

For cases where a user would want higher accuracy out in the PSF wings, once they generate the appropriate library, Mirage will still be able to do this, at the cost of computation time. My plan is to start with gridded libraries that are either 50x50 for all filters, or perhaps scaling each filter's library to be the same number of lambda/D as the 50x50 library in F070W. And of course the oversampling factor is another parameter that can be played with for these libraries.

mperrin:
And as discussed, the outer wings of the PSF are only needed for cases in which the imaging setup will achieve a dynamic range > 1e4 on the PSF in question. (i.e. the outer wings of the PSF, outside the 50 pix box, are at least that much fainter than the PSF core).

To answer your question about discontinuities from Slack, I've ran some proof of concept tests pasting an inner core 50x50 region onto a set of the PSF wings calculated for a different field point. If you blink the two images you can see where the boundary is. But if you just look at an individual image then there is no obvious discontinuity. The overall complexity/nonuniformity of the PSF speckle pattern makes it hard to see the boundary at 50 pixels where the two are pasted together. I tested this for both F070W (broadband, under sampled) and F212N (narrow band, well sampled) and it's basically the same aesthetically in both cases. 

As I stated in Slack, the field dependence of the distortion does result in small changes/displacements in the angles of the big diffraction spikes, but these are small, of order ~1 degree. We expect few if any use cases will need higher precision in the outer PSF halo diffraction spikes than that. So, for all but those probably-rare & currently-unidentified use cases, it seems acceptable to use just use a single version of the outer PSF across each entire detector. 

One way to motivate the choice of different sizes for gridded libraries would be in terms of the encircled energy. Because the optical system isn't fully diffraction limited at the shortest wavelengths, the radius for a given encircled energy doesn't scale simply as lambda/D.  Let's say for instance we want to ensure the gridded library contains > 95% of the PSF flux.  I found that for F070W, 95% of the energy is within 0.65"=21 pixels radius, while for F212N, 95% of the energy is within 1.0"=32 pixels radius. So if we scaled the PSF grid size from ~42 to ~64 pixels over the SW filters, with an appropriate values for each filter we could ensure that in all cases at least 95% of the flux is in the gridded library. The choice of 95% is semi-arbitrary of course; this is just an idea for one possible way to think about this.

bhilbert4:
nircam_position_error_vs_pixel_phase_x
nircam_position_error_vs_pixel_phase_y


I've run a comparison of Mirage using the new gridded PSF libraries versus the same scene created with the current master branch. For this test I used a series of well-separated sources that are  decreasing in brightness from left to right and bottom to top on the detector.

These plots show the results. The first two show the difference between the measured source locations and the requested source locations (in the input source catalogs) versus the pixel phase of the source location, in x and y. The current NIRCam PSF library contains PSFs centered every 0.25 pixels in x and y. This is why the location differences are comparable every 0.25 pixels. For pixel phases other than those contained in the current PSF library, the new gridded PSF library does a much better job at placing the sources close to the requested location. This will be useful for simulations used for astrometric purposes, including the simulation of NIRCam pre-imaging for NIRSPec observations. The handful of points on the left side of the plots with large differences for both the gridded and current libraries are for sources with the peak of the PSF on the edge of the detector.

nircam_photometry_vs_source_location_y
 
The third plot shows the ratio in the photometry (really just the aperture sum from `photutil`'s `aperture_photometry` using `CircularApertures`) of the gridded PSF library compared to the current library, plotted versus the location in y on the detector. This shows that as the sources get intrinsically dimmer (as you go from the bottom to the top of the detector), the results from the gridded library are getting dimmer more slowly than those from the current library, up to a difference of 1.5%. Could this be due to the gridded library containing distortion effects while the current library does not? (@mperrin)

Note that all of this analysis was done on the noiseless seed images from Mirage. Other than the point sources, the only other signal present is from the background, which is also getting dimmer from the bottom to the top of the detector, but is identical in both cases.

mperrin:
> the results from the gridded library are getting dimmer more slowly than those from the current library, up to a difference of 1.5%. Could this be due to the gridded library containing distortion effects while the current library does not? (@mperrin)

I do not understand that at all. The geometric distortion effect should be irrelevant for this: it very slightly redistributes light between adjacent pixels, but in a way that is completely independent of source brightness. WebbPSF knows nothing about the brightness of different sources in MIRAGE of course. I think the origin of this must be somewhere in the MIRAGE code. 

I'm skeptical this is some unintended interaction with the background levels. Perhaps try re-running this simulation with a constant background, or better yet with precisely zero background, and see if the flux anomaly still shows up then?

bhilbert4:
The difference seems to be coming from the griddedPSFmodel itself. I don't quite understand what though. 

Here I read in a library, and move up the detector, evaluating the library every 200th row, and printing the sum of the evaluated PSF. You can see that the PSFs are getting brighter as the row number increases. The difference between the PSFs at the bottom and top of the detector is about 1.5%, in agreement with what I saw in my earlier test case. @mperrin any idea what this could be from?

```
    from webbpsf.utils import to_griddedpsfmodel
    import numpy as np 
    file = 'nircam_nrcb1_f200w_clear_fovp69_samp5_npsf9_predicted_realization0.fits' 
    lib = to_griddedpsfmodel(file) 
    xc = 600
    for yc in np.arange(200, 1800, 200):
        ypts, xpts = np.mgrid[yc-30:yc+31, xc-30:xc+31]
        gridded = lib.evaluate(x=xpts, y=ypts, flux=lib.oversampling**2, x_0=xc, y_0=yc)
        print(yc, np.sum(gridded))
```

```
200 0.9627568066466106
400 0.9651547023238878
600 0.9675525980011652
800 0.9699504936784424
1000 0.9723483893557199
1200 0.9742666667562383
1400 0.9761195416644715
1600 0.9779724165727045
```

mperrin:
Ohhhh - in that case, this is probably the fact that NIRCam has different image quality in the center of each module compared to in the corners. The impact of field dependent WFE!  If you make the same calculation ordered by x instead of y, I predict you will also see a gradient. 

Except wait, the sign of the effect confuses me: the image quality for NIRCam is the best near the center of each module, so near (0,0) for NRCB1. Therefore I would expect slightly higher encircled energy in lower rows, which seems to be opposite of what you're getting. 

It could be interesting to check what the sum is for each PSF in the reference library grid. Does that have the tightest PSF and best encircled energy nearest (0,0) for NRCB1 and a more spread out PSF nearest (2047, 2047)?  That's what I would expect based on the SI WFE models.

mperrin:
Reference chart of NIRCam spatially variable WFE in SW: 

[NIRCAM_SW_WFE_from_CV3.pdf](https://github.com/spacetelescope/mirage/files/3173740/NIRCAM_SW_WFE_from_CV3.pdf)

mperrin:
Looking at some output PSF grids I've calculated myself here, the behavior that you're seeing in MIRAGE does seem to be consistent with what I'm seeing in the output grids from WebbPSF. In other words I am seeing the reference PSFs in the grid have higher summed intensities for larger Y coordinates on NRCB1. That's not what I expect based on the optical models and I will have to investigate more. Maybe my intuition is wrong, maybe there's a bug in something  related to the use of the different field points.

bhilbert4:
It looks to be opposite of what you expect. Could I have flipped the coordinate system somehow when I created the files?

```
file = 'nircam_nrcb1_f200w_clear_fovp69_samp5_npsf9_predicted_realization0.fits'  
d = fits.getdata(file)
```

The header of the file gives the PSF locations:
```
DET_YX0 = '(0.0, 0.0)'         / The #0 PSF's (y,x) detector pixel position     
DET_YX1 = '(1024.0, 0.0)'      / The #1 PSF's (y,x) detector pixel position     
DET_YX2 = '(2047.0, 0.0)'      / The #2 PSF's (y,x) detector pixel position     
DET_YX3 = '(0.0, 1024.0)'      / The #3 PSF's (y,x) detector pixel position     
DET_YX4 = '(1024.0, 1024.0)'   / The #4 PSF's (y,x) detector pixel position     
DET_YX5 = '(2047.0, 1024.0)'   / The #5 PSF's (y,x) detector pixel position     
DET_YX6 = '(0.0, 2047.0)'      / The #6 PSF's (y,x) detector pixel position     
DET_YX7 = '(1024.0, 2047.0)'   / The #7 PSF's (y,x) detector pixel position     
DET_YX8 = '(2047.0, 2047.0)'   / The #8 PSF's (y,x) detector pixel position     
```

```
xloc = [0, 0, 0, 1024, 1024, 1024, 2048, 2048, 2048] 
yloc = [0, 1024, 2048, 0, 1024, 2048, 0, 1024, 2048]  
for i, (x, y) in enumerate(zip(xloc, yloc)): 
    print(x, y, np.sum(d[i, 172-150:172+151, 172-150:172+151]))

0 0 0.9583333376710805
0 1024 0.9707103851430886
0 2048 0.9802533492744265
1024 0 0.9607357162530266
1024 1024 0.9729319038811644
1024 2048 0.982349378264462
2048 0 0.9605447294836333
2048 1024 0.9724728258275833
2048 2048 0.9813440141129403
```

mperrin:
@shanosborne, @obi-wan76 - see the last few comments in this thread for an interesting puzzle regarding field of dependence of PSFs  and encircled energies in WebbPSF output grids.  Any chance either of you might have time to help investigate some this week, please?

mperrin:
After some further investigation, this variation in PSF ensquared energy is being driven by the distortion terms, not the variation in WFE. If I calculate a grid with `add_distortion=False` in the call to `psf_grid`, then I see almost no change in the PSF sum across the various points in the PSF grid. 

So, I think what is happening here is that the distortion term is stretching the PSF out slightly, and moving some flux outside of the fixed size aperture, in a way that is mostly variable across the Y dimension. Whether that makes sense quantitatively given the distortion polynomials present in the SIAF files is something we will need to dig into.

bhilbert4:
Interesting. For what it's worth, I checked the encircled energy using radii of 25 and 45 pixels, and the differences as I stepped up the detector in y were nearly identical for the two cases.

bhilbert4:
@KevinVolkSTScI @Johannes-Sahlmann, as the photometry effects described above seem to be related to distortion rather than a problem in Mirage, do either of you have any interest in testing out this branch? I intend to run a similar test for NIRISS as I ran for NIRCam above, in order to confirm the astrometric and photometric performance. But if you have other tests you'd like to run before this branch is merged, you're welcome to go for it.

Johannes-Sahlmann:
I am going to give it a try.

bhilbert4:
In informal conversations, @KevinVolkSTScI and @Johannes-Sahlmann have both said that they approve of merging this branch. So I'll merge it now and we can deal with any lingering issues in separate PRs.

Issue #132, opened 2018-07-31
catalog_seed_image throws errors for moving targets


laurenmarietta:
When I was running `MovingTarget_simulator_use_examples.ipynb`, I ran into the following error while trying to run the `create_seed_image` module:

```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
~/TEL/mirage/mirage/seed_image/catalog_seed_image.py in readPointSourceFile(self, filename)
   1950         try:
-> 1951             gtab = ascii.read(filename)
   1952             # Look at the header lines to see if inputs

~/miniconda3/envs/OTECommSims/lib/python3.5/site-packages/astropy/io/ascii/ui.py in read(table, guess, **kwargs)
    350         # more useful traceback.
--> 351         dat = _guess(table, new_kwargs, format, fast_reader)
    352         if dat is None:

~/miniconda3/envs/OTECommSims/lib/python3.5/site-packages/astropy/io/ascii/ui.py in _guess(table, read_kwargs, format, fast_reader)
    496             reader.guessing = True
--> 497             dat = reader.read(table)
    498             _read_trace.append({'kwargs': guess_kwargs,

~/miniconda3/envs/OTECommSims/lib/python3.5/site-packages/astropy/io/ascii/core.py in read(self, table)
   1145         # Get a list of the lines (rows) in the table
-> 1146         self.lines = self.inputter.get_lines(table)
   1147 

~/miniconda3/envs/OTECommSims/lib/python3.5/site-packages/astropy/io/ascii/core.py in get_lines(self, table)
    292                 with get_readable_fileobj(table,
--> 293                                           encoding=self.encoding) as fileobj:
    294                     table = fileobj.read()

~/miniconda3/envs/OTECommSims/lib/python3.5/contextlib.py in __enter__(self)
     58         try:
---> 59             return next(self.gen)
     60         except StopIteration:

~/miniconda3/envs/OTECommSims/lib/python3.5/site-packages/astropy/utils/data.py in get_readable_fileobj(name_or_obj, encoding, cache, show_progress, remote_timeout)
    192                 timeout=remote_timeout)
--> 193         fileobj = io.FileIO(name_or_obj, 'r')
    194         if is_url and not cache:

FileNotFoundError: [Errno 2] No such file or directory: 'temp_non_sidereal_point_sources.list'

During handling of the above exception, another exception occurred:

OSError                                   Traceback (most recent call last)
[...]

PR #135, opened 2018-08-01
Moving Targets Bug Fixes

laurenmarietta:
Fix bugs (#132) with temp file in `catalog_seed_image`, replace galaxy list file in `moving_target_example.yaml`, update example notebook, update version to 2.0.1


examples/movingtarget_example_data/moving_target_test.yaml
```diff
@@ -55,7 +55,7 @@ simSignals:
   psfpixfrac: 0.25                           #Fraction of a pixel between entries in PSF library (e.g. 0.1 = files for PSF centered at 0.1 pixel intervals within pixel)
   psfwfe: predicted                               #PSF WFE value (0,115,123,132,136,150,155)
   psfwfegroup: 0                             #WFE realization group (0 to 9)
-  galaxyListFile: movingtarget_example_data/moving_galaxies.list
```
bhilbert4: I think I may have gotten myself confused yesterday and mentioned the wrong catalogs for this example. This notebook creates a scene assuming non-sidereal tracking of a target. So `movingTargetToTrack` should be populated with `movingtarget_example_data/moving_target_to_track.cat`. Then the other catalogs will provide background (sidereal) sources that are going to be trailed in the exposure. So `pointsource` should be set to `imaging_example_data/seed_im_from_catalog_test_ptsrc_catalog.list`. The `movingTargetList`, `movingTargetSersic`, and `movingTargetExtended` should all be set to None. If you wanted trailed ellipses in the output, then you could set `galaxyListFile` to `imaging_example_data/seed_im_from_catalog_test_galaxies_catalog.list`.

The `movingtarget_example_data/moving_galaxies.list` catalog comes into play for a sidereal exposure where you want to have ellipses trailing through the background. In this case you can specify a different velocity for each target, whereas in the non-sidereal case above, the velocity is the same for all background sources.

I'm running the notebook now with the catalogs I mentioned above, but it takes a long time because there are so many sources. Maybe I should update the example catalogs so that they have only a couple sources in them.

laurenmarietta:
Okay, I will update the `moving_target_example.yaml` to reflect what you've said here - so just the `movingTargetToTrack` and `pointsource`.

I'm also thinking that (a separate issue) it would be really great to have the different yaml parameters explicitly defined in documentation somewhere - be it in the notebook examples, in a readme, or somewhere else - so each individual parameter and what they mean for different simulation cases is documented unambiguously. Especially since I'm realizing that, in this particular case, the naming of movingTargetList, movingTargetSersic, movingTargetExtended, movingTargetConvolveExtended, and movingTargetToTrack isn't really detailed enough to distinguish between them...

examples/movingtarget_example_data/moving_target_test.yaml
laurenmarietta: Done!

bhilbert4:
Looks good! One final request, in the spirit of not publishing full paths in the repo, push a version of the notebook where none of the cells have been run. This is just because some of the output cells contain path information. This can change once we update `mirage` to use logging rather than printing to the screen.

laurenmarietta:
I was also just talking to Erik Tollerud and he said that generally it is better not to push executed notebooks, ever! (whoops.)

From him: 
>In the notebook as you've checked it in, you've executed all the cells. That means the notebook is 8MB, primarily because of all the images. This is not desirable because git isn't meant for tracking content changes like that. Instead, you should check into git the unevaluated notebook (i.e. do a "clear outputs" before the git add). That's a much better file both for reviewing on github and for storing in git. Then when the notebooks are actually hosted somewhere, we automatically run them. That's also a good check because it determines if the notebook actually works as written, rather than only working because you did something in a cell you then later deleted (for example).

PR #143, opened 2018-08-07
Updated FGS and NIRISS subarray lists

bhilbert4:
Updates to the Guider and NIRISS subarray list files in the config directory. No code changes associated with this PR.


bhilbert4:
Note that the dependence on these files should go away once we integrate with `pysiaf`.

Johannes-Sahlmann:
Do you want me to double-check the numbers in the subarray lists?

For some obscure reason the travis checks fail for this PR.

bhilbert4:
I noticed the Travis failure. I thought it was odd but wasn't too worried about it. 

Could you just spot check a couple of entries?

KevinVolkSTScI:
Which case failed, was it FGS or NIRISS?  Most of the values are pulled from the SIAF .csv file, but (a) if I am not using the most up to date SIAF .csv files then there could be an error, and (b) some of the arrays names are called FULL when the SIAF entry is a full frame image but with a different reference point as in the MIMF cases.  Maybe that is causing an error.

bhilbert4:
As far as I can tell, the failure is simply that the attempt to install `mirage` failed. It shouldn't have anything to do with the subarray file itself.

Johannes-Sahlmann:
I agree, the test failure seem to be unrelated to the changes introduced by the PR.

KevinVolkSTScI:
I am using NIRISS_SIAF_09-28-2017.csv and FGS_SIAF_2017-08-31.csv, those being the latest that I have on my disk.

All the xref and yref values should be offset by 0.5 pixels compared to the SIAF, because of the way mirage counts the pixel center compared to whta the SIAF file does.  If there are 1.5 pixel offsets then I need to look into what is happening in these instances.  The code itself only puts in fixed 0.5 pixel offsets from the SIAF values to the values written out for mirage.  There should not be 1.5 pixel offsets if the SIAF file values are up to date.

Johannes-Sahlmann:
The 1.5 pixel offset for NIRISS is for refpix_x of NIS_AMITA

KevinVolkSTScI:
Once I get access to the "official" SIAF .xlsx files for NIRISS and the Guider I will redo the generation of the files and see whether there is any issue with NIS_AMITA.  As far as use in the code is concerned there is nothing about that entry that is treated differently than the others, so it may be that there is a bad value in the particular version of the SIAF file that i was using.

Johannes-Sahlmann:
@bhilbert4 :  Is it necessary that xstart<xend and ystart<yend? 

Also, can you please point out the place in the code that uses these values?

bhilbert4:
xstart and xend are used to define self.subarray_bounds, line 3493 in catalog_seed_image.py. If xstart is larger than xend then the code should fail, I believe, since then there would be an attempt to do something like array[2000:1000, 2000:1000].

For what it's worth, as soon as I get this updated treatment of the PSFs done, I think I'm going to start work on moving to `pysiaf` so that we can avoid all of these types of issues.

Johannes-Sahlmann:
To approve this PR and to implement the `pysiaf` dependence, we have to write a piece of code that replaces `self.subarray_bounds = [self.subdict['xstart'].data[mtch][0], self.subdict['ystart'].data[mtch][0], self.subdict['xend'].data[mtch][0], self.subdict['yend'].data[mtch][0]]`  
with code that depends on `pysiaf`. I am not fully there yet.

bhilbert4:
Is that because there is some problem with the values in the subarray file? I started a separate branch yesterday for the `pysiaf` conversion. There are a number of things that need to be replaced in addition to setting the `subarray_bounds`.

Johannes-Sahlmann:
OK. Please let me know if you need help with the pysiaf stuff.

Before continuing on NIRISS/FGS, I wrote code to try to reproduce the current mirage/config/NIRCam_subarray_definitions.list on the basis of the current NIRCam_SIAF.xml file (PRDOPSSOC-H-015).
For most apertures I can reproduce `xend` and `yend` exactly but I find a one pixel offset for `xstart` and `ystart`. 
Half-pixel offsets in x/y-start/end are found for NRCA3_DHSPIL_SUB96, NRCB4_DHSPIL_SUB96, NRCA3_FP1_SUB64, NRCA3_FP1_SUB8. 
Large offsets in x/y-start/end are found for NRCA5_TAGRISMTS32, NRCA3_SUB64P, NRCA5_SUB64P, NRCA5_TAGRISMTS32.

I suggest we meet to discuss, understand, and clean this up. Then I can generalize my code for NIRISS/FGS and contribute it to your pysiaf-dependency branch.

bhilbert4:
As I am working on fixing bugs in the PSF updates that I am making, I'm keeping a close eye on indexes. @Johannes-Sahlmann, what is the convention used for SIAF? It's indexed to 1, but is the center of a pixel (0.0,0.0) or (0.5,0.5)? In my PSF test case, I placed a source at (1023.5, 1023.5) and the source landed on the corner of the pixel. I placed another source at (900.0, 900.0), and it was centered on the pixel.

bhilbert4:
I should mention that it appears that convention agrees with what ds9 presents. Moving left to right across a single pixel, the x-coordinate goes from -0.5 to +0.5.

Johannes-Sahlmann:
SIAF used 1-based pixel indexing and the center of a pixel corresponds to integer values. Pixel corners have half-integer values.

bhilbert4:
The updated NIRISS and FGS subarray definition files are included in #148, after some modifications to remove columns containing information now retrieved from `pysiaf`. Closing this PR.

PR #148, opened 2018-08-16
WIP: Pysiaf integration

bhilbert4:
This PR introduces code to use `pysiaf` when calculating things like subarray sizes and positions, and for translating between detector pixels and RA, Dec. This will eliminate the dependence on the subarray definition file in the `config` directory except for 2 pieces of information: the number of amplifiers used to read out a particular aperture, as well as the OPGS (I think?) name for the aperture. This name is needed only to populate metadata in the output file. The number of amplifiers used is needed in order to get an accurate result for the calculation of the readout time. In addition, once this work is complete, the distortion reference files and csv versions of the SIAF will no longer be needed.

### DIFF
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -312,7 +312,7 @@ class AptInput:
                         obsnum, visitnum = v.split(':')
                         obsnum = str(obsnum).zfill(3)
                         visitnum = str(visitnum).zfill(3)
-                        if skip == True:
+                        if skip is True:
                             print('Skipping observation {} ({})'.format(obsnum, obslabel))
 
                     try:
@@ -334,10 +334,16 @@ class AptInput:
                             observation_label.append(obslabel)
                             observation_number.append(obsnum)
                             visit_number.append(visitnum)
-                            vid = str(propid) + visitnum + obsnum
+                            prop_5digit = "{0:05d}".format(int(propid))
+                            vid = "{}{}{}".format(prop_5digit, obsnum, visitnum)
                             visit_id.append(vid)
+                            # Visit group hard coded to 1. It's not clear how APT divides visits up into visit
+                            # groups. For now just keep everything in a single visit group.
                             vgrp = '01'
                             visit_grp.append(vgrp)
+                            # Parallel sequence id hard coded to 1 (Simulated instrument as prime rather than
+                            # parallel) at the moment. Future improvements may allow the proper sequence
+                            # number to be constructed.
                             seq = '1'
                             seq_id.append(seq)
                             tar.append(np.int(elements[0]))
@@ -370,8 +376,8 @@ class AptInput:
                             expar.append(np.int(elements[19]))
                             dkpar.append(np.int(elements[20]))
                             ddist.append(np.float(elements[21]))
-                            observation_id.append('V' + vid + 'P00000000' + vgrp + seq + act)
-                            act_counter  += 1
+                            observation_id.append("jw{}_{}{}{}_{}".format(vid, vgrp, seq, act, exnum))
+                            act_counter += 1
 
                     except:
                         pass
--- a/mirage/yaml/yaml_generator.py
+++ b/mirage/yaml/yaml_generator.py
@@ -880,7 +880,7 @@ class SimInput:
         input -- dictionary containing all needed exposure
                  information for one exposure
         """
-
+        instrument = input['Instrument']
         # select the right filter
         if input['detector'] == 'NIS':
             filtkey = 'FilterWheel'
@@ -901,12 +901,15 @@ class SimInput:
             outfile = input['outputfits']
             yamlout = input['yamlfile']
         else:
+            if instrument.upper() == 'NIRCAM':
+                fulldetector = 'nrc{}'.format(input['detector'].lower())
+            else:
+                fulldetector = input['detector'].lower()
             outtf = True
-            outfile = input['observation_id'] + '_' + input['detector'] + '_' + input[filtkey] + '_uncal.fits'
-            yamlout = input['observation_id'] + '_' + input['detector'] + '_' + input[filtkey] + '.yaml'
+            outfile = input['observation_id'] + '_' + fulldetector + '_uncal.fits'
+            yamlout = input['observation_id'] + '_' + fulldetector + '.yaml'
 
         yamlout = os.path.join(self.output_dir, yamlout)
-        instrument = input['Instrument']
         with open(yamlout, 'w') as f:
             f.write('Inst:\n')
             f.write('  instrument: {}          # Instrument name\n'.format(instrument))


bhilbert4:
@Johannes-Sahlmann, I'm seeing some funny behavior with the corner calculations that I don't quite understand. Here's an example that was causing a shift in y in the output files of my test:

```
import pysiaf
# Get info for NRCB5_SUB160
aperture = pysiaf.Siaf("NIRCam")["NRCB5_SUB160"]

# In subarray coord system we see the reference location is centered
# within the subarray
print(aperture.XSciRef, aperture.YSciRef)
# (80.5, 80.5)

# Location of SUB160 reference location in full frame coord system
print(aperture.XDetRef, aperture.YDetRef)
# (984.5, 1028.5)

# Now calculate the corner locations of the subarray within the full frame
from mirage.utils import siaf_interface
xcorner, ycorner = siaf_interface.sci_subarray_corners("NIRCam", "NRCB5_SUB160")

print(xcorner)
# array([ 904, 1063])
print(ycorner)
# array([ 940, 1099])

# If this is right, then the reference location should be halfway between
# the top and bottom/left and right edges. (ADD 1 to the corner values above
# since those are 0-indexed.)
# In x, the coordinate of the center is:
print((905 + 1064)/2)
# 984.5

# and in y:
print((941 + 1100) / 2)
# 1020.5
```
The x value above agrees with `aperture.XDetRef`, but the y value differs from `aperture.YDetRef` by 8.

Johannes-Sahlmann:
Concerning the pysiaf dependence, I am not sure If you mean to eliminate the dependence on the wcs reference files entirely? That would not be desirable, because the wcs reference file will carry the science-grade distortion information and provide filter-dependence. In my opinion, the transformations between `sci` and `tel` frames should therefore be implemented using the wcs_reference file.

bhilbert4:
Thanks. I updated my test script, and things look better now. The same sources in NRCB5 full and sub160 land where I expect them to, and extracting the SUB160 aperture from the full frame data gives an array that matches the SUB160 data. There are still differences in B1, but I think that is the round-trip error showing up.

bhilbert4:
As for the distortion reference files, NIRCam currently has no filter dependence, but once we are in flight, and filter dependencies to the distortion will not be captured in SIAF? Is that correct?

Johannes-Sahlmann:
Correct. SIAF currently captures one set of distortion coefficients without association to a specific filter, see also https://jira.stsci.edu/browse/JWSTSIAF-121

bhilbert4:
My NIRCam tests of this code have been successful so far, modulo the roundtrip error in the current version of our reference files and the current delivered version of pysiaf. For a given catalog of sources I simulate a full frame image, and then a subarray image at the appropriate RA and Dec, and the sources show up in the same locations for both.

The `sci_subarray_corners` function needs to be updated to select the appropriate corners for NIRISS and FGS before this branch is merged.

bhilbert4:
No need for version 0.5 specifically. But at some point in the near future, EPSFModel should be added to photutils, at which point I believe we'll want to update Mirage to use EPSFModel (rather than FittableImageModel), and update to using whatever version of photutils that happens to be.

bhilbert4:
Hmm. So now it's failing on `abc.py` and `typing.py`. I don't really understand what it's trying to do.

Johannes-Sahlmann:
It may be that some unwanted side-effect of the pysiaf dependency is causing the Travis failure? The problems start after `Adding pysiaf 0.1.10 to easy-install.pth file`.
I am not sure how to tackle this.

bhilbert4:
Travis also fails for the FittableImageModel PR. Comparing this PR to that one, it seems like `jinja2` is the last thing being worked on before the failure. In both cases the final line before the trouble starts seems to be `File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/jinja2/utils.py", line 486, in <module>
    MutableMapping.register(LRUCache)`

Maybe I'll see if anyone on the python slack channel has any thoughts on it.

laurenmarietta:
Odd - why does MIRAGE require Jinja2? I understand that package to be used for templates for web applications...

bhilbert4:
My only guess is that Travis itself is using it.

setup.py
```diff
@@ -70,7 +70,9 @@ def run_tests(self):
         'lxml>=3.6.4',
         'asdf>=1.2.0',
         'scipy>=0.17',
-        'photutils>=0.4.0',
+        'photutils==0.4.0',
```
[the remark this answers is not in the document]
pllim: Huh... I agree with Dan's suggestion to try installing this branch on a fresh conda environment locally and see what happens.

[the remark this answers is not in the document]
Johannes-Sahlmann: That was my attempt to see whether the issue we see is related to the photutils version. It should be reverted.

bhilbert4:
@pllim edits to .travel.yml seem to have worked. Checks now pass. She recommended that we remove pytest from setup.py and pin them in the .travis.yml file. Since this is working at the moment, I'll save that for a separate PR.

mirage/utils/siaf_interface.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Parameters:
+    -----------
+    instrument : str
+        Instrument name.
+
+    aperture : str
+        Aperture name (e.g. "NRCA1_FULL")
+
+    Returns:
```
pllim: You don't need a "returns" section if it does not return anything. FYI.

setup.py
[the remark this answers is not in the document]
pllim: Yup, I recommend reverting this also. `==` pinning will only bring you pain.

setup.py
```diff
@@ -70,7 +70,9 @@ def run_tests(self):
...
+        'pysiaf',
+        'pytest<=3.6'
```
pllim: There is no need to include `pytest` unless you are writing a `pytest` plugin or something. This is because typical users won't care about the tests; they won't run the tests, so they don't need `pytest` to use your package. Adding it as a hard requirement here will also bring you pain.

tests/test_siaf_interface.py
```diff
@@ -0,0 +1,27 @@
+"""Test the functions provided by siaf_interface.
+
+Authors
```
pllim: FWIW there is no need to list authors explicitly like this. `git blame` and `git log` will tell you anyway. And such lists will get out of date unless all the authors are careful about updating it. To give proper git/GitHub credits, it is better to amend the commit that added this file to the original author (if possible).

tests/test_siaf_interface.py
```diff
@@ -0,0 +1,27 @@
...
+-------
+    - Johannes Sahlmann
+
+Use
```
pllim: Is this useful? Might be confusing as it is not really what `.travis.yml` is doing (it is running all tests, right?). People who wish to run the tests can get the commands from `.travis.yml`.

bhilbert4:
@Johannes-Sahlmann is it much work to figure out subarray corners for NIRISS and FGS? Is it something I can do with a few test cases, or is it more complicated and better left to you as the SIAF expert? You mentioned the other day that it might depend on an angle of some sort?

I think that other than that addition, this PR is ready for review.

bhilbert4:
I don't want to merge this PR before having the NIRISS and FGS subarray corners being chosen correctly since that would mean that NIRISS and FGS simulations would potentially be wrong, whereas at the moment they are (mostly) correct.

I'll try running some test cases and see if it becomes clear which subarray corners the code should be using for each.

bhilbert4:
For FGS, `sci_subarray_corners` fails because it looks like there is no FGS definition file in the `source_data` directory of the `pysiaf` package. I see directories for NIRCam, NIRISS, and NIRSpec, but not FGS nor MIRI.

Johannes-Sahlmann:
I just updated pysiaf (version 0.1.12) to include the FGS definition file. It's also on pypi.

Johannes-Sahlmann:
The NIRISS SIAF TR is JWST-STScI-006317. 
corner coordinates in X of -24 are also reported there.

bhilbert4:
For what it's worth, I just ran a test simulating an FGS subarray using the `FGS1_SUB32CNTR` aperture and while `mirage` completed successfully, the `jwst` package threw an error saying that `SUB32CNTR` is not a valid entry for the SUBARRAY header keyword. I'm not sure if the FGS subarray list I have is wrong, or if DMS does not include all of the latest FGS subarray names. @KevinVolkSTScI

bhilbert4:
If `mirage` completed successfully though, it means that `FGS1_SUB32CNTR` is defined in `pysiaf`, so maybe DMS is behind in their list of valid subarrays?

mirage/utils/siaf_interface.py
bhilbert4: Removed...

tests/test_siaf_interface.py
bhilbert4: I think it's useful to have, if only to be able to quickly know who to send any questions to.

tests/test_siaf_interface.py
bhilbert4: I'm a little confused by this, primarily because of my inexperience with Travis and `setup.py`. But at the moment, I think the only tests that are being run are those in `test_apt.py`, as mentioned in `setup.py`. But anyway, don't we need to have a `test_siaf_interface.py` function in order to have Travis run it?

bhilbert4:
I think this is ready for a final review. Does anyone have time to grab the branch and run a test or two? Everything works for me, but it'd be nice to have confirmation. @KevinVolkSTScI? @laurenmarietta?

KevinVolkSTScI:
I can test this later this week or next week.

laurenmarietta:
An error with git closed this PR, which was recreated as #156

Issue #133, opened 2018-07-31
Implement versioning


laurenmarietta:
Use `relic` to implement versioning.

From Johannes in #130:

> For https://github.com/spacetelescope/pysiaf I adopted the versioning solution from the stsci package template which relies on tags defining the version number (e.g. 0.1.3) and the relic package. I can help implementing this if desired.
> 
> btw, see https://semver.org/ for versioning best practices. If we follow this, I guess this merge should lead to mirage version 2.0.0

This also relates to creating releases for the repository itself, as described here: https://help.github.com/articles/creating-releases/

laurenmarietta:
Source code for `relic`: https://github.com/spacetelescope/relic 
I didn't realize this was an ST package!

Description:
>RELIC stands for "Release I Control" and maintains a git project's version information without the need for hardcoding values into the source code. It was designed to aid developers with rapid release cycles.
>
>Just tag your code and get on with with your life. It's that simple.

laurenmarietta:
Closed via #177

PR #152, opened 2018-08-23
Add license, code of conduct, style guide

bhilbert4:
This PR adds a license file, the `mirage` code of conduct, and a style guide to the repo. Those of you working on the `jwql` project will notice that large parts of these files have been adapted from `jwql` files.


README.md
```diff
@@ -230,4 +230,45 @@ simulated exposure. Example use shown in the WFSS notebook.
...
+
+
+## Current Development Team
+- Bryan Hilbert
+- Johannes Sahlmann
+- Lauren Chambers
+- Kevin Volk
+- Shannon Osborne
+- Marshall Perrin
+
```
laurenmarietta: Like Matt did for JWQL, it might be nice to include links to the GitHub profiles for each person on the team?

LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
```
laurenmarietta: A simple question: should this file have an extension? i.e. should it be `LICENSE.md` or `LICENSE.txt` rather than just `LICENSE`?

README.md
```diff
@@ -230,4 +230,45 @@ simulated exposure. Example use shown in the WFSS notebook.
```
laurenmarietta: And while you're at it, I just noticed a typo in the definition of raw MIRAGE output files - "calirbations".

LICENSE
KevinVolkSTScI: I suggest that it be named LICENSE.md to be similar to the CODE_OF_CONDUCT.md file name (although I do not know what ".md" is supposed to signify).

bhilbert4: There was just a discussion about this in the `spacetelescope:style-guide` repo. Apparently `LICENSE` is preferred because the file contains no code of any kind (including markdown, which is what .md stands for). It sounds like the other acceptable option is `LICENSE.txt`.

KevinVolkSTScI:
As per Bryan's e-mail that came in while I was writing the review text, if LICENSE is preferred to LICENSE.md because there is no code in the file, then feel free to ignore that part of the review.

bhilbert4:
For versioning, I've added this line to `style_guide.md`. Should I add anything else? 

Currently the version number is set in setup.py. Updating the version number should be one of the last things you change prior to merging a feature branch into `master`. The branch's author and reviewer should agree on the version number prior to merging.

As for having paths in the repo, as I understand it, placing them in the repo is a security risk. This is part of the reason why we switched over to using the MIRAGE_DATA environment keyword. That way example yaml files in the repo can contain the abbreviated paths. As for folks internal to STScI, I have been giving them the path to the `mirage` reference files that are sitting on the central store. I wasn't planning on putting that path in the documentation. @mperrin, what are your thoughts on this? For the `jwql` project we have been instructed not to publish full path names anywhere.

README.md
bhilbert4: Good idea! I've added them.

README.md
bhilbert4: Fixed. Thanks.

KevinVolkSTScI:
I was thinking of whether we may want to put a file path for STScI users in the documentation, or as a comment in the code, for getting the PSF libraries.  I was never intending that a path would be part of the active code.  We may wish to say "STScI users set $MIRAGE_DATA to /some/file/path" somewhere in the documentation and in the readme file.

bhilbert4:
I agree that no paths should be, or are, in the actual code anywhere. But my thought is that if having full path names in the repo is a security risk, then having them in the documentation is also a no-no. But it does mean that internal STScI users will have to rely on personal communication to get the location of the local reference files. That's not an ideal answer either.

mperrin:
**Paths:** I think putting it in the documentation and README makes a lot of sense, agreed. I don't think there is any substantial security concern there. `jwql` is a slightly different case, since web application internal paths can have more security implications if they provide information on internal interfaces that might be accessible via HTTP to hypothetical attackers. A plain directory path name on Central Store doesn't raise that concern. Any attacker would need to already be on the internal network to be able to do anything with that information, and all it would get them is a bunch of FITS files rather than executable code to exploit. 

**License file name:** Over in the style-guides repo, Arfon said that either LICENSE or LICENSE.txt is preferred when there's not actually any markdown syntax in the license file. Personally I don't like bare files with no extension indicating file type so I'd vote for LICENSE.txt in preference over LICENSE. 

**Version numbers:** Version numbers and tags aren't necessary to be set for every individual branch or PR. That's mostly intended for public releases of the code for others to use. E.g. "This is version 1.0... Now it's several months later and here's version 1.1" etc.  I do not see value in trying to increment these for internal development work; that's just not how most projects work. It gets updated by the project leader or release manager shortly before packaging things up as a new release. (for delivery to users via PyPI and Conda, and so on).

KevinVolkSTScI:
How do we keep track of the version numbers that we use in setup.py?  The main version number is not so big an issue as I assume we would need internal discussion before the main version number of the code is changed, but the second and third level version numbers need to be assigned in some way.  Also, do we number these from zero or from one?

LICENSE
bhilbert4: Changed to LICENSE.txt

mperrin:
See https://semver.org for recommended best practices on assigning version numbers and their meanings.   (WebbPSF doesn't follow this model strictly, for historical reasons, but if I were starting from scratch now I probably would.)

bhilbert4:
I added `txt` to the `LICENSE` file.

bhilbert4:
But in the model of CI, it seems like you would need to increment the version number for each feature branch merge, right? That way if someone installs the code at some point and finds a problem, they can say "I have version 3.2.1", and the developers can say "We fixed that problem 2 weeks later, in version 3.2.2." Or maybe you don't worry about version numbers in that case and tell the user "We fixed that problem recently. Download the latest version of the code and your problem will be fixed."

laurenmarietta:
Regarding the versioning question: see issue #133.

Johannes has mentioned in the past that he used the `relic` package to implement versioning with `pysiaf`. I'm not wholly certain how it works, but it seems check the current version on GitHub to automatically determine the version number in a way that is helpful. Here is a link to the `setup.py` for `pysiaf` that includes `relic`:
https://github.com/spacetelescope/pysiaf/blob/master/setup.py 

Again comparing to the JWQL project: there, a new version is "released" basically every time a substantive new PR is merged. This means that, even though the project isn't public yet, we are on version 0.13.1. I don't personally feel like there's much harm to updating the versions this frequently, as it is a good way to track the development history of a project. But I also here Marshall in that I'm not sure there's necessarily a lot to be gained from that method, either...

(Also, maybe future discussion about versioning should take place on the issue I linked above, so as to not further crowd this PR?)

bhilbert4:
When Johannes mentioned `Relic`  in #133, I looked it up, and figured that at some point we could implement that as well for `mirage`.  In my very quick read at the time it wasn't obvious how to do it, and it was low priority. Seems like it's worth a better read. I do agree that any work on that should be a separate PR. For the purposes of this PR, and since we still don't have a lot of people contributing code, should we just leave things a little vague for the time being? Keep our definition of the version numbers (which agrees with recommended practices I believe), but not specify details on the method we use to increment them?

bhilbert4:
Any further thoughts on this? If everyone is happy with having a more detailed versioning discussion in another PR, is this ready to merge?

PR #155, opened 2018-08-27
Update file naming scheme to match JWST naming conventions

bhilbert4:
This is a small update to `yaml_generator.py` and `apt_inputs.py` that updates the file naming convention of the outputs (both yaml file names, as well as the requested `mirage` output fits filename) to match that of JWST.

See https://jwst-docs.stsci.edu/display/JDAT/File+Naming+Conventions+and+Data+Products for details.

Filenames produced by `mirage` should now match those produced by the JWST pipeline.

Note that at the moment this has only been tested on NIRCam, as the yaml_generator only currently supports NIRCam and one engineering-only NIRISS mode.

### DIFF
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -312,7 +312,7 @@ class AptInput:
                         obsnum, visitnum = v.split(':')
                         obsnum = str(obsnum).zfill(3)
                         visitnum = str(visitnum).zfill(3)
-                        if skip == True:
+                        if skip is True:
                             print('Skipping observation {} ({})'.format(obsnum, obslabel))
 
                     try:
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -334,7 +334,8 @@ class AptInput:
                             observation_label.append(obslabel)
                             observation_number.append(obsnum)
                             visit_number.append(visitnum)
-                            vid = "{}{}{}".format(str(propid).zfill(5), obsnum, visitnum)
+                            prop_5digit = "{0:05d}".format(int(propid))
+                            vid = "{}{}{}".format(prop_5digit, obsnum, visitnum)
                             visit_id.append(vid)
                             # Visit group hard coded to 1. It's not clear how APT divides visits up into visit
                             # groups. For now just keep everything in a single visit group.
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -334,7 +334,7 @@ class AptInput:
                             observation_label.append(obslabel)
                             observation_number.append(obsnum)
                             visit_number.append(visitnum)
-                            vid = str(propid) + obsnum + visitnum
+                            vid = "{}{}{}".format(str(propid).zfill(5), obsnum, visitnum)
                             visit_id.append(vid)
                             # Visit group hard coded to 1. It's not clear how APT divides visits up into visit
                             # groups. For now just keep everything in a single visit group.
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -334,7 +334,7 @@ class AptInput:
                             observation_label.append(obslabel)
                             observation_number.append(obsnum)
                             visit_number.append(visitnum)
-                            vid = str(propid) + visitnum + obsnum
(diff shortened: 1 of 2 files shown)


mirage/apt/apt_inputs.py
```diff
@@ -334,7 +334,7 @@ def get_pointing_info(self, file, propid):
                             observation_label.append(obslabel)
                             observation_number.append(obsnum)
                             visit_number.append(visitnum)
-                            vid = str(propid) + visitnum + obsnum
+                            vid = str(propid) + obsnum + visitnum
                             visit_id.append(vid)
                             vgrp = '01'
```
laurenmarietta: I am curious why the visit group and the sequence number are hard-coded, while every other parameter is either read from the pointing file or calculated. Could you maybe add a couple inline comments to denote that?

bhilbert4: Added some explanation. The answer is that for the sequence ID,  I'm not sure how to pull this from the xml. Specific parallel modes are mentioned in there (e.g. NIRCam-MIRI-imaging), but no dictionary of what mode relates to what number. Also, these modes were not used in APT at the time I initially wrote apt_inputs.py. For the purposes of simulating data, I figured it wasn't terribly important. For `visit groups`, I believe the values depend on how everything is scheduled/grouped in APT, and I'm not convinced it is doing that correctly at the moment with the limited information in the PPS database. Unless each 'observation' within APT is called a 'visit group' outside of APT? Something to be investigated later. Again for the purposes of creating simulated data I didn't think it was important at the time I was putting apt_inputs.py together.

laurenmarietta: Okay, this all makes sense, thanks for clarifying! 

You're definitely right, from the perspective of generating data with MIRAGE, it doesn't really matter whether things are being "observed" in parallel or prime.

The visit group question is interesting, but also (like you say) not urgent or particularly important.

bhilbert4:
One final tweak, to force the Proposal ID field to be 5 characters in the output filenames. The recent ERS and GTO proposals have 4 digit proposal IDs, but filename convention says that there should be 5, so I updated to pad the output names with a '0' if necessary.

mirage/apt/apt_inputs.py
```diff
@@ -334,7 +334,7 @@ def get_pointing_info(self, file, propid):
                             observation_label.append(obslabel)
                             observation_number.append(obsnum)
                             visit_number.append(visitnum)
-                            vid = str(propid) + obsnum + visitnum
+                            vid = "{}{}{}".format(str(propid).zfill(5), obsnum, visitnum)
```
laurenmarietta: A more "pythonic" way of doing this would be to specify the length of the prop_id in the string formatter. For example:

`'{0:05d}'.format(757)` returns `00757`

bhilbert4: I originally had it that way, but was getting errors that I didn't understand...maybe because after printing the updated value I was intentionally crashing the code, but was within a try statement. Let me try again...

bhilbert4: Slightly modified version. `propid` is already a string coming out of the xml. For some reason I couldn't combine the `0:05d` format code with the others, like this:
`vid = "{0:05d}{}{}".format(int(propid), obsnum, visitnum)`
I had to have it on its own separate line to avoid an error.

bhilbert4:
@laurenmarietta, it seems like there was some kind of Github or browser hiccup. Your review has disappeared, and reloading this page failed to show my latest commits for 10 minutes or so. I'm not sure if my responses to your comment ever reached you. Anyway, I updated the method used to force a 5 digit proposal ID. Turns out `propid` was already a string at the point that things were being combined to create the filename. So I accounted for that. Also, python didn't like me combining the `{0:05d}` format with other strings, like this:

vid = "{0:05d}{}{}".format(int(propid), obsnum, visitnum)

For some reason I had to put the `{0:05d}' on it's own line to work without error.

laurenmarietta:
That's very strange - I can see the commits you made and the comments, too. Did yours end up fixing itself?

As for the commits, they look good. It's also strange that you couldn't use that string formatting with the rest of the string, but your method works fine.

bhilbert4:
The page eventually fixed itself. I'm still not sure what was going on, but I'm just happy my changes weren't lost.

Thanks! I'll merge this now.

PR #156, opened 2018-08-28
Pysiaf integration

bhilbert4:
NOTE: Re-opening this PR. An error broke the previous PR before it was merged.

This PR introduces code to use pysiaf when calculating things like subarray sizes and positions, and for translating between detector pixels and RA, Dec. This will eliminate the dependence on the subarray definition file in the config directory except for 2 pieces of information: the number of amplifiers used to read out a particular aperture, as well as the OPGS (I think?) name for the aperture. This name is needed only to populate metadata in the output file. The number of amplifiers used is needed in order to get an accurate result for the calculation of the readout time. 

CRDS formatted distortion reference files will still be used for RA, Dec -> pixel space coordinate transforms. This is because in the future these files will be the only source for filter-dependent distortion solutions. If these files are not present, then the code will fall back onto `pysiaf` for these transformations. 

I think this is ready for a final review. Does anyone have time to grab the branch and run a test or two? Everything works for me, but it'd be nice to have confirmation. @KevinVolkSTScI? @laurenmarietta?


laurenmarietta:
As I'm running these tests, I'm realizing that many of them (e.g. `test_RunNIRCamAPTTemplates` in `test_apt.py` and `test_fgs_imaging.py`) are marked with `@pytest.mark.xfail` such that if they fail, that doesn't show up in the CI. However I am now discovering bugs that cause these tests to fail, which were not evident in the CI. Do we want to keep those xfail decorators...?

mirage/utils/siaf_interface.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    # get SIAF
+    siaf = pysiaf.Siaf(instrument)
+
+    # get master aperture names
+    siaf_detector_layout = iando.read.read_siaf_detector_layout()
+    master_aperture_names = siaf_detector_layout['AperName'].data
+
+    # read pysiaf aperture definition file containing DetRef and SciRef values
+    siaf_aperture_definitions = iando.read.read_siaf_aperture_definitions(instrument)
```
laurenmarietta: This line returns the error: 
```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/lchambers/miniconda3/envs/OTECommSims/lib/python3.5/site-packages/pysiaf/source_data/FGS/fgs_siaf_aperture_definition.txt'
```
But I'm not sure how the `pysiaf.iando` module works, so I'm not sure how to fix it. Is this supposed to be pointing to a aperture definition file under `$MIRAGE_DATA`, rather than under my local package installation?

mirage/seed_image/catalog_seed_image.py
```diff
@@ -2541,7 +2328,7 @@ def makeGalaxyImage(self, file, psf):
 
         # create attitude matrix so we can calculate the North->V3 angle for
         # each galaxy
-        attitude_matrix = self.getAttitudeMatrix()
+        # attitude_matrix = self.getAttitudeMatrix()
```
laurenmarietta: Below, on line 2344, the `attitude_matrix` variable is being called but its declaration is being commented out here, so that results in an error. Should the `attitude_matrix` be coming from somewhere else?

mirage/utils/siaf_interface.py
bhilbert4: Try updating your pysiaf installation. Johannes just released a new version with the FGS file last week.

mirage/seed_image/catalog_seed_image.py
bhilbert4: Good catch. That should be `self.attitude_matrix`. I must not have tested with a galaxy input file.

bhilbert4:
As for the tests decorated with `@pytest.mark.xfail`, it would be good to have these as useful tests whose results we pay attention to. I'll admit I haven't looked very closely at them since they were written. I see now that the output name of the yaml files will have to be changed in order to agree with the updated file naming convention, and with this update we should be able to remove the input siaf file as well. Have you found many other bugs? Should we put this in a new PR, or are there few enough changes that we should just bundle it in here?

mirage/yaml/yaml_generator.py
```diff
@@ -702,18 +708,18 @@ def make_start_times(self):
...
-            xe = subarray_def['xend'][match][0]
-            ys = subarray_def['ystart'][match][0]
-            ye = subarray_def['yend'][match][0]
-            xd = xe - xs + 1
-            yd = ye - ys + 1
-            frametime = self.calcFrameTime(xd, yd, amp)
+            siaf_inst = self.info['Instrument'][i].upper()
+            if siaf_inst == 'NIRCAM':
+                siaf_inst = "NIRCam"
+            siaf_obj = pysiaf.Siaf(siaf_inst)[aperture]
```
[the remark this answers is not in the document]
laurenmarietta: Do you think adding this if statement is an acceptable way to solve the problem?

laurenmarietta:
Thanks for these fixes. I am still running into bugs that seem like they need to be addressed before this PR is merged - at the moment the only test script I can run without problems is `test_niriss.py`, for some reason. I'll working on figuring out what is happening more first thing next week.

I definitely think that adding the tests into CI (and modifying however they need to be modified) would be a task for a separate PR!

mirage/yaml/yaml_generator.py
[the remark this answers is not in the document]
bhilbert4: Ah ok, I see the problem now. It should be safe to use `[aperture[0]]` in all cases. The only way that `aperture` should be a list with more than one element is if there are duplicate entries in the subarray definition file.

bhilbert4:
I made a couple small changes, and now the only tests that fail for me are `test_nircam_imaging.py` and `test_fgs_imaging.py`. There's something odd going on with the TEST_DATA environment variable. Somehow it always seems to end with 'NIRISS', which is why `test_nirisis_imaging.py` passes while the others fail.

Somehow, the following test fails:

```
os.environ['TEST_DATA'] = os.path.join(os.path.dirname(__file__), 'test_data/')

def test_env_var():
    full_dir = os.path.expandvars('$TEST_DATA')
    assert 'NIRISS' not in full_dir
```

bhilbert4:
After a lot of trial and error, I was able to get things so that all of the tests passed. I'm still not sure why the `TEST_DATA` environment variable was getting stuck with NIRISS in the name. I updated the tests so that they use `TEST_FGS_DATA`, `TEST_NIRISS_DATA` and `TEST_NIRCAM_DATA` instead. I also fixed the `aperture` issue in `yaml_generator` so that `aperture` is now always a string.

bhilbert4:
I should also mention that now that the tests are passing, I removed the `xfail` decorators.

tests/test_apt.py
```diff
@@ -64,7 +63,7 @@ def RunAllAPTTemplates(instrument):
     yam = yaml_generator.SimInput()
     yam.input_xml = xml_file
     yam.pointing_file = pointing_file
-    yam.siaf = os.path.expandvars('$MIRAGE_DATA/nircam/reference_files/SIAF/NIRCam_SIAF_2018-01-08.csv')
+    # yam.siaf = os.path.expandvars('$MIRAGE_DATA/nircam/reference_files/SIAF/NIRCam_SIAF_2018-01-08.csv')
     yam.output_dir = out_dir
     yam.simdata_output_dir = out_dir
     yam.observation_table = observationlist_file
```
laurenmarietta: I ran this on my branch, which has master pulled in, and got an error from the assert statement in line 81. The glob there is still looking for the old YAML filenames. If you replace it with this, it will work:
```
assert len(glob.glob(os.path.join(out_dir, 'jw{0:05d}*.yaml'.format(int(PROPOSAL_ID))))) \
        >= n_obs, 'Fewer yaml files created than observations'
```

bhilbert4: My intuition (which could be wrong) is that this will be fixed automatically when the `pysiaf-integration` is merged into `master`, right?

laurenmarietta: Ohhh, I guess so! I thought it might throw a merge conflict but you're right, since that part of the file hasn't been changed. Thanks for pointing that out.

laurenmarietta: Oh wait, actually no, because this is in `test_apt.py` and I don't think this particular assert statement was changed on `master`. If it were, I wouldn't have had the error (since I have already pulled master into my local version of this branch). Or am I still confused?

bhilbert4: Huh, that's odd. How was this test passing for me yesterday and today? Ah, I bet the yaml files in my tests were still specifying filenames that start with V. I'll update the code with the line you provided.

bhilbert4: Ok, I swapped out the old line for yours above. I had to rebase and then force push in order to get the other name convention changes, which then allowed all of the pytests to pass.

bhilbert4: Now Travis is failing on the check to see if the MIRAGE_DATA environment variable is set. I guess it makes sense that it fails here but not in my local check though.

bhilbert4: Re-applied the `@pytest.mark.xfail` decorator to the check of the MIRAGE_DATA environment variable.

bhilbert4:
Hmmm. `test_apt.py` calls `yaml_generator.py`, which does actually need a real value for the MIRAGE_DATA environment variable, in order to populate the names of various reference files in the yaml outputs. I'm not sure how to do this in terms of the Travis run then.

KevinVolkSTScI:
I ran a set of NIRISS models with a single source at the defined (RA,Dec) position given in the .yaml file, for all the 30 NIRISS sub-arrays.  What should happen, and what was seen with the "old" niriss_subarrays.list file, is that the star should appear in all of the sub-array images but one at the defined reference position.  The only case where that star will not be visible is for SUBAMPCAL where the reference position is outside the sub-array.

What I see from the output is that the star is not in the subarrays when these are smaller than the full frame.  All the full frame tests have the star in the center of the array as expected.  I need to run tests with more stars in the field to try to see what is happening.

KevinVolkSTScI:
Bryan comments that the number of amplifiers needs to be read in the from niriss_subarrays.list file or the equivalent subarray list files for the other instruments.  I was told that the number of amplifiers is determined by the number of pixels in the fast read direction.  If the value is 2048 then 4 amplifiers are used, otherwise 1 amplifier is used.  If I am correct about this, one could determine the number of amplifiers from the sub-array dimensions as long as the fast read direction is known for each instrument.  I think the fast read direction is the same pixel axis for both Guider detectors.  If that is also true for all the NIRCam detectors one could in principle only use the subaray list file to get the names of the subarrays in the data management system.
If only the subarray name matching is done via the subrray list files one could just move these into the code as they are fixed values that are not expected to change.  But this is a matter of taste not of substance.

bhilbert4:
Unfortunately NIRCam doesn't follow the same convention with respect to fast read direction size and number of amplifiers. We have several subarrays that span the entire width of the detector (in the fast read direction) and the user can specify whether they want to use 1 amp or all 4. This case is the driver for keeping the number of amplifiers in the subarray definition file. 

I'll look into the source positions on the subarrays on Monday. If I recall correctly my NIRCam test was successful.

KevinVolkSTScI:
I looked at some of the star positions from one sub-array to another in my NIRISS tests, and it appears that the reference position is being changed from one sub-array to the next.  However, I have not been able to make sense of the changes in position with respect to what was in the niriss_subarrays.list file generated by me from the SIAF file.  It looks like the star positions in one sub-array with respect to another are distorted, so the offsets from one sub-array to another are not quite matching the SIAF values (i.e. where the SIAF defines and offset of 28 pixels between AMI1 and AMI2 the offset listed is 27.812 pixels, for example).  The code is not putting the target star at the reference sky position in the right place.  It is present in 12 of 30 sub-arrays, those being full frame read-out instances with different reference positions, but is not in any of the sub-array images.  It looks like there are two problems:  it looks like the reference position is not being placed in the sub-array area, and then there is a small distortion issue (although I may be seeing the differential distortion between where the sub-array should be and where it is being placed, if these are significantly different).

laurenmarietta:
I'm finally getting around to testing your new changes here - do each of those new environment variables (TEST_FGS_DATA, TEST_NIRISS_DATA and TEST_NIRCAM_DATA) need to be included in my `.bash_profile` now, for this to work?

laurenmarietta:
Okay, it seems like the MIRAGE_DATA environment variable is getting messed up somehow... if I print `self.data_dir` (in `yaml_generator.py`), I get `/mypath/mirage/refdata/`. It looks like the same thing is somehow happening in all the specific instrument tests, too.

Any idea why this is happening...?

bhilbert4:
@KevinVolkSTScI, I am seeing the same problems with the NIRISS subarrays that you are. I've traced it to the output from the distortion reference file model. Right now I have a subarray where the reference location is set to RA=53.0, Dec=-27.0. My point source list attempts to place a star on the reference location. When I run mirage using the distortion reference file (jwst_niriss_distortion_0008.asdf), I get:

input ra, dec is 53.0, -27.0
v2, v3 is -294.72122299997494, -757.5053950000259
pixelx, pixely from reffile is 1075.957597281965, 111.60209658822146

but when I run the same situation without the reference file, meaning mirage falls back to using pysiaf, I get:

input ra, dec is 53.0, -27.0
v2, v3 is -294.72122299997494, -757.5053950000259
pixelx, pixely from pysiaf is 31.999999999612676, 31.999999999607816

Could it be that we are pointing to an old or out of date reference file? I'll also add that my test of a NIRCam subarray seemed to work fine.

bhilbert4:
@laurenmarietta I set MIRAGE_DATA to a dummy value in `test_apt.py` in an attempt to get the test to pass, if I recall correctly. But in the end I don't think that helped, because `yaml_generator.py` (or `apt_inputs.py`, actually searches in the reference file directory for available files. So I guess I can take that dummy MIRAGE_DATA value out. I wonder if that change was still present during the instrument tests? Were you seeing that same value there too?

KevinVolkSTScI:
Hi Bryan,

  The reference file should be consistent with the SIAF file (Johannes made the last revisions of both files as I recall…).  However, there is always the chance that the reference file in CRDS is somehow out of date.  I can run a test with no pipeline reference file and see if that fixes the issues with the positions.  If that works, then I will try to determine whether the CRDS file is out of date.

Kevin

KevinVolkSTScI:
I have started some systematic tests of the sub-arrays with the use of PYSIAF for the distortion rather than using the pipeline reference file.  So far it all looks good aside from an offset of 0.5 pixels between the SIAF values and the values in the pointsources.list output files, which I assume is expected because of the difference in what is considered the center of the pixel and the 0-based vs. 1-based numbering differences.
We NIRISS people need to investigate why the reference distortion file use produces bad results.  However, I wonder why mirage uses the file at all if the information is in the SIAF file.  The pipeline reference file for NIRISS comes from the SIAF file, and I would assume that this is the case for the other instruments as well, so we should stick with the source and not use a derived file if this is the case.  We have to provide the SIAF file anyway to get the sub-arrays in the right places when using pysiaf.

Johannes-Sahlmann:
@KevinVolkSTScI the SIAF only contains distortion coefficients that meet the requirements for operations. These do not need to be science-grade and have no filter dependence. For best performance in terms of distortion correction, the wcs reference file should be used.

bhilbert4:
@laurenmarietta are you still seeing errors when you try running the tests?

mirage/seed_image/catalog_seed_image.py
```diff
@@ -2033,91 +1893,37 @@ def parseRADec(self, rastr, decstr):
...
 
+        if self.coord_transform is not None:
+            # Use the distortion reference file to translate from V2, V3 to RA, Dec
+            pixelx, pixely = self.coord_transform.inverse(loc_v2, loc_v3)
+        else:
+            pixelx, pixely = self.siaf.tel_to_sci(loc_v2, loc_v3)
+            # Subtract 1 from SAIF-derived results since SIAF works in a 1-indexed coord system
+            pixelx -= 1
+            pixely -= 1
+        stophere
```
laurenmarietta: This "stophere" line is throwing an error.

KevinVolkSTScI: I did not run into this in any of the tests I made, which seems odd as I had an input point source list with many objects in it.

bhilbert4: Oops. Sorry. That was leftover from investigating NIRISS issues.

bhilbert4:
When I just pushed the update to `setup.py` I saw the Travis banner appear for just a second, and it then disappeared and was replaced with the mention of the conflict. So maybe the Travis run is being skipped.

PR #161, opened 2018-09-07
Changes to add the NRM mode of NIRISS as an option.

KevinVolkSTScI:
I have made some (fairly small) code changes to add the NIRISS AMI mode to mirage.  To use this mode one needs separate PFS files.  I have chosen to change the names of these PSF files from the regular imaging file names by replacing "NIS" by "NIS_NRM".  Only four filters can be used with the NRM for science: F277W, F380M, F430M, and F480M.  I do not have a check of this in the code because the PSF library will only cover the four allowed filters.  The other change is to scale the count rates for a given magnitude by a fixed factor due to the throughput of the NRM compared to the throughput of the CLEARP element that is in the pupil wheel of NIRISS for regular imaging in these filters.  The changes were tested for a simple case of a single star in the field.  More detailed tests need to wait until the PSF library is fully generated.

I also added a test to the tests directory, but I have not tried that out because I have never used pytest.  The steps in the test file were tried out manually and they work, so I think that the test should work OK.  I need to learn how to run the test explicitly....

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -32,7 +32,7 @@ from ..utils import set_telescope_pointing_separated as set_telescope_pointing
 
 INST_LIST = ['nircam', 'niriss', 'fgs']
 MODES = {'nircam': ["imaging", "ts_imaging", "wfss", "ts_wfss"],
-         'niriss': ["imaging"],
+         'niriss': ["imaging","ami"],
          'fgs': ["imaging"]}
 TRACKING_LIST = ['sidereal','non-sidereal']
 inst_abbrev = {'nircam': 'NRC',
@@ -402,21 +402,31 @@ class Catalog_seed():
 
     def mag_to_countrate(self, magsys, mag, photfnu=None, photflam=None):
         # Convert object magnitude to counts/sec
+        #
+        # For NIRISS AMI mode, the count rate values calculated need to be 
+        # scaled by a factor 0.15/0.84 = 0.17857.  The 0.15 value is the 
+        # throughput of the NRM, while the 0.84 value is the throughput of the 
+        # imaging CLEARP element that is in place in the pupil wheel for the 
+        # normal imaging observations.
+        if self.params['Inst']['mode'] in ['ami']:
+            count_scale = 0.15 / 0.84
+        else:
+            count_scale = 1.
         if magsys.lower() == 'abmag':
             try:
-                return 10**((mag + 48.6) / -2.5) / photfnu
+                return count_scale * (10**((mag + 48.599934378) / -2.5) / photfnu)
             except:
                 raise ValueError(("AB mag to countrate conversion failed."
                                   "magnitude = {}, photfnu = {}".format(mag, photfnu)))
         if magsys.lower() == 'vegamag':
             try:
-                return 10**((self.vegazeropoint - mag) / 2.5)
+                return count_scale * (10**((self.vegazeropoint - mag) / 2.5))
             except:
                 raise ValueError(("Vega mag to countrate conversion failed."
                                   "magnitude = {}".format(mag)))
         if magsys.lower() == 'stmag':
             try:
-                return 10**((mag + 21.1) / -2.5) / photflam
+                return count_scale * (10**((mag + 21.099934378) / -2.5) / photflam)
             except:
                 raise ValueError(("ST mag to countrate conversion failed."
                                   "magnitude = {}, photflam = {}".format(mag, photflam)))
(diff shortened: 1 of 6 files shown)


tests/test_niriss_imaging_modes.py
```diff
@@ -0,0 +1,47 @@
...
+
+from mirage import imaging_simulator 
+
+# os.environ['MIRAGE_DATA'] = ''
+os.environ['TEST_DATA'] = os.path.join(os.path.dirname(__file__), 'test_data/NIRISS')
+
+@pytest.mark.xfail
+def test_niriss_imaging():
+    nis = imaging_simulator.ImgSim()
+    nis.paramfile = os.path.join(os.path.dirname(__file__), 'test_data/NIRISS/niriss_imaging_test.yaml')
```
bhilbert4: I think you should be able to make this line:
nis.paramfile = os.path.expandvars('$TEST_DATA/niriss_imaging_test.yaml')

tests/test_niriss_imaging_modes.py
```diff
@@ -0,0 +1,47 @@
...
+    nis.create()
+    nis.paramfile = os.path.join(os.path.dirname(__file__), 'test_data/NIRISS/niriss_nrm_test.yaml')
```
bhilbert4: And similarly here:
nis.paramfile = os.path.expandvars('$TEST_DATA/niriss_nrm_test.yaml')

mirage/seed_image/catalog_seed_image.py
```diff
@@ -3179,6 +3189,12 @@ def checkParams(self):
 
                 self.params['simSignals']['psfpath'] = os.path.join(self.params['simSignals']['psfpath'], pathaddition)
                 self.psfname = os.path.join(self.params['simSignals']['psfpath'], psfname)
+                # In the NIRISS AMI mode case, replace NIS by NIS_NRM as the PSF 
+                # files are in a separate directory and have the altered file names
+                # compared to imaging.  Hence one can point to the same base 
+                # directory to run both imaging and NRM models.
+                if self.params['Inst']['mode'] in ['ami']:
+                    self.psfname = self.psfname.replace('NIS','NIS_NRM')
```
bhilbert4: I'm a little concerned about getting the PSF filename this way. This assumes that the NRM files are in a particular location relative to the imaging files. What if someone creates a new or specialized NRM library of some sort that they want to test? (I'm not sure how realistic that situation would be.) 

But, I can see how really this is probably the best solution given the assumptions about filenames and directory structure that mirage makes when constructing the PSF filename. 

My hope is that when the ePSF changes are introduced, that the PSF library input value will be the full path to the directory containing the files (no assumption of subdirectory structure), since there should be many fewer and more easily organized files in that case. So for example, the yaml inputs for imaging and NRM in that case would be something like:
/myfiles/data/psf_libraries/NIRISS/imaging/  and 
/myfiles/data/psf_libraries/NIRISS/NRM/
With this method, there's no built-in assumption of where files are located nor what kind of subdirectories are under the given path.

tests/test_niriss_imaging_modes.py
```diff
@@ -0,0 +1,47 @@
+"""System test of mirage/NIRISS for regular and NRM imaging.
+
+Authors
+-------
+    - Kevin Volk
+
+Use
+---
+    >>> pytest -s test_niriss_imaging_modes.py
```
bhilbert4: You should be able to use this command from the command line in order to run this test using pytest.

tests/test_niriss_imaging_modes.py
```diff
@@ -0,0 +1,47 @@
...
+import os
+import pytest
+import numpy
+
+from mirage import imaging_simulator 
+
+# os.environ['MIRAGE_DATA'] = ''
+os.environ['TEST_DATA'] = os.path.join(os.path.dirname(__file__), 'test_data/NIRISS')
+
+@pytest.mark.xfail
```
bhilbert4: This decorator is for tests that we allow to fail without any consequences (or tests that we know will fail). It seems like this test should pass, and that there should be a warning raised if it does fail, right? If so you should remove this line.

mirage/seed_image/catalog_seed_image.py
KevinVolkSTScI: I did not think about it in these terms, but of course you are right that this requires the specific directory structure for the PSF files.  I think we just need to document it in the code, and assume that when we move to ePSF that this will no longer be an issue.  One already requires the specific directory structure for the PSF libraries, and so extending this one level for imaging versus NRM PSF libraries is not a big change in "philosophy".  And it simplifies using the yaml files as one only needs to change the mode from imaging to AMI and everything works.  Hence I will add some comments to the code but not change the mechanics.

KevinVolkSTScI:
I made some changes to the comments in catalog_seed_image.py.  I also tried running the test.  Some changes were needed to the .yaml files to point to the point source list input files properly, but once that was done the test ran successfully.

KevinVolkSTScI:
I made another small change to clean up the output files from the mode test.

KevinVolkSTScI:
The .yaml files for the mode test have explicit file paths in them, which now that I think about it is something that one is not supposed to do according to the code standards.  I need to figure out how to fix that.

PR #166, opened 2018-09-19
Temporary change to use SIAF pre-delivery data for NIRCam

bhilbert4:
This PR introduces a small, temporary change to use SIAF pre-delivery data for NIRCam only. This is because the correct SIAF distortion coefficients are not yet in the PRD, but with the merge of the pysiaf changes recently (), the manual work-around that had been in `mirage` was removed. So we need access to the correct distortion coefficients in order to have `mirage` continue to place sources in the correct locations on the detector when source positions are given in RA, Dec.

For NIRISS and FGS, `mirage` will continue to use the PRD version of the SIAF data.

### DIFF
--- a/mirage/utils/siaf_interface.py
+++ b/mirage/utils/siaf_interface.py
@@ -22,6 +22,7 @@ import numpy as np
 
 import pysiaf
 from pysiaf import iando
+from pysiaf.constants import JWST_DELIVERY_DATA_ROOT
 from ..utils import rotations
 from ..utils import set_telescope_pointing_separated as set_telescope_pointing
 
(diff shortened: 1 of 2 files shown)


mirage/utils/siaf_interface.py
```diff
@@ -37,7 +38,16 @@ def get_siaf_information(instrument, aperture, ra, dec, telescope_roll):
...
         Aperture name (e.g. "NRCA1_FULL")
     """
-    siaf = pysiaf.Siaf(instrument)[aperture]
+    # Temporary fix to access good NIRCam distortion coefficients which
+    # which are not yet in the PRD
+    if instrument.lower() == 'nircam':
+        import os
+        print("NOTE: Using pre-delivery SIAF data")
+        pre_delivery_dir = os.path.join(JWST_DELIVERY_DATA_ROOT, instrument)
+        siaf = pysiaf.Siaf(instrument, basepath=pre_delivery_dir)[aperture]
```
Johannes-Sahlmann: This will fail when the updated NIRCam SIAF will have been delivered to PRD and the pre_delivery_dir will be empty. I think this is fine, because it will make sure that this code is removed when it is no longer needed.

PR #169, opened 2018-09-28
Fix source filtering for catalogs that span 0.0 RA and/or Dec

bhilbert4:
This PR introduces a small fix to the `remove_outside_fov_sources` function. This function is used to filter out sources that are very far from the detector's field of view. The motivation for this function was that the translation from RA, Dec to detector x,y would hang when attempting to work on a source far from the detector.

This fix affects only source catalogs that span across 0.0 RA and/or Dec. Previously in this case the code would create an upper limit > 360. or a lower limit < 0. The sources' RA, Dec were then compared to these values, resulting in improper filtering. The code now checks to see if RA or Dec crosses 0. and if so, takes steps to filter the sources correctly.

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -1601,6 +1601,8 @@ class Catalog_seed():
         source : Table
             astropy Table containing filtered list of sources
         """
+        x_comparison_operator = 'and'
+        y_comparison_operator = 'and'
         if pixflag:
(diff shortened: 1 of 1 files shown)


bhilbert4:
@Johannes-Sahlmann do you have time to give this a quick check?

bhilbert4:
Good point. I should have thought of that. I've updated the code to use SkyCoord for RA,Dec comparisons.

PR #172, opened 2018-10-01
WIP: Niriss apt interface

Johannes-Sahlmann:
Enable mirage to run on the basis of two APT files (.xml and .pointing) plus a source catalog file.

### DIFF
--- a/mirage/yaml/write_observationlist.py
+++ b/mirage/yaml/write_observationlist.py
@@ -4,6 +4,7 @@ output files
 Authors
 -------
     - Lauren Chambers
+    - Johannes Sahlmann
 
 Use
 ---
@@ -20,25 +21,31 @@ import numpy as np
 from ..apt import read_apt_xml
 
 
-def write_yaml(xml_file, pointing_file, yaml_file, ps_cat_sw=None, ps_cat_lw=None):
-    # Define "default" values (probably should be changed eventually)
-    date = '2019-07-04'
-    PAV3 = '0.'
-    GalaxyCatalog = 'None'
-    ExtendedCatalog = 'None'
-    ExtendedScale = '1.0'
-    ExtendedCenter = '1024,1024'
-    MovingTargetList = 'None'
-    MovingTargetSersic = 'None'
-    MovingTargetExtended = 'None'
-    MovingTargetConvolveExtended = 'True'
-    MovingTargetToTrack = 'None'
-    BackgroundRate_sw = '0.5'
-    BackgroundRate_lw = '1.2'
+def write_yaml(xml_file, pointing_file, yaml_file, catalog_files, ps_cat_sw=None, ps_cat_lw=None):
+    """Write observation list file (required miareg input) on the basis of APT files.
+
+    Parameters
+    ----------
+    xml_file : str
+        path to APT .xml file
+    pointing_file
+    yaml_file
+    catalog_files : str or list(str)
+        path to file(s) that contain the catalog of sources.
+    ps_cat_sw
+    ps_cat_lw
+
+    Returns
+    -------
+
+    """
 
     # Read in observations from XML file
     with open(xml_file) as f:
         tree = etree.parse(f)
+
+    if type(catalog_files) is str:
+        catalog_files = [catalog_files]
 
     apt = '{http://www.stsci.edu/JWST/APT}'
     observation_data = tree.find(apt + 'DataRequests')
@@ -48,10 +55,13 @@ def write_yaml(xml_file, pointing_file, yaml_file, ps_cat_sw=None, ps_cat_lw=Non
     observations = []
     i_observations = []
     obs_names = []
+    instruments = []
     for i_obs, o in enumerate(obs_results):
-        if o.find(apt + 'Instrument').text in ['NIRCAM', 'WFSC']:
+        instrument = o.find(apt + 'Instrument').text
+        if instrument in ['NIRCAM', 'WFSC', 'NIRISS', 'FGS']:
             observations.append(o)
             i_observations.append(i_obs)
+            instruments.append(instrument)
 
             # Find observation name, and modify to exclude parantheticals
             label = o.find(apt + 'Label').text
@@ -65,26 +75,84 @@ def write_yaml(xml_file, pointing_file, yaml_file, ps_cat_sw=None, ps_cat_lw=Non
     readxml_obj = read_apt_xml.ReadAPTXML()
     xml_table = readxml_obj.read_xml(xml_file)
 
-    sw_filters = {}
-    lw_filters = {}
-    sw_filters_all = np.array(xml_table['ShortFilter'])
-    lw_filters_all = np.array(xml_table['LongFilter'])
-    tile_nums = xml_table['TileNumber']
-    observation_ids = xml_table['ObservationID']
+    # array of unique instrument names
+    used_instruments =  np.unique(instruments)
 
-    for i_obs_all in set(observation_ids):
-        # i_obs_all = int(i_obs_all)
-        current_obs_indices = [i == i_obs_all for i in observation_ids]
-        if len(set(np.array(sw_filters_all)[current_obs_indices])) > 1:
-            print('Note: Multiple filters in observation {}'.format(i_obs_all))
-            # At some point could use the tile_nums to fix this
-        sw_filters[i_obs_all] = sw_filters_all[current_obs_indices]
-        lw_filters[i_obs_all] = lw_filters_all[current_obs_indices]
+    if len(used_instruments) > 1:
+        # Several instruments used within one APT file, e.g. parallels
+        raise NotImplementedError
+    else:
+        if used_instruments[0] in ['NIRCAM', 'WFSC']:
+            # Define "default" values (probably should be changed eventually)
+            date = '2019-07-04'
+            PAV3 = '0.'
+            GalaxyCatalog = 'None'
+            ExtendedCatalog = 'None'
+            ExtendedScale = '1.0'
+            ExtendedCenter = '1024,1024'
+            MovingTargetList = 'None'
+            MovingTargetSersic = 'None'
+            MovingTargetExtended = 'None'
+            MovingTargetConvolveExtended = 'True'
+            MovingTargetToTrack = 'None'
+            BackgroundRate_sw = '0.5'
+            BackgroundRate_lw = '1.2'
 
-    # Check that all parameters have the right length
-    all_param_lengths = [len(ps_cat_sw), len(ps_cat_lw), len(sw_filters),
-                         len(lw_filters), len(observations), len(i_observations),
-                         len(obs_names)]
+            sw_filters = {}
+            lw_filters = {}
+            sw_filters_all = np.array(xml_table['ShortFilter'])
+            lw_filters_all = np.array(xml_table['LongFilter'])
+            tile_nums = xml_table['TileNumber']
+            observation_ids = xml_table['ObservationID']
+
+            for i_obs_all in set(observation_ids):
+                # i_obs_all = int(i_obs_all)
+                current_obs_indices = [i == i_obs_all for i in observation_ids]
+                if len(set(np.array(sw_filters_all)[current_obs_indices])) > 1:
+                    print('Note: Multiple filters in observation {}'.format(i_obs_all))
+                    # At some point could use the tile_nums to fix this
+                sw_filters[i_obs_all] = sw_filters_all[current_obs_indices]
+                lw_filters[i_obs_all] = lw_filters_all[current_obs_indices]
+
+            # Check that all parameters have the right length
+            all_param_lengths = [len(ps_cat_sw), len(ps_cat_lw), len(sw_filters),
+                                 len(lw_filters), len(observations), len(i_observations),
+                                 len(obs_names)]
+
+        elif used_instruments[0] in ['NIRISS']:
+            # Define "default" values (probably should be changed eventually)
+            date = '2019-07-04'
+            PAV3 = '0.'
+            GalaxyCatalog = 'None'
+            ExtendedCatalog = 'None'
+            ExtendedScale = '1.0'
+            ExtendedCenter = '1024,1024'
+            MovingTargetList = 'None'
+            MovingTargetSersic = 'None'
+            MovingTargetExtended = 'None'
+            MovingTargetConvolveExtended = 'True'
+            MovingTargetToTrack = 'None'
+            BackgroundRate = '0.5'
+
+            filters = {}
+            filters_all = np.array(xml_table['PupilWheel'])
+            # lw_filters_all = np.array(xml_table['LongFilter'])
+            tile_nums = xml_table['TileNumber']
+            observation_ids = xml_table['ObservationID']
+
+            for i_obs_all in set(observation_ids):
+                # i_obs_all = int(i_obs_all)
+                current_obs_indices = [i == i_obs_all for i in observation_ids]
+                if len(set(np.array(filters_all)[current_obs_indices])) > 1:
+                    print('Note: Multiple filters in observation {}'.format(i_obs_all))
+                    # At some point could use the tile_nums to fix this
+                filters[i_obs_all] = filters_all[current_obs_indices]
+                # lw_filters[i_obs_all] = lw_filters_all[current_obs_indices]
+
+            # Check that all parameters have the right length
+            all_param_lengths = [len(catalog_files), len(filters), len(observations),
+                                 len(i_observations), len(obs_names)]
+
     if len(set(all_param_lengths)) > 1:
         print(all_param_lengths)
         raise ValueError('Not all provided parameters have compatible '
@@ -93,46 +161,77 @@ def write_yaml(xml_file, pointing_file, yaml_file, ps_cat_sw=None, ps_cat_lw=Non
     write = ["# Observation list created by write_observationlist.py  in\n",
              "# mirage/yaml. Note: all values except filters and\n",
              "# observation names are default.\n\n"]
-    for i_obs in range(num_obs):
-        obs_number = i_observations[i_obs] + 1
-        write += [
-            "Observation{}:\n".format(obs_number),
-            "  Name: '{}'\n".format(obs_names[i_obs]),
-            "  Date: {}\n".format(date),
-            "  PAV3: {}\n".format(PAV3),
-        ]
 
-        for i_filt, (sw_filt, lw_filt) in enumerate(zip(sw_filters[obs_number],
-                                                        lw_filters[obs_number])):
+    if used_instruments[0] in ['NIRCAM', 'WFSC']:
+        for i_obs in range(num_obs):
+            obs_number = i_observations[i_obs] + 1
             write += [
-                "  FilterConfig{}:\n".format(i_filt + 1),
-                "    SW:\n",
-                "      Filter: {}\n".format(sw_filt),
-                "      PointSourceCatalog: {}\n".format(ps_cat_sw[i_obs]),
-                "      GalaxyCatalog: {}\n".format(GalaxyCatalog),
-                "      ExtendedCatalog: {}\n".format(ExtendedCatalog),
-                "      ExtendedScale: {}\n".format(ExtendedScale),
-                "      ExtendedCenter: {}\n".format(ExtendedCenter),
-                "      MovingTargetList: {}\n".format(MovingTargetList),
-                "      MovingTargetSersic: {}\n".format(MovingTargetSersic),
-                "      MovingTargetExtended: {}\n".format(MovingTargetExtended),
-                "      MovingTargetConvolveExtended: {}\n".format(MovingTargetConvolveExtended),
-                "      MovingTargetToTrack: {}\n".format(MovingTargetToTrack),
-                "      BackgroundRate: {}\n".format(BackgroundRate_sw),
-                "    LW:\n",
-                "      Filter: {}\n".format(lw_filt),
-                "      PointSourceCatalog: {}\n".format(ps_cat_lw[i_obs]),
-                "      GalaxyCatalog: {}\n".format(GalaxyCatalog),
-                "      ExtendedCatalog: {}\n".format(ExtendedCatalog),
-                "      ExtendedScale: {}\n".format(ExtendedScale),
-                "      ExtendedCenter: {}\n".format(ExtendedCenter),
-                "      MovingTargetList: {}\n".format(MovingTargetList),
-                "      MovingTargetSersic: {}\n".format(MovingTargetSersic),
-                "      MovingTargetExtended: {}\n".format(MovingTargetExtended),
-                "      MovingTargetConvolveExtended: {}\n".format(MovingTargetConvolveExtended),
-                "      MovingTargetToTrack: {}\n".format(MovingTargetToTrack),
-                "      BackgroundRate: {}\n\n".format(BackgroundRate_lw)
+                "Observation{}:\n".format(obs_number),
+                "  Name: '{}'\n".format(obs_names[i_obs]),
+                "  Date: {}\n".format(date),
+                "  PAV3: {}\n".format(PAV3),
             ]
+
+            for i_filt, (sw_filt, lw_filt) in enumerate(zip(sw_filters[obs_number],
+                                                            lw_filters[obs_number])):
+                write += [
+                    "  FilterConfig{}:\n".format(i_filt + 1),
+                    "    SW:\n",
+                    "      Filter: {}\n".format(sw_filt),
+                    "      PointSourceCatalog: {}\n".format(ps_cat_sw[i_obs]),
+                    "      GalaxyCatalog: {}\n".format(GalaxyCatalog),
+                    "      ExtendedCatalog: {}\n".format(ExtendedCatalog),
+                    "      ExtendedScale: {}\n".format(ExtendedScale),
+                    "      ExtendedCenter: {}\n".format(ExtendedCenter),
+                    "      MovingTargetList: {}\n".format(MovingTargetList),
+                    "      MovingTargetSersic: {}\n".format(MovingTargetSersic),
+                    "      MovingTargetExtended: {}\n".format(MovingTargetExtended),
+                    "      MovingTargetConvolveExtended: {}\n".format(MovingTargetConvolveExtended),
+                    "      MovingTargetToTrack: {}\n".format(MovingTargetToTrack),
+                    "      BackgroundRate: {}\n".format(BackgroundRate_sw),
+                    "    LW:\n",
+                    "      Filter: {}\n".format(lw_filt),
+                    "      PointSourceCatalog: {}\n".format(ps_cat_lw[i_obs]),
+                    "      GalaxyCatalog: {}\n".format(GalaxyCatalog),
+                    "      ExtendedCatalog: {}\n".format(ExtendedCatalog),
+                    "      ExtendedScale: {}\n".format(ExtendedScale),
+                    "      ExtendedCenter: {}\n".format(ExtendedCenter),
+                    "      MovingTargetList: {}\n".format(MovingTargetList),
+                    "      MovingTargetSersic: {}\n".format(MovingTargetSersic),
+                    "      MovingTargetExtended: {}\n".format(MovingTargetExtended),
+                    "      MovingTargetConvolveExtended: {}\n".format(MovingTargetConvolveExtended),
+                    "      MovingTargetToTrack: {}\n".format(MovingTargetToTrack),
+                    "      BackgroundRate: {}\n\n".format(BackgroundRate_lw)
+                ]
+
+    elif used_instruments[0] in ['NIRISS', 'FGS']:
+        for i_obs in range(num_obs):
+            obs_number = i_observations[i_obs] + 1
+            write += [
+                "Observation{}:\n".format(obs_number),
+                "  Name: '{}'\n".format(obs_names[i_obs]),
+                "  Date: {}\n".format(date),
+                "  PAV3: {}\n".format(PAV3),
+            ]
+
+            # for i_filt, filt in enumerate(filters[obs_number]):
+            for i_filt, filt in enumerate(filters[i_obs]):
+                write += [
+                    # "  FilterConfig{}:\n".format(i_filt + 1),
+                    # "    SW:\n",
+                    "  Filter: {}\n".format(filt),
+                    "  PointSourceCatalog: {}\n".format(catalog_files[i_obs]),
+                    "  GalaxyCatalog: {}\n".format(GalaxyCatalog),
+                    "  ExtendedCatalog: {}\n".format(ExtendedCatalog),
+                    "  ExtendedScale: {}\n".format(ExtendedScale),
+                    "  ExtendedCenter: {}\n".format(ExtendedCenter),
+                    "  MovingTargetList: {}\n".format(MovingTargetList),
+                    "  MovingTargetSersic: {}\n".format(MovingTargetSersic),
+                    "  MovingTargetExtended: {}\n".format(MovingTargetExtended),
+                    "  MovingTargetConvolveExtended: {}\n".format(MovingTargetConvolveExtended),
+                    "  MovingTargetToTrack: {}\n".format(MovingTargetToTrack),
+                    "  BackgroundRate: {}\n".format(BackgroundRate),
+                ]
 
     f = open(yaml_file, 'w')
     for line in write:
--- a/mirage/yaml/yaml_generator.py
+++ b/mirage/yaml/yaml_generator.py
@@ -870,7 +870,7 @@ class SimInput:
             self.configfiles['filter_throughput'] = 'placeholder.txt'
         elif self.instrument.lower() == 'niriss':
             self.reference_file_dir = os.path.join(self.datadir, 'niriss/reference_files')
-            self.psfpath = os.path.join(self.datadir, 'niriss/webbpsf_files')
+            self.psfpath = os.path.join(self.datadir, 'niriss/webbpsf_library')
             self.psfbasename = 'niriss'
             self.psfpixfrac = 0.1
             self.configfiles['subarray_def_file'] = 'niriss_subarrays.list'
@@ -1115,11 +1115,15 @@ class SimInput:
             f.write('  library: SUNMIN    # Type of cosmic rayenvironment (SUNMAX, SUNMIN, FLARE)\n')
             f.write('  scale: 1.5     # Cosmic ray scaling factor\n')
             # temporary tweak here to make it work with NIRISS
+            detector_label = input['detector']
+
             if instrument.lower() in ['nircam', 'wfsc']:
-                detector_label = input['detector']
+                # detector_label = input['detector']
+                f.write('  suffix: IPC_NIRCam_{}    # Suffix of library file names\n'.format(
+                    detector_label))
             elif instrument.lower() == 'niriss':
-                detector_label = 'B5'
-            f.write('  suffix: IPC_NIRCam_{}    # Suffix of library file names\n'.format(detector_label))
+                f.write('  suffix: IPC_NIRISS_{}    # Suffix of library file names\n'.format(
+                    detector_label))
             f.write('  seed: {}                 # Seed for random number generator\n'.format(np.random.randint(1, 2**32-2)))
             f.write('\n')
             f.write('simSignals:\n')


Johannes-Sahlmann:
This PR addresses some necessary changes to mirage to enable the APT
interface for non-trivial programs (e.g. non-NIRCam, multi-instrument, coordinated
parallel, dithers, or mosaic).  The aim is to enable mirage to run with only three inputs:
apt_program.xml, apt_program.pointing (both exported from APT), and a
target catalog (or a list of target catalogs).

Several changes possibly merit individual github issues. There are several changes I made that will certainly break the code for NIRCam, for which is was originally developed. We have to include
NIRCam test cases to iron those braking changes out.

I included three programs in `tests/test_apt_interface.py`, which pass on Travis:
1. program 1087 (NIRISS astrometric calibration): NIRISS imaging with FGS imaging in parallel, dithers, mosaic
2. m31_field_test_observation: NIRISS imaging with dithers (provided by @KevinVolkSTScI )
3. 1087_minimal: single exposure NIRISS imaging (only testing `write_observationlist.write_yaml`)

So far this is mostly tested for functionality, not accuracy, so there may be errors in the produced yaml files. I am now proceeding to running the data generation, but I am requesting preliminary reviews so that we can work together towards the best solutions for the issues below:

One of the fundamental problems is that the `SimInput` class was written for a single-instrument use case (as far as I can see). That breaks down for parallel or multi-instrument APT programs. I suggest that the solution is to populate SimInput will all necessary configuration files (see the small method `set_global_definitions` I added) by using dictionaries with the instrument or detector name as key. This can be done in advance or when a method is called.

Other problems I see are
- The relevant code uses a mix of passing data as method arguments and/or class/instance attributes. It is not easy to follow what is happening, and which function modifies a certain object.
- The naming of variables is sometimes misleading. `_tab` or `_list` variables actually hold dictionaries

Another open (but not critical) issue is how to set the ObsIds of parallel observations correctly (in apt_inputs.py `AptInput.get_pointing_info`). Currently its all zeros.

I admit that I lost a bit the overview of the code architecture and made changes to the existing infrastructure to allow the NIRISS programs to be processed. There may be more elegant solutions that require more fundamental changes to mirage.

KevinVolkSTScI:
I ran the M31 test case and it produced the proper number of .yaml files.  I have to revise the files because the source lists are a function of the filter whereas the .yaml files are all generated with the same point source list.  I noted, however, that the dark file name is not entered properly.  The files have either "dark: n " or "dark: o ".  Is this what is expected?
I am working on running a test with these files to check the offsets explicitly, but the values listed in the output file from the test appear to be correct so I assume that the stars will be in the expected places in the test.

KevinVolkSTScI:
I have run into some issue with the .yaml files that are generated from the test case, and I do not know what the problem is.  Having made a small code to revise the original .yaml files to have a suitable dark current file name and the right stars list and galaxies list files for a given filter, when I try to run the code I get an error about not having a distortion parameter.  Putting the distortion_coeffs parameter in by hand gets past this error but then there is another error about a detector parameter

>>> t1.create()
Requested readout pattern NISRAPID is valid. Using the nframe = 1 and nskip = 0
  File "<stdin>", line 1, in <module>
  File "/grp/jwst/nis1/kevin/anaconda3/envs/mirage_linux/lib/python3.6/site-packages/mirage-2.0.1-py3.6.egg/mirage/imaging_simulator.py", line 57, in create
    cat.make_seed()
  File "/grp/jwst/nis1/kevin/anaconda3/envs/mirage_linux/lib/python3.6/site-packages/mirage-2.0.1-py3.6.egg/mirage/seed_image/catalog_seed_image.py", line 108, in make_seed
    self.checkParams()
  File "/grp/jwst/nis1/kevin/anaconda3/envs/mirage_linux/lib/python3.6/site-packages/mirage-2.0.1-py3.6.egg/mirage/seed_image/catalog_seed_image.py", line 3113, in checkParams
    detector = self.subdict[self.subdict['AperName'] == aper_name]['Detector'][0]
  File "/grp/jwst/nis1/kevin/anaconda3/envs/mirage_linux/lib/python3.6/site-packages/astropy/table/table.py", line 1222, in __getitem__
    return self.columns[item]
  File "/grp/jwst/nis1/kevin/anaconda3/envs/mirage_linux/lib/python3.6/site-packages/astropy/table/table.py", line 106, in __getitem__
    return OrderedDict.__getitem__(self, item)
KeyError: 'Detector'

I have not been able to determine what parameter value in the .yaml file is causing this issue.

Johannes-Sahlmann:
@KevinVolkSTScI I am not sure what the issue is in your case. Everything is working fine now for me, mirage is generating NIRISS, FGS1 and FGS2 images according to the NIRISS-12 program APT (not checked for astrometric accuracy yet). Can you please pull in the latest version of this branch and try again?

KevinVolkSTScI:
I will try again.  However another question has come up.  For the M31 test case the rotation needs to be 216 degrees for a given date assumed for the observation (I think it is 1 September 2020...).  The .yaml files have a rotation value of 0.0 degrees.  I think that rotation value is not allowed for this field.  Do we know from the APT file listing the dithers what the assumed rotation angle on the sky is?  In making the mirage test I set the rotation angle to 216 degrees, but one probably needs to offset by the NIRISS/V3 angle of about 1.5 degrees.  I do not know which sign is the correct sign for this correction, is it +1.5 degrees with respect to V3 or is it -1.5 degrees with respect to V3?

I have the measurements of the source positions in the four dithers and I have to check these against what the OCD specifies for the dither pattern.

KevinVolkSTScI:
The measured offsets in the images are not what is expected from the OCD.  As noted above I suspect that the rotation is causing this.  I will re-do the simulations with zero rotation and see what happens.

KevinVolkSTScI:
When I ran the test with rotation 0.0 degrees the offsets match the official table of values, as expected.  Hence we need to know whether this is a problem in the APT or a problem in the code translating the APT values to make the .yaml files.  It may be the latter, if the APT file only reports the offsets.  In any case, we need to get this fixed somehow.

bhilbert4:
From my reading of the code (but not running any tests yet), `read_apt_xml.py` will fail for cases where one of NIRCam/NIRISS/FGS are not prime but are the parallel instrument (e.g. MIRI with NIRISS parallel). This is because the observation template name will reference MIRI and will not be in the list of known templates, at which point `read_apt_xml.py` will raise an error.

mirage/yaml/write_observationlist.py
```diff
@@ -33,117 +99,227 @@ def write_yaml(xml_file, pointing_file, yaml_file, ps_cat_sw=None, ps_cat_lw=Non
...
+                    "  ExtendedCenter: {}\n".format(ExtendedCenter),
+                    "  MovingTargetList: {}\n".format(MovingTargetList),
+                    "  MovingTargetSersic: {}\n".format(MovingTargetSersic),
+                    "  MovingTargetExtended: {}\n".format(MovingTargetExtended),
+                    "  MovingTargetConvolveExtended: {}\n".format(MovingTargetConvolveExtended),
+                    "  MovingTargetToTrack: {}\n".format(MovingTargetToTrack),
+                    "  BackgroundRate: {}\n".format(BackgroundRate),
+                ]
+            # multiply by the number of dithers
+            text_out += text * xml_dict['ImageDithers'][exposure_index]
```
bhilbert4: I'm not sure that you want a separate entry for each dither, do you? Previously it was one entry per observation (or for NIRCam, one entry per filter configuration in each observation).

KevinVolkSTScI: For simulation purposes we need one .yaml file per position and filter.

bhilbert4: This is for specifying which catalogs go with which observations within the proposal. Previously the code was set up using the assumption that the same source catalogs would be used for all dithers of a particular filter set within a particular observation. Requiring the user to specify a different source catalog for each dither seems to me like overkill.

KevinVolkSTScI: We will only have one catalogue per observation (actually we want one catalogue per filter to allow for the changes in brightness from one filter to another, but that can be fixed after the .yaml files are generated).  The catalogue will cover an area rather larger than the NIRISS field of view and so the dithers will not carry the field beyond the catalogue area.

bhilbert4: Right. That's the way the observation list was originally designed to work. (Although you should be able to specify different catalogs for each filter and not have to change the yaml files after they are generated.)

bhilbert4:
One other APT file case that occurs to me: I've started looking at a NIRCam commissioning APT file that contains a mix of NIRCamEngineeringImaging and NIRCamDark templates. Mirage supports the former, but not the latter. In general I think it makes sense for an error to be raised when a file contains an unknown template, but Mirage will never simulate darks, so maybe in this case Mirage should just skip that observation? I'm assuming there are also dark current templates for NIRISS (and FGS?) that we would want to skip over as well.

Johannes-Sahlmann:
@bhilbert4 @laurenmarietta I think this is ready for a first thorough review, focusing on coding and functional aspects. However, I guess this goes beyond a normal review, because you would ideally throw as many APT programs at this branch as possible, and report the cases when the code raises errors. That is what I mean by "functional aspects".
In the meantime, I will systematically use the produced yaml files to generate actual mirage simulations (so far I have only run some example checks). This will tell how accurate the code is.

Note that I added `xfail` to test_apt.py because the hardcoded number do no longer apply.

KevinVolkSTScI:
I will try again to install this code and test it.  I was having issues with git on witserv with mirage last week, and so this may affect how soon I can do any testing if it shows up again.

KevinVolkSTScI:
Johannes, is there now some way to enter the desired V3 rotation angle when generating the .yaml files from the APT file?  Or does it do something based on a proposed date of observation?  In the one test I did make previously the V3 rotation angle was assumed to be zero degrees in making the dither offset calculations, which was wrong for the particular case I was trying out.

Johannes-Sahlmann:
@KevinVolkSTScI That interface is not well developed. Currently you have to hardcode the desired roll angle in `write_observationlist.py` (line 211). It will be the same for all observations in the program. 
We already discussed that in the future, we will need a clean interface for this and the possibility to specify different values of the roll angle (and other parameters) for different observations/exposures (e.g. to reflect timing or roll stability).

Johannes-Sahlmann:
The updated observation list format specifies the roll angle for every exposure. After generating the observation list from the APT files, you could modify it by hand. What I am referring to is that there is no automated way to do this yet. For example, I could think about using a `roll uncertainty/jitter` keyword to introduce some random variation in the individual roll angles.

bhilbert4:
I'm getting an error from `write_observationlist.py` for my first example case (a NIRCam-only proposal). It looks like maybe it is looking for a directory that doesn't exist. Is it a directory that it should have created? I never mentioned the `observation_list_apt_files` directory anywhere. 

I'm calling it as:
`write_observationlist.write_yaml(xml_file, yaml_output, catalog_files=catalog_dir,
                                     ps_cat_sw=ps_cat_sw, ps_cat_lw=ps_cat_lw)`

`catalog_dir` is a string: 'catalogs'
`ps_cat_sw` and `lw` are also strings.

```
Filter config index 0
Filter config index 1
Filter config index 2
Filter config index 3
Number of entries before expanding dithers: 16
Number of entries after expanding dithers:  48
  File "create_yaml_files.py", line 114, in <module>
    create_data()
  File "create_yaml_files.py", line 28, in create_data
    create_observation_list(xmlfile, 'catalogs', 'ptsrc.cat', 'ptsrc.cat')
  File "create_yaml_files.py", line 55, in create_observation_list
    ps_cat_sw=ps_cat_sw, ps_cat_lw=ps_cat_lw)
  File "/Users/hilbert/python_repos/mirage/mirage/yaml/write_observationlist.py", line 321, in write_yaml
    f = open(yaml_file, 'w')
FileNotFoundError: [Errno 2] No such file or directory: 'observation_list_apt_files/1069.yaml'
```

bhilbert4:
I don't see the `1069.yaml` file anywhere.

Johannes-Sahlmann:
Line 193 in `write_observationlist.py` says `# temporary fix for NIRCam`. The inputs `ps_cat_sw` and `ps_cat_lw` are actually ignored in this case.  Also `catalog_files` should be a filename (or list of filenames) not a directory.

This is clearly not finished and not well documented. Can you please make that combination of APT files and catalogs available to me, so I can make the necessary changes? Please also indicate if I can incorporate them in a test case, i.e. commit them to the repo.

bhilbert4:
Ah ok, it was my mistake. I forgot to strip the path from the  xml filename when I was constructing the yaml output name. That path was then propagating into the yaml output name.

mirage/yaml/write_observationlist.py
```diff
@@ -33,117 +218,110 @@ def write_yaml(xml_file, pointing_file, yaml_file, ps_cat_sw=None, ps_cat_lw=Non
...
+                    sw_filt = xml_dict['ShortFilter'][index]
+                    lw_filt = xml_dict['LongFilter'][index]
+
+                    text += [
+                        "    FilterConfig:\n",
+                        # "  FilterConfig{}:\n".format(exposure_index),
+                        "      SW:\n",
+                        "        Filter: {}\n".format(sw_filt),
+                        "        PointSourceCatalog: {}\n".format(
+                            ps_cat_sw[observation_index]),
```
bhilbert4: if pt_cat_sw is a string rather than a list, then this statement will set the catalog value to the first character of the string.

mirage/yaml/write_observationlist.py
```diff
@@ -33,117 +218,110 @@ def write_yaml(xml_file, pointing_file, yaml_file, ps_cat_sw=None, ps_cat_lw=Non
...
+                        "        MovingTargetSersic: {}\n".format(MovingTargetSersic),
+                        "        MovingTargetExtended: {}\n".format(MovingTargetExtended),
+                        "        MovingTargetConvolveExtended: {}\n".format(
+                            MovingTargetConvolveExtended),
+                        "        MovingTargetToTrack: {}\n".format(MovingTargetToTrack),
+                        "        BackgroundRate: {}\n".format(BackgroundRate_sw),
+                        "      LW:\n",
+                        "        Filter: {}\n".format(lw_filt),
+                        "        PointSourceCatalog: {}\n".format(
+                            ps_cat_lw[observation_index]),
```
bhilbert4: If pt_cat_lw is a string, this will assign the first character to the catalog name.

mirage/yaml/write_observationlist.py
Johannes-Sahlmann: Yes. I need to put in a check like for `catalog_files` in Line 166.

bhilbert4:
It is successfully creating the observation file for me, but crashes when getting the pointing information within `yaml_generator.py`. I'll email you the xml and pointing files.

```
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Observation `1` labelled `None` uses template `NircamImaging`
APTObservationParams Dictionary holds 0 entries before reading template
Filter config index 0
Filter config index 1
Filter config index 2
Filter config index 3
Dictionary read from template has 4 entries.
Found 4 tile(s) for observation 1 Observation 1
Found 4 visits with numbers: [1, 2, 3, 4]
APTObservationParams Dictionary holds 16 entries after reading template (+16 entries)
Finished reading APT xml file.
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Number of entries before expanding dithers: 16
Number of entries after expanding dithers:  48

Wrote 1 observations and 48 entries to apt_files/observation_list_1069.yaml

Filter config index 0
Filter config index 1
Filter config index 2
Filter config index 3
  File "create_yaml_files.py", line 122, in <module>
    create_data()
  File "create_yaml_files.py", line 31, in create_data
    yam_object = create_yaml_files(xmlfile, pointingfile, obs_file)
  File "create_yaml_files.py", line 82, in create_yaml_files
    yam.create_inputs()
  File "/Users/hilbert/python_repos/mirage/mirage/yaml/yaml_generator.py", line 255, in create_inputs
    apt.create_input_table()
  File "/Users/hilbert/python_repos/mirage/mirage/apt/apt_inputs.py", line 298, in create_input_table
    pointing_tab = self.get_pointing_info(self.pointing_file, propid=xmltab['ProposalID'][0])
  File "/Users/hilbert/python_repos/mirage/mirage/apt/apt_inputs.py", line 560, in get_pointing_info
    if ((np.int(elements[1]) > 0) & ('NRC' in elements[4]
IndexError: list index out of range
```

Johannes-Sahlmann:
Thanks for running this test. The pointing file reader is obviously still fragile. I will work on it.

Johannes-Sahlmann:
The code runs fine for me on the set of APT files for program 1069  you sent me:
```
====================================================================================================
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Observation `1` labelled `None` uses template `NircamImaging`
APTObservationParams Dictionary holds 0 entries before reading template
Dictionary read from template has 4 entries.
Found 4 tile(s) for observation 1 Observation 1
Found 4 visits with numbers: [1, 2, 3, 4]
APTObservationParams Dictionary holds 16 entries after reading template (+16 entries)
Finished reading APT xml file.
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Number of entries before expanding dithers: 16
Number of entries after expanding dithers:  48

Wrote 1 observations and 48 entries to /Users/jsahlmann/jwst/code/github/spacetelescope/mirage/tests/temp_data/nircam_1069_observation_list.yaml
Wrote exposure table to /Users/jsahlmann/jwst/code/github/spacetelescope/mirage/tests/temp_data/expand_for_detectors.csv
NOTE: Using pre-delivery SIAF data
[...]
csv exposure list written to /Users/jsahlmann/jwst/code/github/spacetelescope/mirage/tests/temp_data/Observation_table_for_1069.xml.csv
Updated observation table file saved to /Users/jsahlmann/jwst/code/github/spacetelescope/mirage/tests/temp_data/Observation_table_for_1069.xml_with_yaml_parameters.csv


Observation 100: 
   1 visit(s) 
   4 exposure(s)
   10 detector(s) in modules A and B

4 exposures total.
480 output files written to: /Users/jsahlmann/jwst/code/github/spacetelescope/mirage/tests/temp_data


```

bhilbert4:
Hmm. I'm still not getting it to run all the way through. I was having network connectivity problems which meant I couldn't see the reference files. But now I'm getting different errors.

For the case where I provide the apt_xml_dictionary to `yaml_generator.create_inputs`:

```
yaml_output = '1069.yaml'
obs_dict = write_observationlist.write_yaml(xml_file, yaml_output, catalog_files=catalog_dir,
                                                              ps_cat_sw=ps_cat_sw, ps_cat_lw=ps_cat_lw, verbose=True)
yam = yaml_generator.SimInput()
yam.input_xml = xml_file
yam.pointing_file = pointing_file
yam.output_dir = os.path.join('yaml_files', basename)
yam.simdata_output_dir = os.path.join('simulated_data', basename)
yam.observation_table = yaml_output
yam.use_JWST_pipeline = True
yam.use_linearized_darks = True
yam.datatype = 'raw'
yam.background_rate = "low"
yam.reffile_setup()
yam.create_inputs(apt_xml_dict=obs_dict)
```

I get the following error, which I find very odd. `global_readout_patterns` is defined in the `set_global_definitions` function, although in looking through the `yaml_generator` I don't see anywhere that `set_global_definitions` is being called.

```
csv exposure list written to /Users/hilbert/python_repos/test_mirage_updated_apt/yaml_files/1069/Observation_table_for_1069.xml.csv
  File "create_yaml_files.py", line 123, in <module>
    create_data()
  File "create_yaml_files.py", line 31, in create_data
    yam_object = create_yaml_files(xmlfile, pointingfile, obs_file, obs_dict=obs_dict)
  File "create_yaml_files.py", line 83, in create_yaml_files
    yam.create_inputs(apt_xml_dict=obs_dict)
  File "/Users/hilbert/python_repos/mirage/mirage/yaml/yaml_generator.py", line 261, in create_inputs
    self.make_start_times()
  File "/Users/hilbert/python_repos/mirage/mirage/yaml/yaml_generator.py", line 744, in make_start_times
    readpatt_def = self.global_readout_patterns[instrument.lower()]
AttributeError: 'SimInput' object has no attribute 'global_readout_patterns'
```




And in the case where I provide the name of the observation list file but not the dictionary itself, I get a different error:
```
yaml_output = '1069.yaml'
obs_dict = write_observationlist.write_yaml(xml_file, yaml_output, catalog_files=catalog_dir,
                                                              ps_cat_sw=ps_cat_sw, ps_cat_lw=ps_cat_lw, verbose=True)
yam = yaml_generator.SimInput()
yam.input_xml = xml_file
yam.pointing_file = pointing_file
yam.output_dir = os.path.join('yaml_files', basename)
yam.simdata_output_dir = os.path.join('simulated_data', basename)
yam.observation_table = yaml_output
yam.use_JWST_pipeline = True
yam.use_linearized_darks = True
yam.datatype = 'raw'
yam.background_rate = "low"
yam.reffile_setup()
yam.create_inputs()
```

Here is this traceback:

```
  File "create_yaml_files.py", line 123, in <module>
    create_data()
  File "create_yaml_files.py", line 31, in create_data
    yam_object = create_yaml_files(xmlfile, pointingfile, obs_file)
  File "create_yaml_files.py", line 83, in create_yaml_files
    yam.create_inputs()
  File "/Users/hilbert/python_repos/mirage/mirage/yaml/yaml_generator.py", line 255, in create_inputs
    apt.create_input_table()
  File "/Users/hilbert/python_repos/mirage/mirage/apt/apt_inputs.py", line 302, in create_input_table
    'Inconsistent table size from XML file ({}) and pointing file ({}). Something was not processed correctly in apt_inputs.'.format(len(xmltab['ProposalID']), len(pointing_tab['obs_num']))
AssertionError: Inconsistent table size from XML file (16) and pointing file (48). Something was not processed correctly in apt_inputs.
```

bhilbert4:
I haven't looked at the code yet to see what it's doing, but the DitherIndex is zero for all entries in my observation file.

KevinVolkSTScI:
When I tried to run the test cases in test_apt_interface.py two of the three tests failed.  The problem is something in reading the .xml files (it seems).  The same type of error comes up when I try to run an example myself for NIRISS CARS program 1089.  The error message I get is 

(mirage_osx) kvolk:test_apt_conversion$ python test_conversion.py
  File "test_conversion.py", line 17, in <module>
    apt_xml_dict = write_observationlist.write_yaml(xml_name,pointing_name)
  File "/Users/kvolk/anaconda/envs/mirage_osx/lib/python3.6/site-packages/mirage-0.0.0.dev380+g1a02ebe9-py3.6.egg/mirage/yaml/write_observationlist.py", line 172, in write_yaml
    xml_dict = readxml_obj.read_xml(xml_file, verbose=verbose)
  File "/Users/kvolk/anaconda/envs/mirage_osx/lib/python3.6/site-packages/mirage-0.0.0.dev380+g1a02ebe9-py3.6.egg/mirage/apt/read_apt_xml.py", line 141, in read_xml
    tree = etree.parse(f)
  File "src/lxml/etree.pyx", line 3426, in lxml.etree.parse
  File "src/lxml/parser.pxi", line 1861, in lxml.etree._parseDocument
  File "src/lxml/parser.pxi", line 1881, in lxml.etree._parseFilelikeDocument
  File "src/lxml/parser.pxi", line 1776, in lxml.etree._parseDocFromFilelike
  File "src/lxml/parser.pxi", line 1187, in lxml.etree._BaseParser._parseDocFromFilelike
  File "src/lxml/parser.pxi", line 601, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 707, in lxml.etree._handleParseResult
  File "src/lxml/etree.pyx", line 316, in lxml.etree._ExceptionContext._raise_if_stored
  File "src/lxml/parser.pxi", line 370, in lxml.etree._FileReaderContext.copyToBuffer
  File "/Users/kvolk/anaconda/envs/mirage_osx/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 6431: ordinal not in range(128)
(mirage_osx) kvolk:test_apt_conversion$ python test_conversion.py
  File "test_conversion.py", line 17, in <module>
    apt_xml_dict = write_observationlist.write_yaml(xml_name,pointing_name)
  File "/Users/kvolk/anaconda/envs/mirage_osx/lib/python3.6/site-packages/mirage-0.0.0.dev380+g1a02ebe9-py3.6.egg/mirage/yaml/write_observationlist.py", line 172, in write_yaml
    xml_dict = readxml_obj.read_xml(xml_file, verbose=verbose)
  File "/Users/kvolk/anaconda/envs/mirage_osx/lib/python3.6/site-packages/mirage-0.0.0.dev380+g1a02ebe9-py3.6.egg/mirage/apt/read_apt_xml.py", line 141, in read_xml
    tree = etree.parse(f)
  File "src/lxml/etree.pyx", line 3426, in lxml.etree.parse
  File "src/lxml/parser.pxi", line 1861, in lxml.etree._parseDocument
  File "src/lxml/parser.pxi", line 1881, in lxml.etree._parseFilelikeDocument
  File "src/lxml/parser.pxi", line 1776, in lxml.etree._parseDocFromFilelike
  File "src/lxml/parser.pxi", line 1187, in lxml.etree._BaseParser._parseDocFromFilelike
  File "src/lxml/parser.pxi", line 601, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 707, in lxml.etree._handleParseResult
  File "src/lxml/etree.pyx", line 316, in lxml.etree._ExceptionContext._raise_if_stored
  File "src/lxml/parser.pxi", line 370, in lxml.etree._FileReaderContext.copyToBuffer
  File "/Users/kvolk/anaconda/envs/mirage_osx/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 6431: ordinal not in range(128)

The error messages from the run of the pytest are similar, and so it looks like the same issue is present in all three cases.   Here is a link to the log of the pytest:
[test_apt_interface.log](https://github.com/spacetelescope/mirage/files/2519421/test_apt_interface.log)
I do not know for certain if this is a problem with my installation under conda or a problem in the code.  It does looks like a problem with the Python installation.  I installed mirage on my Mac laptop using the recommended procedure with mirage_environment_osx.yml for this test, so ideally all the dependencies should be OK.  Nonetheless the tests fail.  Looking at the APT file for program 1089 shows no particular special characters in the files, the only one I saw was the mu symbol for microns in the description of the program.

KevinVolkSTScI:
A minor comment about the pytest file for the APT interface: in the docstring at the top the file has 

Use
---
    >>> pytest -s test_niriss_imaging.py

which should be

Use
---
    >>> pytest -s test_apt_interface.py

KevinVolkSTScI:
Looking online it appears that this issue is caused by Python trying to interpret a unicode string from the .xml file that has a unicode character in it.  This does not explain why Johannes can run the test case and not get this error while I get the error.  Apparently one can fix this by changing the encoding (see https://stackoverflow.com/questions/43555593/unicodedecodeerror-ascii-codec-cant-decode-byte-0xe2-in-position-6-ordinal).  I am not sure where one would put this in the code or whether it would break something else.

mirage/yaml/yaml_generator.py
```diff
@@ -379,6 +389,13 @@ def create_inputs(self):
...
             file_dict['subpix_dither_num'] = subpix_dither + 1
+            file_dict['subarray_def_file'] = self.global_subarray_definition_files[self.info['Instrument'][i].lower()]
+            file_dict['readpatt_def_file'] = self.global_readout_pattern_files[self.info['Instrument'][i].lower()]
+            file_dict['crosstalk_file'] = self.global_crosstalk_files[self.info['Instrument'][i].lower()]
+            file_dict['filtpupilcombo_file'] = self.global_filtpupilcombo_files[self.info['Instrument'][i].lower()]
+            file_dict['flux_cal_file'] = self.global_flux_cal_files[self.info['Instrument'][i].lower()]
+            file_dict['psfpath'] = self.global_psfpath[self.info['Instrument'][i].lower()]
+
             fname = self.write_yaml(file_dict)
             yamls.append(fname)
```
laurenmarietta: Further down, at line 408, the `mosaic_numbers` are being defined based on the old file naming scheme. Related to a previous comment, we should figure out what the correct naming scheme is and then make sure this matches. Currently it does not, so `yaml_generator` is telling me about Observation 000, 100, 200, etc. instead of 001, 002...

mirage/yaml/yaml_generator.py
```diff
@@ -379,6 +389,13 @@ def create_inputs(self):
...
             subpix_dither = (tot_dith-1) % subpixtot
             file_dict['subpix_dither_num'] = subpix_dither + 1
+            file_dict['subarray_def_file'] = self.global_subarray_definition_files[self.info['Instrument'][i].lower()]
+            file_dict['readpatt_def_file'] = self.global_readout_pattern_files[self.info['Instrument'][i].lower()]
+            file_dict['crosstalk_file'] = self.global_crosstalk_files[self.info['Instrument'][i].lower()]
+            file_dict['filtpupilcombo_file'] = self.global_filtpupilcombo_files[self.info['Instrument'][i].lower()]
+            file_dict['flux_cal_file'] = self.global_flux_cal_files[self.info['Instrument'][i].lower()]
+            file_dict['psfpath'] = self.global_psfpath[self.info['Instrument'][i].lower()]
+
             fname = self.write_yaml(file_dict)
```
laurenmarietta: Also further down, at line 432 in `yaml_generator`, it breaks for OTE-12 with the SUB96DHSPILA aperture. This should probably be fixed in the same PR that addresses #190, if you agree that should be separate.

mirage/apt/apt_inputs.py
```diff
@@ -38,20 +38,22 @@
...
-from astropy.table import Table
+from astropy.table import Table, vstack
 from astropy.io import ascii
 import numpy as np
-import pysiaf
+from pysiaf import rotations
 import yaml
 
 from . import read_apt_xml
 from ..utils import rotations
```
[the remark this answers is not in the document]
bhilbert4: Removing utils.rotations completely is on the to do list.

mirage/apt/apt_inputs.py
```diff
@@ -677,14 +612,19 @@ def get_pointing_info(self, file, propid):
...
                             dkpar.append(np.int(elements[20]))
-                            ddist.append(np.float(elements[21]))
+                            if elements[18] == 'PARALLEL':
+                                ddist.append(None)
+                            else:
+                                ddist.append(np.float(elements[21]))
                             # For the moment we assume that the instrument being simulated is not being
                             # run in parallel, so the parallel proposal number will be all zeros,
                             # as seen in the line below.
                             observation_id.append("V{}P{}{}{}{}".format(vid, '00000000', vgrp, seq, act))
```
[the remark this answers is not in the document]
bhilbert4: I think that's correct as is. The yaml filename follows the just naming convention, which is jw**yaml. But the observation ID should be in the (e.g) V80600013001P0000000002101 format. See the keyword dictionary entry for OBS_ID: https://mast.stsci.edu/portal/Mashup/Clients/jwkeywords/index.html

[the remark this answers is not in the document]
laurenmarietta: Oh! I see. 
In that case, it might be useful to include a new `'filename'` key in the pointing dictionary, so we have a connection between the dictionary and the files.
I see that `yaml_generator` adds a `yamlfile` key, but that isn't actually in the same format as the yaml files (e.g. Act01_A3_imaging_Dither01.yaml).

Johannes-Sahlmann:
@bhilbert4 I think the errors you got can be avoided by calling
```yam.reffile_setup(instrument=instrument, offline=False)
            yam.set_global_definitions()
```
And keeping
```yam.create_inputs(apt_xml_dict=obs_dict)```

For reference, see the code I am using in test_apt_interface.test_complete_input_generation

**Comment by bhilbert4** [comment]
That worked for me, thanks. Should `set_global_definitions` be called from somewhere within `yaml_generator`? I'm just wondering if it should be that the user has to call it separately. 

Is the intention that `create_inputs` should now always have `apt_xml_dict` passed to it? I still get an error when I don't pass the dictionary.

**Comment by Johannes-Sahlmann** [comment]
Yes, I think `set_global_definitions` should be called implicitly and not by the user.
I would also vote for making the `apt_xml_dict` argument implicit (i.e. replace it by an attribute),and that is part of the discussion I want to have.

**Comment by bhilbert4** [comment]
I agree with both of those points. I'm also kicking myself for not originally having `reffile_setup` called implicitly.

**Comment by bhilbert4** [comment]
I'm working with an APT file that has NIRCam as prime and MIRI as parallel, and I'm getting an error. It looks like in this case, `instrument` is not defined in `read_apt_xml.read_generic_imaging_template` before the print statement that lists the prime and parallel instruments. (line 413).

```
Working on apt_files/Final_DeepField.xml   apt_files/Final_DeepField.pointing
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Observation `1` labelled `F410M-1` uses template `NircamImaging`
APTObservationParams Dictionary holds 0 entries before reading template
Filter config index 0
  File "create_yaml_files.py", line 124, in <module>
    create_data()
  File "create_yaml_files.py", line 28, in create_data
    obs_file, obs_dict = create_observation_list(xmlfile, 'catalogs', 'ptsrc.cat', 'ptsrc.cat')
  File "create_yaml_files.py", line 57, in create_observation_list
    ps_cat_sw=ps_cat_sw, ps_cat_lw=ps_cat_lw, verbose=True)
  File "/Users/hilbert/python_repos/mirage/mirage/yaml/write_observationlist.py", line 172, in write_yaml
    xml_dict = readxml_obj.read_xml(xml_file, verbose=verbose)
  File "/Users/hilbert/python_repos/mirage/mirage/apt/read_apt_xml.py", line 277, in read_xml
    parallel_exposures_dictionary = self.read_parallel_exposures(obs, exposures_dictionary, proposal_parameter_dictionary, verbose=verbose)
  File "/Users/hilbert/python_repos/mirage/mirage/apt/read_apt_xml.py", line 1291, in read_parallel_exposures
    verbose=verbose)
  File "/Users/hilbert/python_repos/mirage/mirage/apt/read_apt_xml.py", line 413, in read_generic_imaging_template
    print('Prime: {}   Parallel: {}'.format(prime_instrument, instrument))
UnboundLocalError: local variable 'instrument' referenced before assignment
```

**Comment by KevinVolkSTScI** [comment]
Does anyone have any clues about the error message that I get when running the tests?  I tried to install the current version of mirage on witserv3 and that fails, apparently because the version of git is too old and does not produce the proper response to a version number query in the code.  So I am stuck both on witserv and my laptop.  I assume that Johannes and Bryan have not seen the same error that I get.  It in principle is a codecs error, but I have not been able to determine how I might fix this based on the documentation on such places as StackOverflow.  I have not made any configuration changes on the laptop that would affect the codecs, as far as I know.

File: mirage/apt/apt_inputs.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (Johannes-Sahlmann) [reply mirage/apt/apt_inputs.py]: I fixed this here but a separate PR will be needed to remove the remaining occurrences.

File: mirage/apt/read_apt_xml.py [hunk 8]
```diff
@@ -12,6 +14,53 @@
 PACKAGE_DIR = os.path.dirname(APT_DIR)
 
 
+def append_dictionary(base_dictionary, added_dictionary, braid=False):
```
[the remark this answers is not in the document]
Johannes-Sahlmann: I moved this function to `utils`

mirage/yaml/write_observationlist.py
```diff
@@ -1,29 +1,214 @@
...
+    xml_file : str
+        path to APT .xml file
+    yaml_file : str
+        output_file
+    catalog_files : str or list(str)
+        path to file(s) that contain the catalog of sources.
+    ps_cat_sw : str or list(str)
+        NIRCam SW catalog, one per observations
+    ps_cat_lw : str or list(str)
+        NIRCam LW catalog, one per observations
```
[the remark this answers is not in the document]
Johannes-Sahlmann: I changed the arguments of the function to require a dictionary `catalogs` with one key per instrument. This interface is not polished yet but functional.

laurenmarietta:
@KevinVolkSTScI: I believe I have seen an error like that in a different project. I think it might have something to do with older versions of the `lxml` package needing the input into `tree` to be _explicitly_ encoded as utf-8.

In my other project, it was fixed by changing this (which is in `read_apt_xml`):
```
with open(infile) as f:
    tree = etree.parse(f)
```

to this: 
```
with open(infile, encoding="utf-8") as f:
    tree = etree.parse(f)
```
Maybe that will work for you, as well?

KevinVolkSTScI:
I am trying to test the code on witserv3.  (After pestering the IT people they installed git 2.19.1 on the witserv machines, it can be accessed by the "module load git" command.  This allows me to install mirage on witserv again.)  I selected NIRISS CAR 1089 for the test, but when I tried running the APT conversion I got the following error message:

witserv3 mirage_apt_test > python test_conversion.py
  File "test_conversion.py", line 17, in <module>
    apt_xml_dict = write_observationlist.write_yaml(xml_name,pointing_name)
  File "/grp/jwst/nis1/kevin/anaconda3/envs/mirage_linux_testing/lib/python3.6/site-packages/mirage-0.0.0.dev394+gcb97182b-py3.6.egg/mirage/yaml/write_observationlist.py", line 172, in write_yaml
    xml_dict = readxml_obj.read_xml(xml_file, verbose=verbose)
  File "/grp/jwst/nis1/kevin/anaconda3/envs/mirage_linux_testing/lib/python3.6/site-packages/mirage-0.0.0.dev394+gcb97182b-py3.6.egg/mirage/apt/read_apt_xml.py", line 217, in read_xml
    raise ValueError('No protocol written to read {} template.'.format(template_name))
ValueError: No protocol written to read NirissWfss template.

Program 1089 is a WFSS trace measurement program.  I had not considered the mode when I selected this to test, I just wanted a CAR with one target field to simplify the test.  I assume that this is the expected result for NRIISS WFSS programs, which currently cannot be handled in mirage.  I would like to verify this with Johannes.

I would rather that that code either (a) only produces .yaml files for the direct images in the WFSS program, and skips the spectral ones or, (b) produces all the .yaml files for direct images at all the pointings whether spectral or imaging ones, so that the direct images for the spectral exposures can be made as input to grizli.  As well, one hopes that eventually mirage can simulate the NIRISS WFSS case as it can the NIRCam WFSS one.  However, it may be that either of these things would take a lot of coding, in which case it may not be worth doing.

KevinVolkSTScI:
I have been trying to get an example APT -> .yaml conversion to work.  The code is

#! /usr/bin/env python
#
from mirage.yaml import write_observationlist, yaml_generator
from mirage.apt import read_apt_xml, apt_inputs
from mirage.utils import siaf_interface
xml_name = '1094rev1.xml'
pointing_name = '1094rev1.pointing'
output_directory = './'
data_type = 'linear, raw'
point_source_names = ['stars_field11_combined_F090W.list','stars_field11_combined_F277W.list',
                      'stars_field11_combined_F115W.list','stars_field11_combined_F356W.list',
                      'stars_field11_combined_F140M.list','stars_field11_combined_F380M.list',
                      'stars_field11_combined_F150W.list','stars_field11_combined_F430M.list',
                      'stars_field11_combined_F158M.list','stars_field11_combined_F444W.list',
                      'stars_field11_combined_F200W.list','stars_field11_combined_F480M.list']
background = 0.6
apt_xml_dict = write_observationlist.write_yaml(xml_name,pointing_name,point_source_names)
yam = yaml_generator.SimInput()
yam.input_xml = xml_name
yam.pointing_file = pointing_name
yam.output_dir = './'
yam.use_JWST_pipeline = True
yam.use_linearized_darks = False
yam.datatype = 'raw'
yam.point_source = 'stars_field11_combined_F090W.list'
yam.bkgdrate = background
yam.reffile_setup(instrument='NIRISS', offline=True)
yam.set_global_definitions()
yam.create_inputs(apt_xml_dict=apt_xml_dict)

What I get is an error:

witserv3 mirage_apt_test > python test_conversion.py
Number of entries before expanding dithers: 32
Number of entries after expanding dithers:  64
  File "test_conversion.py", line 17, in <module>
    apt_xml_dict = write_observationlist.write_yaml(xml_name,pointing_name,point_source_names)
  File "/grp/jwst/nis1/kevin/anaconda3/envs/mirage_linux_testing/lib/python3.6/site-packages/mirage-0.0.0.dev394+gcb97182b-py3.6.egg/mirage/yaml/write_observationlist.py", line 301, in write_yaml
    "    PointSourceCatalog: {}\n".format(catalog_files[index]),
IndexError: list index out of range

Can someone tell me what I am doing wrong in passing the list of input source catalogues to the code?

KevinVolkSTScI:
Once I got things installed on witserv3 I tried to run the test_apt_interface.py test.  One test passed and one failed.  The log file is attached
[test_apt_interface.log](https://github.com/spacetelescope/mirage/files/2543205/test_apt_interface.log)
I am wondering if this is another variation on the error that I am getting on my Mac laptop where the code does not parse the .xml file properly.  I tried the fix suggested by Lauren above on my Mac laptop machine, but that did not fix the problem.  Ditto for witserv, although since the error message is about something in astropy this fix may not actually have anything to do with the problem.  On the other hand the various messages about the cparser in the error output may be indicating that there is the same sort of error parsing the .xml file.

mirage/apt/apt_inputs.py
```diff
@@ -120,172 +122,87 @@ def add_observation_info(self, intab):
...
+                for name, item in nircam_mapping.items():
+                    key = '{}_{}'.format(channel.lower(), name)
+                    intab[key] = []
+
+        # loop over entries in input dictionary
+        for index, instrument in enumerate(intab['Instrument']):
+            instrument = instrument.lower()
+
+            # retrieve corresponding entry from observation list
+            entry = get_entry(self.obstab, intab['entry_number'][index])
```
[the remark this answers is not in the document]
Johannes-Sahlmann: The way the code is called and run has been changed significantly.

mirage/apt/apt_inputs.py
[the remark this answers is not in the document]
Johannes-Sahlmann: I suggest we take this into a separate PR. I have all the necessary information from a Help Desk ticket I opened.

mirage/yaml/yaml_generator.py
Johannes-Sahlmann: Can we take this to a separate PR?

mirage/yaml/yaml_generator.py
Johannes-Sahlmann: At least the observation list generation ran fine with OTE-12, so maybe that is solved?

Johannes-Sahlmann:
I think we should include support for NirissWfss templates in a separate issue/PR. The present PR is already much too heavy to handle.

@KevinVolkSTScI I refactored the way the code is run and the catalogs are passed. Please see https://github.com/Johannes-Sahlmann/mirage/blob/bf22bdf7d384000b49963c768916b4de410ad7af/tests/test_apt_interface.py#L129 ff for examples of how to do it.

mirage/yaml/yaml_generator.py
laurenmarietta: Definitely.

bhilbert4:
This may very well be solved already in this PR, or it may be that it should be pushed to a new PR, but I wanted to check.

Using the current master branch, I just tried creating yaml files from an APT file where I'm using NIRCam's B module and the SUB400P subarray. In spite of selecting the B module in APT (implying that all 5 detectors should be used), the SUB400P subarray is defined only in detectors B1 and B5. Therefore, when I run the yaml_generator, I get an error:

```File "/Users/hilbert/python_repos/mirage/mirage/yaml/yaml_generator.py", line 253, in create_inputs
    apt.create_input_table()
  File "/Users/hilbert/python_repos/mirage/mirage/apt/apt_inputs.py", line 393, in create_input_table
    self.ra_dec_update()
  File "/Users/hilbert/python_repos/mirage/mirage/apt/apt_inputs.py", line 769, in ra_dec_update
    into valid aperture name.'.format(detector, sub))
ValueError: Cannot combine detector NRCB2 and subarray SUB400P                        into valid aperture name.```

This seems like something that can be easily checked via pysiaf rather than the way it is being checked now, and for cases where the aperture is not allowed, `apt_inputs.py` should just skip it and move on to the next entry.

File: mirage/apt/read_apt_xml.py [hunk 11]
```diff
@@ -176,74 +255,102 @@ def read_xml(self, infile):
             except AttributeError:
                 offset_x, offset_y = 0, 0
             if (offset_x != 0) or (offset_y != 0):
-                print('* * * OFFSET OF ({}, {}) IN OBS {} NOT APPLIED ***'.format(offset_x,
+                print('* * * OFFSET OF ({}, {}) IN OBS {} NOT APPLIED (should be accounted for in pointing file)***'.format(offset_x,
```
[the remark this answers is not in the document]
Johannes-Sahlmann: I will keep in in for now.

mirage/apt/apt_inputs.py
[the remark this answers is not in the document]
Johannes-Sahlmann: See #205

Johannes-Sahlmann:
I realized some reference files for the Guiders are not set correctly in `reffile_setup`. I will address this in #209

bhilbert4:
Would it be possible to update such that observations using `NircamDark` and `NirissDark` are skipped over? We have commissioning proposals that contain darks mixed with external observations.

Johannes-Sahlmann:
@bhilbert4 please make example programs containing `NircamDark` and `NirissDark` observations available to me (#198 ?) and I will try to implement it (should not be hard.)

bhilbert4:
Done. Program 1068.

Johannes-Sahlmann:
Done. Note that `NircamTimeSeries` templates are also ignored.

bhilbert4:
Thanks! Right. I forgot to strip out the TimeSeries observations from 1068. Supporting TimeSeries would be nice as well, but we can push that off to a separate PR. I haven't looked in detail at the xml to see how time series parameters are different from those in the Imaging template.

bhilbert4:
I got program 1071 to run successfully. The `yaml_generator` produced the correct number of files. But I don't think the parallel sequence number is being treated correctly (unless I misunderstand how it should be used). My impression is that for this program, which has NIRCam as prime and NIRISS as parallel, the sequence number in the NIRCam files should be '1' and that in the NIRISS files should be '2'. Is that correct? 

In the yaml files just created, it is '2' for all files. This is true in the filename of the yaml files, as well as the `Output:file` and `Output:sequence_id` fields within the yaml file. However, the `Output:obs_id` field is inconsistent and shows a value of '1' for the sequence id in all files.

I've placed a copy of the 1071 xml and pointing files in the `test_programs` directory.

Johannes-Sahlmann:
I think we should handle that aspect at a later time, see #205 
Would that be acceptable?

bhilbert4:
Sure, sounds good.

Johannes-Sahlmann:
@bhilbert4 The problem was that I was ignoring the number of `SubpixelPositions`. Should be fixed now for program 1068.

KevinVolkSTScI:
After consulting with Johannes I found that my version of lxml was 4.2.5 whereas he has 3.8.0 in his installation.  By forcing the installation to use version 3.8.0 and applying the change suggested by Lauren I got the test code to work at least for one case without getting the encoding error.  That is progress.  Later in the testing I got the following error

 FileNotFoundError: [Errno 2] No such file or directory: '/Users/kvolk/anaconda/envs/mirage_osx_testing/lib/python3.6/site-packages/pysiaf/source_data/MIRI/miri_siaf_aperture_definition.txt'

so I am curious about whether this file is supposed to be in pysiaf.

  For reference since no longer see Lauren's comment above the change on line 94 of read_apt_xml.py is to put

with open(infile,encoding="utf-8") as f:

rather than what was there originally

with open(infile) as f:

I still need to fix the miri aperture issue above to run the test of the APT interface.

laurenmarietta:
I just ran into this issue yesterday, too - it is due to the version of pysiaf being out of date. The latest version, 0.2.4, includes this MIRI file.

Johannes-Sahlmann:
I updated the enviroment.yml files to include the required pysiaf version.

bhilbert4:
It looks like Travis isn't running all of the pytests? I see it testing `test_apt.py`, `test_apt_interface.py`, and `test_siaf_interface.py`. Maybe this should be dealt with in a separate PR, but it looks like the other tests (test_<instr>_imaging.py) need to be added (They are also failing for me, so it seems like a separate PR would be best anyway.)

Johannes-Sahlmann:
Yes, we had to restrict the tests in `setup.py` for different reasons. A separate PR should deal with opening this up again.
For now, only tests that are explicitly specified in `setup.py` are run.

Johannes-Sahlmann:
Since this is ready to merge and introduced several user interface and functionality changes, how are we going to communicate this? 
@bhilbert4 maybe you could make this a new release and point the users to the most relevant changes, in particular point to e.g. `tests/test_apt_interface.py` to see how to call the APT interface.
Thoughts?

bhilbert4:
I think that makes sense, although I'm also thinking that before we advertise too broadly, we should solve the issue of how best to have users provide their catalogs (i.e. how to get them into the dictionary used by yaml_generator.py) and what to do for the other catalog types, and other parameters that can potentially vary with observation. (See #197)

PR #175, opened 2018-10-01
Move parse_ra_dec to utils, Rework source filtering

bhilbert4:
This PR updates the code to place the `parse_ra_dec` function (previously `parseRADec`) into utils.py to make it easier to call from other places. Currently catalog_seed_image.py and obs_generator.py call it. 

As part of this, I had to rework a little the way that mirage filters out sources far from the detector's field of view. Previously the table from a source catalog was passed to a function which worked on the entire table at once. However, this only worked if input locations were in units of pixels or RA, Dec in decimal degrees. If the inputs were RA and Dec strings (e.g. 10:34:23.2), it wasn't working. These strings need to be converted to decimal degrees before the sources can be filtered. That conversion is done by `parse_ra_dec`. 

So I have updated the filtering and placed it after `parse_ra_dec` is called. The downside to this is that it now works on only one source at a time, because the conversion from string to decimal degrees happens one source at a time. 

Rough speed test results, using a catalog of 10,000 sources with RA, Dec in decimal degrees. The amount of time it took to work through the catalog:

Updated code: 58 seconds
Old code: <10 seconds

I'm going to rework some more, so that the processing time only increases for catalogs where RA and Dec are given as strings.

### DIFF
--- a/mirage/ramp_generator/obs_generator.py
+++ b/mirage/ramp_generator/obs_generator.py
@@ -47,7 +47,7 @@ from ..utils import set_telescope_pointing_separated as stp
 
 INST_LIST = ['nircam', 'niriss', 'fgs']
 MODES = {"nircam": ["imaging", "ts_imaging", "wfss", "ts_wfss"],
-         "niriss": ["imaging","ami"],
+         "niriss": ["imaging", "ami"],
          "fgs": ["imaging"]}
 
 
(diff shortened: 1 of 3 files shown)


bhilbert4:
New function `get_positions` takes the place of some lines that were previously repeated several times.

PR #176, opened 2018-10-02
Moving targets: use all readpatterns

bhilbert4:
This PR fixes a couple of small bug significant bugs when creating data with moving targets: either non-sidereal observations, or sidereal observations with non-sidereal objects. 

The first fix corrected a bug that was preventing the creation of non-sidereal observations with readout patterns other than RAPID.
`non_sidereal_ramp = np.zeros((ns_int, totframes, ns_yd, ns_xd))` in line 558 of the updated `catalog_seed_images.py`.

The second fix corrected a bug that was failing to add the moving sersic and moving extended targets to the final moving target seed image when creating sidereal data with non-sidereal targets.
`mov_targs_integration += mov_targs_ramps[i]` in line 517 of the updated `catalog_seed_image.py`.

In addition I made some PEP8 updates.

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -468,6 +468,7 @@ class Catalog_seed():
             # print("RA velocity of {} and dec_val of {}".format(ra_vel, dec_vel))
 
         if self.runStep['movingTargets']:
+            print("Adding moving point sources to seed image.")
             mov_targs_ptsrc, mt_ptsrc_segmap = self.movingTargetInputs(self.params['simSignals']['movingTargetList'],
                                                                        'pointSource',
                                                                        MT_tracking=tracking,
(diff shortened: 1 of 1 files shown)


bhilbert4:
@laurenmarietta do you have time to give this a quick look? There are really only two lines that changed to fix the bugs.

bhilbert4:
I added a timing statement for moving targets. It will report expected time separately for point sources, galaxies, and extended targets. I think it's most useful for point sources, where all of the stamp images are the same size. For galaxies and extended sources, the stamp images can vary widely in size and therefore computation time when adding to the 4D seed image. But, I agree that alerting the user that the code is still running rather than hung is useful, even if the time estimate may be off somewhat. What do you think @laurenmarietta?

bhilbert4:
Resolved one small conflict via the github editor.

PR #180, opened 2018-10-03
Add relic to environment files

bhilbert4:
This PR adds the `relic` package to the `mirage` environment files. I created a new environment from scratch for the OSX version. With only minor version updates for a handful of packages, I decided to leave the linux version as-is, and just add `relic` rather than starting from scratch.


bhilbert4:
Minimal review needed here, it would seem...

PR #186, opened 2018-10-10
WIP: Add background as user input in yaml_generator. Default to 'low'

bhilbert4:
This PR makes some small changes to `yaml_generator.py`. The background rate is added as a user input, and defaults to 'low' if the input is not provided. Previously this information was gathered from the observation list file. 

Now that I think about it though, what about the case of an observation with multiple epochs where the background may be different between the epochs? That situation would be better addressed using the observation list file. Perhaps we should go back to getting this info from the observation list file, and default to 'low' if it is not in there? 

I have also removed references to the `ExtendedCenter` parameter. This parameter is not used in the seed image generation. The location on the detector to place extended targets is collected from the extended target catalog.

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -2959,14 +2959,6 @@ class Catalog_seed():
             else:
                 raise ValueError("WARNING: grism_source_image needs to be True or False")
 
-        # Location of extended image on output array, pixel x, y values.
-        try:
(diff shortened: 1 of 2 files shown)


laurenmarietta:
I would agree that if the background rate it something that varies from observation to observation, it would be better to include it in the observation list rather than in `yaml_generator`.

bhilbert4:
Outdated. Closing.

PR #187, opened 2018-10-10
WIP: Add several intermediate outputs as attributes to imaging_simulator.py

bhilbert4:
This PR adds some intermediate outputs from `imaging_simulator.py` as class attributes, so that the user has access to them after the code has been run. 

Currently this includes the seed image, the seed segmentation map, a dictionary of information about the seed image, and the linearized dark.

We should make similar additions to wfss_simulator.py

### DIFF
--- a/mirage/imaging_simulator.py
+++ b/mirage/imaging_simulator.py
@@ -36,8 +36,6 @@ from .utils import read_fits
 
 class ImgSim():
     def __init__(self):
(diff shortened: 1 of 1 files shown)


bhilbert4:
@laurenmarietta, is there anything else you'd like to see?

laurenmarietta:
Nope, looks great!

PR #188, opened 2018-10-11
Correct test_apt.py

Johannes-Sahlmann:
The main test in test_apt.py fails even when run in a local environment with access to central store.
The test consists in comparing numbers of observations and files written, but there seem to be inconsistencies:
```
22 exposures total.
150 output files written to: /Users/jsahlmann/jwst/code/github/spacetelescope/mirage/tests/test_data/NIRCam/APT_NIRCam_out
PROPOSAL_ID: 1111
number of observations: 17
number of files written: 150
```

I suggest some changes to the test that makes it pass (locally), but we should probably improve it to contain a `==` assertion.

### DIFF
--- a/mirage/yaml/yaml_generator.py
+++ b/mirage/yaml/yaml_generator.py
@@ -435,7 +435,7 @@ class SimInput:
         base : str
             JWST formatted filename base (excluding pipeline step suffix and ".fits")
         """
-        proposal_id = input_obj['ProposalID']
+        proposal_id = '{0:05d}'.format(int(input_obj['ProposalID']))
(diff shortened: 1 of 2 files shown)


Johannes-Sahlmann:
@bhilbert4 are you willing to approve this and merge? If yes, I can update my APT interface branch and use it as test for NIRCam.

Johannes-Sahlmann:
To be more stringent, should I hardcode an assertion of the current numbers (17 obs, 150 files) in the test?

bhilbert4:
I think that makes sense.

PR #192, opened 2018-10-19
Adding tests for CreatePSFLibrary

shanosborne:
I wrote tests to check psf_library.py. I ran a coverage test and now this file is 97% covered. I also made some PEP-8 improvements to CreatePSFLibrary - mostly it was re-arranging the order of the methods, so it looks like a lot of edits but there should be no changes to how it is run (which I checked with the new tests!)


mirage/psf/psf_library.py
```diff
@@ -1,120 +1,12 @@
-import os
+import astropy.convolution
 import itertools
-
-import webbpsf
 import numpy as np
-import astropy.convolution
+import os
+import webbpsf
 from astropy.io import fits
```
laurenmarietta: A nitpicky thing since you're implementing PEP8 changes here: I believe the official PEP8 style is to separate imports by category (standard library, third party, then local) and then order alphabetically within that. So this would rather look like:
```
import itertools
import os

import astropy.convolution
from astropy.io import fits
import numpy as np
import webbpsf
```

tests/test_psf_library.py
```diff
@@ -0,0 +1,207 @@
+import astropy.convolution
+import numpy as np
+import os
+import pytest
+import webbpsf
+from astropy.io import fits
+from mirage.psf.psf_library import CreatePSFLibrary
```
laurenmarietta: Same comment here!

tests/test_psf_library.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+# @pytest.mark.skip()
+def test_all():
```
laurenmarietta: I might make this test name a little more descriptive, like `test_all_filters_and_detectors`

tests/test_psf_library.py
```diff
@@ -0,0 +1,207 @@
...
+    fgs.detector = "FGS1"
+    fgs.filter = "FGS"
+    fgs.detector_position = (locx, locy)
+    calcpsf = fgs.calc_psf(oversample=oversample, fov_pixels=fov_pixels)["OVERDIST"].data
+    kernel = astropy.convolution.Box2DKernel(width=oversample)
+    convpsf = astropy.convolution.convolve(calcpsf, kernel)
+
+    # Compare to make sure they are in fact the same PSF
+    assert gridpsf.shape == calcpsf.shape
+    assert np.array_equal(gridpsf, convpsf)
```
laurenmarietta: This is really cool. :)

PR #201, opened 2018-11-05
Catalogs with mulitple filters

bhilbert4:
These changes allow source catalogs to contain magnitude columns for multiple instruments/filters. The motivation behind this is to allow for fewer catalog files that users need to keep track of and place into the observation list file.

Mirage will read in the catalog and search for the appropriate column to use for a given simulation. It will first look for a column with the name <instrument>_<filter>_magnitude (e.g. "niriss_f090w_magnitude"). If that column is present, it will be used. If it is not present, Mirage will search for a generic 'magnitude' column (i.e. what is expected currently in Mirage catalogs). If neither column name is present, an error is raised.

Note that for FGS, the first column name that is searched for is 'fgs_magnitude' since FGS has no filters.

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -640,13 +640,13 @@ class Catalog_seed():
         return non_sidereal_ramp, nonsidereal_segmap
 
 
-    def readMTFile(self, file):
+    def readMTFile(self, filename):
         """
         Read in moving target list file
 
         Arguments:
         ----------
-        file : str
+        filename : str
             name of moving target catalog file
 
         Returns:
@@ -659,7 +659,7 @@ class Catalog_seed():
                 are in units of pixels/hour. If false, arcsec/hour
             magsys -- magnitude system of the moving target magnitudes
         """
-        mtlist = ascii.read(file, comment='#')
+        mtlist = ascii.read(filename, comment='#')
 
         # Convert all relevant columns to floats
         for col in mtlist.colnames:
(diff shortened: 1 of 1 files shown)


bhilbert4:
I went back and forth on this for how the function should work. At first I thought it should remove all magnitude columns other than the column to be used, in order to save memory. (I was envisioning a case where someone has a massive catalog file with thousands of sources and all filters.) But then I figured that the function which crops the catalog down to an area just around the detector would be helping reduce memory usage anyway, and so this function could just return the name of the magnitude column to be used.

mirage/seed_image/catalog_seed_image.py
```diff
@@ -1877,6 +1883,59 @@ def readPointSourceFile(self, filename):
...
+            Name of the catalog file. Used only when raising errors.
+
+        Returns
+        -------
+
+        specific_mag_col : str
+            The name of the catalog column to use for source magnitudes
+        """
+        # Determine the filter name to look for
+        if self.params['Inst']['instrument'].lower() in [nircam, niriss]:
```
laurenmarietta: This line should be: 
`if self.params['Inst']['instrument'].lower() in ['nircam', 'niriss']:`
so the instrument names are strings. I got an error when running this because it didn't know what `nircam` was 😬

laurenmarietta: I'm really confused as to why this didn't fail the CI....

bhilbert4: Or in my tests...
Very strange. Good catch.

bhilbert4: Quotes added.

PR #206, opened 2018-11-05
Adding an option to make NIRISS images for the POM field.

KevinVolkSTScI:
The code changes here allow mirage to make oversized 2322x2322 pixel seed images for the full POM area.  This is useful for SOSS or WFSS simulations, where the sources off the imaging field may produce spectra on the detector.  If one sets the NIRISS mode to "pom" one gets an oversized seed image.  The output ramp is unaffected as the central 2048x2048 pixel part of the seed image is used in the ramp simulation.  This will not work with moving targets.  I have not tested the moving target mode for NIRISS, so that mode should not be used anyway.


KevinVolkSTScI:
The actual NIRISS POM field of view seems to be rotated with respect to the detector by a couple of degrees, and the center of the detector also seems to be offset from the POM center.  Neither of these things are simulated in the code here.  The code changes involved are relatively simple because these complications are not included.

bhilbert4:
Is there something special about 2322 for the number of columns and rows? I ask because at the moment, if a direct image to be turned into a WFSS observation is being simulated, then the size of the array is increased by a factor of sqrt(2) to 2896 pixels in order to keep track of sources just off the detector whose spectrum may extend onto the detector in the dispersed image.  I'm just curious if we can bundle WFSS and POM together and use 2896 for everything?

KevinVolkSTScI:
The above is the size I estimate for the NIRISS POM.  The direct image to spectral shifts can be larger than 137 pixels and in those cases one would not want to use the larger size 2896 pixels as this would have spectra or background in the simulation from sky positions where no light reaches the detector.

mirage/seed_image/catalog_seed_image.py
```diff
@@ -119,6 +119,18 @@ def make_seed(self):
         self.output_dims = (self.nominal_dims * np.array([self.coord_adjust['y'],
                                                           self.coord_adjust['x']])).astype(np.int)
 
+        # change the values for the NIRISS/POM mode.  Add 137 pixels extra space around the main image area, full frame.
```
bhilbert4: I think it would be cleaner to take these lines out of the main function and place them into the `calcCoordAdjust` function, in a separate block for POM observations.

Also, you'll see in there that I use `self.grism_direct_factor`, which is defined  in the `__init__` function. I did this to make it easier on developers in case we ever want to change that factor for the grism. Feel free to set up the pom factor in the same way, or just within `calcCoordAdjust`.

Also, the grism_direct_factor at the moment is put into use if `grism_source_image` is true in the input yaml file. So this should work for NIRISS as well as NIRCam.

mirage/seed_image/catalog_seed_image.py
```diff
@@ -168,9 +180,17 @@ def make_seed(self):
         if ((self.params['Inst']['mode'] in ['wfss','ts_wfss']) & \
             ('FULL' not in self.params['Readout']['array_name'])):
             self.seedimage, self.seed_segmap = self.pad_wfss_subarray(self.seedimage, self.seed_segmap)
-
+         
         # Save the combined static + moving targets ramp
         self.saveSeedImage()
+
```
bhilbert4: Similarly, I think it would be cleaner if these lines were placed in a new function that is then called at the end of the main function.

KevinVolkSTScI: Bryan:  For this second change do you mean put the few lines of code that extract the central 2048x2048 pixels for the seed image and segmentation image into a new subroutine?

KevinVolkSTScI:
I think that I have put in the changes that address Bryan's comments.  Given that estimated the POM size of 2322 pixels is only found in two places in the code and these can be found by searching for "pom" I did not feel that it was necessary to use a variable for the value.  Bryan, let me know if you disagree about this.  I could change it if you think this is bad form.

PR #207, opened 2018-11-07
Catalog index column name and flags. 

bhilbert4:
This is a very small update to `catalog_seed_image.py`. Testing WFSS simulations have revealed that I had a typo in the column name for the index in the galaxies catalog (it was `indexes` rather than `index`). I fixed that, and moved the index checking/generating into a separate function so it's easier to maintain across all catalogs. I also made some PEP-8 changes in the same area. 

I also found that the flag used to denote source positions in pixels rather than RA/Dec was inconsistent between several of the catalogs. It was 'position_pixels' or 'position_pixel'. I've updated the code so that all catalogs use the former.

### DIFF
--- a/mirage/ramp_generator/obs_generator.py
+++ b/mirage/ramp_generator/obs_generator.py
@@ -2424,12 +2424,12 @@ class Observation():
             outModel.zeroframe = np.zeros((numint, ys, xs))
 
         # EXPTYPE OPTIONS
-        # exptypes = ['NRC_IMAGE', 'NRC_GRISM', 'NRC_TACQ', 'NRC_CORON',
+        # exptypes = ['NRC_IMAGE', 'NRC_WFSS', 'NRC_TACQ', 'NRC_CORON',
         #            'NRC_DARK', 'NRC_TSIMAGE', 'NRC_TSGRISM']
         # nrc_tacq and nrc_coron are not currently implemented.
 
         exptype = {"nircam": {"imaging": "NRC_IMAGE", "ts_imaging": "NRC_TSIMAGE",
-                              "wfss": "NRC_GRISM", "ts_wfss": "NRC_TSGRISM"},
+                              "wfss": "NRC_WFSS", "ts_wfss": "NRC_TSGRISM"},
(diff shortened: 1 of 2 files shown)


bhilbert4:
It also appears that DMS has updated their list of allowed exposure types. NRC_GRISM is now NRC_WFSS. I updated `obs_generator.py` to use the latter when creating WFSS observations for NIRCam.

bhilbert4:
@KevinVolkSTScI do you have time to give this a quick read? The changes are fairly trivial, so I'm not sure much testing is needed.

KevinVolkSTScI:
Yes I will take a look at it tomorrow or Friday.

KevinVolkSTScI:
Looking at the code brings up a question about the segmentation map that is produced along with the seed image, because of the max_index code.  Is there some listing of which index number corresponds to which source when there are multiple input source lists?  We may want to use the output segmentation map with grizli since the NIRISS/WFSS is not in mirage, and so we would need to know which source corresponds to which segment.  Even for pue mirage runs that information should be available because the segmentation map is saved with the seed image.  If one has both point sources and galaxies, for example, which list comes first?  This just needs to be documented somewhere, but I do not know the answer.

KevinVolkSTScI:
There is a note about conflicting files, which appears to refer to the changes I made to allow the POM mode for NIRISS.  I thought I merged those changes in properly but something may be wrong.

bhilbert4:
Currently Mirage will check for "index" columns each input source catalog. If the numbers are overlapping, then Mirage adjusts them so they are not. Point sources are done first, then galaxies, then extended objects. The adjusted index numbers are present in the output source list files, and these are what match up with the values in the segmentation map. You're right that this needs to be documented clearly.

I'll also point out that I think all we would need to do to enable WFSS for NIRISS is to include the set of config files with the Mirage reference files. The equivalent of what's in the GRISM_NIRCAM directory. Nor has these files.

bhilbert4:
The conflict looks like it's just the NRC_GRISM -> NRC_WFSS change. I'm not sure why github isn't able to resolve this on it's own, but it's an easy manual fix that I can take care of.

PR #208, opened 2018-11-08
Fixing PSF locations in header, default PSF center, and PEP8 cleanup

shanosborne:
The PSFs locations in the header were not quite correct for PSFs with even-length arrays. There is a subtle change inside calc_psf that shifts even length arrays by 0.5 to center them correctly, and while this change was happening inside calc_psf, it wasn't being documented in the header of the final grid  object. I have now updated the header and updated 2 of the tests to reflect this. 

I also switched the default center location of the num_psfs=1 case to (1023,1023). The technically "correct" center is (1023.5, 1023.5), so before this 0.5 shift update above, the center being (1024,1024) was fine because it was just off by 0.5 (and would be no matter what). But now that that shift is added, the case of a single even-length array PSF would be centered on (1024.5, 1024.5), which is not the most accurate option. So I changed the default center so that a single centered PSF would either be located at (1023, 1023) if odd or (1023.5, 1023.5) if even.

A side effect is that now instead of the header reading DET_YX0 = (1023,1023), it will look like a float and read (1023.0, 1023.0) (or if it's an even length array, (1023.5, 1023.5)).

I also made a bunch of PEP8 updates and clean up to the code.

### DIFF
--- a/mirage/psf/psf_library.py
+++ b/mirage/psf/psf_library.py
@@ -6,6 +6,7 @@ from astropy.io import fits
 import numpy as np
 import webbpsf
 
+
 class CreatePSFLibrary:
 
     # Class variables for NIRCam short vs long wave information:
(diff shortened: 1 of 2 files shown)


mirage/psf/psf_library.py
```diff
@@ -6,118 +6,112 @@
...
-        fileloc: str
+        fileloc : str
             Where to save your file if "save" keyword is set to True. Default of None
             will save in the current directory
 
-        filename: str
+        filename : str
             The name to save your current file under if "save" keyword is set to True.
-            Default of None will save it in the form: "INSTR_FILT_fovp####_samp#_npsf##.fits"
+            Default of None will save it in the form: "{intr}_{filt}_fovp{}_samp{}_npsf{}.fits"
```
bhilbert4: I'm not sure if this was intentional or not, but the designation for "instrument" has changed from `INSTR` in the old version to `intr`.

Issue #190, opened 2018-10-17
Fiducial point override use in APT files


bhilbert4:
I'm working with a NIRCam APT file that contains an observation where FiducialPointOverride is present (with the NIRCamEngineeringIImaging template). When I ran the `yaml_generator`, I'm not sure it was doing the right thing with the FiducialPointOverride.

In my case, the data were defined as being full frame images in all A module detectors. The FiducialPointOverride was set to NRCA3_DHSPIL, and there were accompanying xoffset and yoffset values. From my reading of the [APT special requirements page on JDox](https://jwst-docs.stsci.edu/display/JPP/JWST+APT+Special+Requirements), this means that the target RA, Dec will be placed at the reference location of NRCA3_DHSPIL. But the way the `apt_inputs` interprets this situation is that it specifies that the data to be created is only that for the NRCA3_DHSPIL aperture (rather than full frame for all A module detectors). It also says that the xoffset, yoffset is not applied to the simulated data, although I believe that it is since those offsets are baked in to the dither values in the `pointing` file.

My thought is that in this case, Mirage should be making exposures for all A module detectors, and that the pointing should be correct since the pointing file contains all the necessary information. @laurenmarietta, does that make sense, or am I misinterpreting how FiducialPointOverride is supposed to work?

laurenmarietta:
Yes, I think you are correct. I implemented the FiducialPointOverride part early on in my exploration of `apt_inputs` and I would chalk this up to my own misunderstanding of how that was supposed to work. 

Specifically, I think I ignored the xoffset/yoffset because I didn't know how MIRAGE was using them (you say it's just taken care of in the pointing file?), but these parameters definitely should not be ignored.

I will be happy fix this in a PR ASAP.

In the example you describe, with FiducialPointOverride NRCA3_DHSPIL, would we want to restrict to the module being used (A), or rather to generate yamls for all detectors?

Johannes-Sahlmann:
Please wait before implementing this. This part of the code is part of the general revision in #172.
For instance, I already took out the erroneous behaviour related to FiducialPointOverride.

laurenmarietta:
Oh! Good to know, thank you. (I am working my way through my GitHub notifications and haven't gotten to that PR yet.)

laurenmarietta:
@Johannes-Sahlmann discovered another related problem while testing #172:

[...]

PR #214, opened 2018-11-13
Correctly implement fiducial point override and subarrays

laurenmarietta:
Addresses #190.

This PR re-enables simulation of programs that include a fiducial point override, hopefully accurately this time. It includes a work-around for the particular case of an FGS exposure with a NIRCam aperture override - but no other cross-instrument overrides have been handled, since I'm not sure how best to proceed with them.

(This is a half-WIP; I am not certain these changes are correct or sufficient, but am looking for feedback.)


mirage/apt/apt_inputs.py
```diff
@@ -338,6 +340,39 @@ def create_input_table(self, verbose=False):
         ascii.write(Table(self.exposure_tab), self.output_csv, format='csv', overwrite=True)
         print('csv exposure list written to {}'.format(self.output_csv))
 
+
+    def check_aperture_override(self):
+        instruments = self.exposure_tab['Instrument']
+        apertures = self.exposure_tab['aperture']
+
+        aperture_key = {'nircam': 'NRC',
```
Johannes-Sahlmann: Since this dictionary is of general use (I think there are similar ones in other parts of mirage) I suggest to move it to a generic place (`utils/constants.py`?) so that it can be reused when necessary. 
(The same applies to other general definitions, see the top section of `catalog_seed_image.py` for instance but that's maybe for a separate cleanup PR.)

mirage/apt/apt_inputs.py
```diff
@@ -338,6 +340,39 @@ def create_input_table(self, verbose=False):
         ascii.write(Table(self.exposure_tab), self.output_csv, format='csv', overwrite=True)
         print('csv exposure list written to {}'.format(self.output_csv))
 
+
+    def check_aperture_override(self):
+        instruments = self.exposure_tab['Instrument']
```
Johannes-Sahlmann: Should this function not check for the `FiducialPointOverride` key somehwere and only act if that is `True`? Maybe I am misunderstanding.

laurenmarietta:
I see that the Travis build is failing. I'm wondering whether this might be the problem: https://docs.travis-ci.com/user/common-build-problems/#mac-macos-sierra-1012-code-signing-errors

Has anyone else seen this, or does anyone have an idea how to fix it? I am not sufficiently familiar with Travis to know where to implement the fix that they include at that link...

Johannes-Sahlmann:
I think the main reason for the tests to fail currently is that the travis build times out due to "inactivity". I think this will be fixed with #216  because that will significantly shorten the test running time, in particular for `test_apt_interface`.

mirage/apt/apt_inputs.py
laurenmarietta: Good idea - done for this case. (I saw that you added a `utils/constants.py` file in another PR, so there might end up being a merge conflict down the line.)

mirage/apt/apt_inputs.py
laurenmarietta: Good point! Done. Thanks for pointing that out.

mirage/apt/apt_inputs.py
```diff
@@ -338,6 +341,32 @@ def create_input_table(self, verbose=False):
         ascii.write(Table(self.exposure_tab), self.output_csv, format='csv', overwrite=True)
         print('csv exposure list written to {}'.format(self.output_csv))
 
+
+    def check_aperture_override(self):
+        if bool(self.exposure_tab['FiducialPointOverride']) == True:
```
Johannes-Sahlmann: Apparently, the proper way to check for booleans is to use `if bool(var) is True` 
(https://stackoverflow.com/questions/27276610/boolean-identity-true-vs-is-true)

mirage/apt/apt_inputs.py
```diff
@@ -52,6 +52,7 @@
 
 from . import read_apt_xml
 from ..utils import siaf_interface
+from ..utils import constants
```
Johannes-Sahlmann: I think you did not commit the constants.py file.

mirage/apt/apt_inputs.py
```diff
@@ -338,6 +341,32 @@ def create_input_table(self, verbose=False):
...
+                        obs_num = self.exposure_tab['obs_num'][i]
+                        guider_number = read_apt_xml.get_guider_number(self.input_xml, obs_num)
+                        guider_aperture = 'FGS{}_FULL'.format(guider_number)
+                        fixed_apertures.append(guider_aperture)
+                    else:
+                        raise ValueError('Unknown FiducialPointOverride in program. Instrument = {} but aperture = {}.'.format(instrument, aperture))
+                else:
+                    fixed_apertures.append(aperture)
+
+            self.exposure_tab['aperture'] = fixed_apertures
```
Johannes-Sahlmann: Could we in addition define a new key `pointing_aperture` to retain all the information? e.g.
```
self.exposure_tab['pointing_aperture'] = self.exposure_tab['aperture']
self.exposure_tab['aperture'] = fixed_apertures
```
I am not sure if this will satisfy all cases but we will have an available record of the aperture that is specified in the pointing file.

laurenmarietta:
Programs 1141, 1142, 1147, and 1148 all have overrides. I don't think all of those are currently included in `test_apt_interface`, but they are in the `test_programs` directory.

Additionally, I believe a number of the NIRCam/WFSC templates in `test_apt` should have overrides, too.

mirage/apt/apt_inputs.py
laurenmarietta: Sorry about that - just added it.

mirage/apt/apt_inputs.py
laurenmarietta: I added this in the way you described here. One thought is that it's within the if statement that checks if a given program has an override, and so the `self.exposure_tab['pointing_aperture']` key will only be available if `self.exposure_tab['FiducialPointOverride']` is True.

bhilbert4:
Good! I think it makes sense that we have at least one proposal with an override that is present in our tests.

mirage/apt/apt_inputs.py
Johannes-Sahlmann: Yes, good catch. That is not good. `self.exposure_tab['pointing_aperture']` should always be populated.

laurenmarietta: Okay, this should be fixed now!

laurenmarietta:
While working on my commissioning branch for MIRAGE, I had a concerning realization - that I don't see how the information from the pointing file actually gets passed to `catalog_seed_image.py`. 

When looking through the yamls, they specify the aperture of the given exposure, the RA/Dec of the target, and the V2/V3 offset of the original aperture. However, how does MIRAGE know that the target should be put at the center of the (for instance) NRCALL_FULL aperture when the yaml specifies NRCA3_FULL - if NRCALL_FULL isn't listed anywhere in the yaml? Maybe I am missing something obvious here, but at the moment I am quite perplexed.

(I want to make sure I understand how this works also so I can double-check that the tricky situations of fiducial point overrides are being handled correctly, too.)

bhilbert4:
There's a subtle difference between the target RA, Dec in the APT file, and the RA, Dec values placed into the input yaml files. The values in the yaml files are the RA, Dec values corresponding to the reference location of the detector specified in that yaml file, and not the RA, Dec of the target. So for a case where you have NRCAALL, you'll see that none of the RA, Dec values in the yamls match the target RA, Dec from the APT file, because with an NRCAALL exposure, the target RA, Dec is placed in the gap between the 4 SW detectors. The values in the yaml files come from the pointing file, with shifts made to go from the NRCAALL reference location to the reference locations of the individual detectors.

Johannes-Sahlmann:
@laurenmarietta do you mind if I resolve that conflict to see whether the travis build succeeds now?

laurenmarietta:
@bhilbert4  Ah, thank you for that explanation! That makes much more sense now.

@Johannes-Sahlmann Please feel free to resolve them - or, let me know if you would like me to take care of it.

Johannes-Sahlmann:
Travis build is passing now. Next, I will run this branch on the OTE-10 program which has instances of fiducial point overrides.

Johannes-Sahlmann:
I found some issue with an OTE-10 test program, which I submitted in a PR on @laurenmarietta fork.

Issue #215, opened 2018-11-14
Mirage not grabbing filter name of parallel instrument


bhilbert4:
I'm running mirage on program 1071, as a check for my edits to speed up the yaml generator (#202). When using mirage to simulate the yaml files created for this program, I've found that the requested filter for the parallel instrument (NIRISS in this case) is not being propagated into the yaml files. Filters and pupil values are all `None` despite being `F200W` in the APT file.

bhilbert4:
Fixed!

PR #216, opened 2018-11-14
Speed up siaf interface

bhilbert4:
These edits speed up the use of `siaf_interface` when running Mirage. Currently a new instance of pysiaf.Siaf() is created for each exposure when running the `yaml_generator`, which takes a lot of time. 

With these updates, Mirage creates a single instance of the siaf for each instrument in the proposal, and uses these when looping through the exposures.

### DIFF
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -282,7 +282,6 @@ class AptInput:
         #     readxml_obj = read_apt_xml.ReadAPTXML()
         #     tab = readxml_obj.read_xml(self.input_xml)
 
-
         # This affects on NIRCam exposures (right?)
         # If the number of dithers is set to '3TIGHT'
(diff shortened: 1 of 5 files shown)


bhilbert4:
I have tested these updates using program 1071, which uses NIRCam with NIRISS in parallel. The yaml generation was much faster than previously, and was successful for all exposures. I tried simulating an example NIRCam and NIRISS exposure as well. The NIRCam exposure was successfully simulated. The NIRISS exposure crashed, because if appears that the specified filter from the APT program was not propagated into the yaml file. This is being tracked under #215, and not related to these updates, so I think we should keep it separate.

bhilbert4:
Thanks. Seed values are random numbers generated in yaml_generator, so it makes sense that they are different for different runs of yaml_generator.py.

PR #217, opened 2018-11-14
Correct Filter setting for NIRISS Imaging and ExternalCalibration templates

Johannes-Sahlmann:
Implement correct Filter setting for NIRISS Imaging and ExternalCalibration templates.
Add 1071 to test programs.


Johannes-Sahlmann:
Thanks for reporting #215 
This was caused by the different formatting of the Filter information between NirissImaging and NirissExternalCalibration tamplates.

bhilbert4:
I'm still seeing filter=None when I run 1071 with this update, although I don't understand why.

Johannes-Sahlmann:
In my tests, this in now handling the filter/pupil settings correctly in both the observation_list and the individual yaml files.

bhilbert4:
The `filter` option looks good now in my test, but `dark_prep` is failing because both the `dark` and `linearized_darkfile` options are set to None. As with the filters, this is only true for the NIRISS files. The `linearized_darkfile` is properly populated for the NIRCam files.

Johannes-Sahlmann:
Yes, this is because the `mirage_data/fgs/darks/linearized/` and `mirage_data/niriss/darks/linearized/` directories are empty. I therefore hardcoded the the `linearized_darkfile` value to `None`.
We should probably populate these directories. 

@KevinVolkSTScI would you be willing to select the appropriate files and put them in those directories?

KevinVolkSTScI:
I will find something to populate those directories, yes.  I have been a little reluctant to do so only because the pipeline still is not working well and I am afraid that it will produce bad "linearized" darks in some cases.  On the other hand we need something so that the .yaml files which are created are OK both for the specific APT file and as templates for other models as the users wishes.

KevinVolkSTScI:
I have linearized the seven Guider darks that are in the fgs/darks/raw subdirectory of the $MIRAGE_DATA directory with mirage, and copied these to fgs/darks/linearized.  The names are the same as for the raw darks except that "_linearized" is appended before the ".fits" in the name.

PR #218, opened 2018-11-14
Documentation

bhilbert4:
I've had this branch sitting locally for a while now, and I figure it's about ready for others to look over and comment on.  This PR contains sphinx documentation files for mirage. There is no integration with readthedocs yet, although I hope to do that in the future. With this branch, if you go into the `docs` subdirectory and do:

> make html

Sphinx will create a series of html files in the `_build` subdirectory. You can examine the outputs using:

> open _html/index.html

Feel free to make comments or updates. I see at the moment that the checkpoint files for the quickstart notebook are not being ignored. I'll fix that. There are a handful of links that I still need to add. And the `yaml_generator` page will have to be updated to reflect the new code from #172.

So lots of updates still to be made, but I figured that as we are developing mirage and running test cases, the content on these pages could be helpful for us, and also can be updated as we go.


bhilbert4:
Merging an initial set of documentation, so that I can play with readthedocs to get the pages created.

PR #226, opened 2018-11-20
WIP: Run all test cases.

Johannes-Sahlmann:
Modify setup.py to run all defined tests.

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -20,7 +20,7 @@ import pkg_resources
 from asdf import AsdfFile
 import scipy.signal as s1
 import numpy as np
-from photutils import detect_sources
+from photutils.segmentation import detect_sources
 from astropy.coordinates import SkyCoord
 from astropy.io import fits, ascii
 from astropy.table import Table, Column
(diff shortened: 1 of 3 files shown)


Johannes-Sahlmann:
@shanosborne One of the tests `test_psf_library.py` fails because of a missing dependency on webbpsf. Should that dependency be added to `setup.py` ?

@bhilbert4 The other failures are related to this mysterious error:
```
from photutils import detect_sources
E   ImportError: cannot import name 'detect_sources'
```
I was hoping it would just go away ...

shanosborne:
Yes I think including webbpsf in setup.py would fix this. However, the psf library code (and therefore the tests in test_psf_library.py) are dependent on the master version of the webbpsf repo which hasn't been formally released yet. I don't know how you would (or if you can) add the webbpsf repo as an installation requirement, but that is what would need to happen to fix this. In the meantime, you could add `'webbpsf >= 0.7.0'` and fix the import error in travis- all the tests in test_psf_library.py would still crash but for a different reason.

shanosborne:
Yes I think forcing a `git clone + install` should make all the `psf_library.py` run. In that case, it's the spacetelescope/webbpsf repo at https://github.com/spacetelescope/webbpsf

laurenmarietta:
Any chance that changing the import statement to `from photutils.segmentation import detect_sources` might work instead...?

Johannes-Sahlmann:
Closing this as it is obsolete.

PR #233, opened 2018-12-07
WFSS yaml generator update

bhilbert4:
This PR introduces code to enable the yaml_generator to parse APT files containing NIRISS or NIRCam WFSS observations. This includes parallel observations in the case of NIRISS WFSS (NIRCam WFSS is not allowed to be used in parallel at the moment). 

I still need to write some tests and include the test cases I was using for development.

I also updated the catalog inputs to make things a little more convenient for proposals that mix instruments. Previously, the list of input catalogs had to have the same number of entries as there were observations. However, this meant for an observation that had, for example, one NIRCam observation followed by one NIRISS observation, the user had to provide 2 NIRCam and 2 NIRISS catalogs since there were 2 observations. I've updated the code such that the number of required catalogs is instrument-dependent, and equal to the number of observations a particular instrument is used in.


laurenmarietta:
I'm wondering whether it might be better to re-arrange the order of functions inside modules in a separate PR? That could be merged in at a time when there aren't very many PRs open? I just know from experience that rearranging makes comparison with other changes being made in feature branches basically impossible.

bhilbert4:
Ok, this is ready for review. I added test APT files with:
NIRCam WFSS prime, no parallel
NIRISS WFSS prime, no parallel
NIRISS WFSS prime, NIRCam imaging parallel
NIRCam imaging prime, NIRISS WFSS parallel

The test cases cover all the various dither options as well.

@Johannes-Sahlmann @KevinVolkSTScI could you throw some other NIRISS WFSS APT files at this and make sure all looks good?

Johannes-Sahlmann:
@bhilbert4 I opened a PR on your fork with a minor fix. 
I do not have NIRISS WFSS APT files at hand.

KevinVolkSTScI:
I have some NIRISS WFSS APT files so I will test the code with them.  Right now I am in the middle of work on the automatic catalogue generation, so it will be a day or two before I get to testing this.

Johannes-Sahlmann:
@KevinVolkSTScI I noticed that your fork comes off @bhilbert4 's fork and not from the upstream spacetelescope repo: https://github.com/spacetelescope/mirage/network/members
I am not sure if that can explain some of the difficulties that you are having, but it would be simpler to work on a 'standard' fork instead.

KevinVolkSTScI:
This may be the cause; however I was just following the instructions on how to install the fork from the command line before merging, as given on the GitHub page for the pull request, so I would think that it should work.  If one follows the steps the code get downloaded and can be installed, but the conflicts noted on this page are still present as markers in the code source files and hence it does not run.

mirage/yaml/yaml_generator.py
```diff
@@ -388,11 +387,12 @@ def create_inputs(self):
...
-            if self.info['Mode'][i] == 'WFSS':
-                grism_source_image[i] = 'True'
-                grism_input_only[i] = 'True'
+            if self.info['Mode'][i].lower() == 'wfss':
+                if self.info['detector'] == 'NIS' or self.info['detector'][i][-1] == '5':
+                    grism_source_image[i] = 'True'
+                    grism_input_only[i] = 'True'
                 # SW detectors shouldn't be wfss
-                if self.info['detector'][i][-1] != '5':
+                if self.info['instrument'] == 'NIRCAM' and self.info['detector'][i][-1] != '5':
```
laurenmarietta: It looks like this line is causing Travis to fail. I think the instrument key is capitalized (`'Instrument'`).

bhilbert4: Good catch! I've capitalized it in my latest fixes.

bhilbert4:
I just resolved some significant conflicts between this new branch and the current master. The conflicts came from #214. I'm going to re-run lots of tests and inspect output yamls to make sure everything is working as it should.

bhilbert4:
Just a note to say that I fixed a very strange/unexpected bug in `apt_inputs.py` at the step where the pointing dictionary and xml dictionary are joined. For some reason prior to joining, the entries in the xml dictionary were being reordered to put the ObservationID in increasing order. This column has values that are strings of the observation number. (e.g. '1', '2', '10'). I don't understand how or why python was reordering them, but since there were no preceding zeros, the reordering put things in order such as ('1', '10', '11', '12',...'2', '20', '21'). This reordered dictionary was then being joined with the pointing dictionary, which is not reordered. The result was that many rows in the joined dictionary were mixes of different rows in the pointing and xml dictionaries. This wasn't caught in the tests because the lengths of the two dictionaries were unchanged and still equal, and the test cases that looked more deeply at the output dictionaries had <10 observations, so reordering was not an issue.

My fix was to prepend zeros to the ObservationID values such that they are e.g. ('001', '002'). This seems to have solved the problem. I still need to update the truth tables in the tests.

But for anyone who has simulated data from a proposal with more than 10 observations, you may want to check and be sure that you don't have mixed up information in your yaml files.

bhilbert4:
I originally found this error while looking at program 01148. @laurenmarietta

bhilbert4:
Hooray! Conflicts are resolved and tests are all passing locally and on Travis. I think everything is working as it should now. I added several proposals with WFSS observations in them and the results look good.

@KevinVolkSTScI if you get a chance, could you run this on any WFSS proposals that you might have? The new tests include NIRCam WFSS, NIRISS WFSS, as well as NIRISS WFSS in parallel with NIRCam imaging and vice versa, and I think the results are good in all cases. Would be nice to have a sanity check though to make sure I'm not missing anything.

KevinVolkSTScI:
I have a couple of other things on the go right now, but I will look at this once I get a chance and see if I can run some of the NIRISS commissioning WFSS observations through the code.

bhilbert4:
@KevinVolkSTScI I've made a couple minor tweaks to this in the last couple weeks, but I think it's ready to go. Do you have a NIRISS WFSS APT file that you could throw at it? Remember, this PR is only for the yaml file generator, and not the simulation code itself. That's in a separate PR that I'll be commenting on in a few minutes.

KevinVolkSTScI:
@bhilbert4 I will give this a try as soon as I can spare the time....

I have not heard from Nor about the WFSS background images, but I will produce something today or tomorrow which will at least provide a starting point even if it is not quite scaled as Nor wants.

NorPirzkal:
I thought I had replied. The way the WFSS simulator adds the background, it is just an image, in e-/s, that gets added to the dispersed seed image. The default is that the background should be good “as-is” in terms of counts. A user can then specify if it needs to be re-scaled to a certain value. In that case, it gets scaled by new_val/np.max(bck),

On Feb 4, 2019, at 4:48 PM, Kevin Volk <notifications@github.com<mailto:notifications@github.com>> wrote:


@bhilbert4<https://github.com/bhilbert4> I will give this a try as soon as I can spare the time....

I have not heard from Nor about the WFSS background images, but I will produce something today or tomorrow which will at least provide a starting point even if it is not quite scaled as Nor wants.

KevinVolkSTScI:
OK, thank you Nor that clarifies things sufficiently.

bhilbert4:
All tests passing. Output yaml files examined and look good to me and also @KevinVolkSTScI. Merging.

Issue #184, opened 2018-10-09
Catalog generator functions


bhilbert4:
Create some functions for catalog generation.

The motivation behind this was feedback from within the NIRCam team (@marthaboyer) with the idea: 

"Wouldn't it be nice if I could tell Mirage to create a globular cluster catalog rather than having to create one myself."

Catalog creation is one of the more time-intensive things that a potential Mirage user will have to do, so it would be nice to have the ability to create catalogs of representative scenes (stellar fields, extragalactic backgrounds, globular cluster).

Perhaps the first step is a function to create a simple point source catalog following some set of defaults (or user inputs) in terms of luminosity function, source density, etc. for a given pointing. @arminrest suggested that @mcorrenti would be able to provide good default values for stars. We should then think about who to consult as an extra galactic expert in order to get good defaults for those. 

@arminrest also suggested adding the ability for Mirage to ingest catalogs containing magnitudes for multiple filters. Use filter names as column headers, and have Mirage look for those. This would cut down on the number of catalog files needed.

What does everyone think?

laurenmarietta:
Sounds like a great - if complicated - feature!

We do have some skeleton of this currently in `catalogs/get_catalog.py`, though that module only queries point sources in specific WISE and 2MASS bands (corresponding to the F212N and F480M NIRCam filters that we use for commissioning). This module would certainly have to be heavily modified to generate catalogs based on fancy parameters beyond just pointing.

bhilbert4:
Random thoughts on catalog generator development that came to me as I was trying to get the kids to go to bed:
Create a catalog object, and then options for sources to add to it:
1. Add background stars following some luminosity function (Matteo has example using Besancon in astroquery) with typical densities in/out of plane of Milky Way, in bulge.
2. Add background galaxies following some luminosity function at some density
3. Add globular cluster with N stars and some density profile extending over M arcseconds with a typical luminosity function.
4. Generate a catalog based on real stars using astroquery and 2MASS, Vizier, etc (note that most of these catalogs have a fairly low maximum magnitude, so fainter sources will be missing)

Also have functions to:
1. Take an input SED with real flux units and translate into magnitudes through a specific filter
2. Take a normalized input SED and a magnitude through filter X and calculate the magnitude in filter Y

KevinVolkSTScI:
[...]

Issue #50, opened 2018-03-07
Tangent plane projection


bhilbert4:
Investigate whether the source locations are correct, or if a tangent plane projection adjustment is needed. 

@Johannes-Sahlmann already has a function that makes this adjustment.

Johannes-Sahlmann:
This will be required for enabling high-fidelity astrometric simulations. Attached here are the residuals of a NIRISS simulation of the LMC field after fitting a fourth-degree polynomial. I suspect that the ring structure originates in the missing tangent-plane-projection.

[V1087001001P0000000001101_NIS_CLEAR_linear_dms_rate_residuals_sky.pdf](https://github.com/spacetelescope/nircam_simulator/files/2013138/V1087001001P0000000001101_NIS_CLEAR_linear_dms_rate_residuals_sky.pdf)

Johannes-Sahlmann:
@bhilbert4 Was the tangent plane projection implemented already?
I repeated the simulation and analysis with the latest mirage version and do not find the ring structure in the residuals anymore.
In any case, can you please point me to the place in the code where this should be introduced. Functions for tangent plane projection are included in pysiaf, we I will use that dependency.

bhilbert4:
No, tangent plane projection has not been implemented. Maybe the residuals have changed due to the switch to pysiaf?

My first thought is that tangent plane calculations should be introduced within the `XYToRADec` and `RADecToXY_astrometric` functions.

Johannes-Sahlmann:
@KevinVolkSTScI : tangent plane projection was not yet implemented but the gross errors I had seen previously seem to have disappeared (see this thread). 
I will implement it when I have the proper setup to confirm its validity, i.e. after the NIRISS+FGS parallels are finished.

Johannes-Sahlmann:
After re-consideration I state that any projection effects should be absorbed by the distortion coefficients. Hence, this issue can be closed and obsolete.

PR #243, opened 2018-12-20
Catalog generators

bhilbert4:
This PR introduces code to help Mirage users create semi-realistic source catalogs.

This includes the ability to query (and cross-match) the 2MASS, GAIA, and WISE catalogs (in order to add real sources), as well as the Besancon model, which can be used to create a representative population of point sources. The Besancon model goes down to dimmer magnitudes  than the other catalogs, and can be used to fill in the background. 

In addition, the generation of extragalactic background sources is supported by querying the 3d HST catalog of the GOODS-S field, and placing the galaxies at random positions in the field, with small tweaks to magnitudes, Sersic indexes, etc. Currently this only works when requesting a catalog for an area smaller than the GOODS-S field, which is 606,909 arcsec^2 (roughly equal to a 6x6 grid of NIRCam LW detectors.)

Note that this code adds a new dependence to Mirage. The catalog creation function currently calls @Johannes-Sahlmann's  `distortion` package, which is not yet used by Mirage (although was planned to be, via #50). We will have to deal with this new dependence before merging.

Currently we still need to add tests for this new functionality. It would also be good to add some examples to a notebook.

Also still need to compare what the code does to options discussed in #184.


KevinVolkSTScI:
We can take the usage of distortion.py out of the code if needed, it is used for convenience so that if the requested (RA,Dec) is near the pole things will still work properly.  It would require only small changes to the code to take it out.  However, it seems better to use this also in the general scene generation for the same reason.

Johannes-Sahlmann:
I have migrated this functionality into pysiaf and it can be imported from there instead, since pysiaf is already a dependency of mirage: https://github.com/spacetelescope/pysiaf/blob/master/pysiaf/utils/projection.py

bhilbert4:
@KevinVolkSTScI I'm doing a "final" clean-up of the catalog generation code, and putting together a notebook to show how to use it. I was wondering, the get_all_catalogs function currently returns a catalog that contains magnitudes in the requested filters, along with the original Besancon JHKL columns. Is there a reason you left JHKL in the catalog? If you think it's potentially useful for a user to have then I'm all for it. I was just confused to see those columns in there.

Also, I wrote an updated `combine_catalogs` function that uses astropy's `join` function, and more importantly, should work for all Mirage catalog types rather than just PointSourceCatalogs. As far as I can tell, your version of `combine_catalogs` was essentially a DIY version of astropy's `join`. I hope it's ok that I slotted the new version into the code. Right now I still have your function in the code. I just renamed it to `combine_catalogs_v0`.

KevinVolkSTScI:
@bhilbert4 I left the Besancon JHKL values in the catalogue for testing to compare the output magnitudes with the input magnitudes and make sure that the transformation results are doing what I thought it should be doing.  If you are unhappy with this they can be removed, I just think it is useful information.

bhilbert4:
No problem. We can leave them in then.

One other change I made yesterday: The Besancon query was being really finicky with the random coordinates I chose, and was failing with non-intuitive error messages. After some investigation, I added some checks, so that in the case of a Besancon failure, a catalog with a single star with magnitude=99 in all bands is returned. In that way a Besancon failure won't crash `get_all_catalogs`.

bhilbert4:
Ok, let me know what you find. Right now I'm using `join` with `join_type='outer` and with `.filled()` tacked onto the end of the command, which seemed to be doing what I wanted in my tests.

bhilbert4:
@KevinVolkSTScI do you consider this branch ready to merge? I've used all the functions you have written and they look good to me.

KevinVolkSTScI:
From my point of view it is ready to merge.  The 2MASS/WISE/GAIA cross-matching may need some improvement later if I can figure out something to make it do better.

PR #256, opened 2019-02-11
Clarify required dependencies in README

eteq:
I was slightly confused by this: the dependencies listed in the README look to actually be *optional* dependencies, and there's (quite sensibly) a required list in the ``setup.py``.  This just clarifies the situation in the README via linking to the ``setup.py``.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,11 @@ pip install .
 
 ## Dependencies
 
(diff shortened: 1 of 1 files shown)


bhilbert4:
Looks good to me! Thanks for the update.

Issue #255, opened 2019-02-06
Better error message when running yaml_generator without reference files


bhilbert4:
I just tried running the yaml_generator from home after forgetting to log in to VPN, which means I couldn't see the collection of reference files. The error message that came up mentioned something about a bad index being used on a list of superbias files. 

Add in some file checking steps before looking for reference files, so that you can raise a better error message if the reference files are not visible.

bhilbert4:
This issue should be solved by #257

PR #257, opened 2019-02-11
Yaml generator environment variable check

bhilbert4:
This PR pushes the MIRAGE_DATA environment variable expansion function into `utils.py`, as it is called in multiple places within Mirage. It also adds a check that the directory specified by the environment variable exists. It was previously checking that the environment variable was set, but in the case where the specified directory didn't exist (or wasn't accessible due to forgetting to log in to the VPN), the resulting exception wasn't very helpful. This PR changes that so that a more informative error statement is given.

### DIFF
--- a/mirage/utils/utils.py
+++ b/mirage/utils/utils.py
@@ -157,7 +157,7 @@ def ensure_dir_exists(fullpath):
         os.makedirs(fullpath)
 
 
-def expand_environment_variable(variable_name):
+def expand_environment_variable(variable_name, offline=False):
     """ Expand an environment variable and check that the directory exists
 
     Parameters
@@ -178,10 +178,11 @@ def expand_environment_variable(variable_name):
                           "input files needed for the simulation. "
                           "These files must be downloaded separately "
                           "from the Mirage package.".format(variable_name)))
-    if not os.path.isdir(variable_directory):
-        raise FileNotFoundError(("The directory contained in the {} "
-                                 "environment variable: {} does not exist or "
-                                 "is not accessible.".format(variable_name, variable_directory)))
+    if not offline:
+        if not os.path.isdir(variable_directory):
+            raise FileNotFoundError(("The directory contained in the {} "
+                                     "environment variable: {} does not exist or "
+                                     "is not accessible.".format(variable_name, variable_directory)))
     return variable_directory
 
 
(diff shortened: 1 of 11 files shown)


bhilbert4:
This addresses #255

bhilbert4:
Pretty simple change to the code. @Johannes-Sahlmann, do you have a moment to look over the changes and make sure they make sense? I ran the `yaml_generator` and created a simulated ramp using the changes and everything worked fine.

Johannes-Sahlmann:
I will have a look asap.

mirage/utils/utils.py
```diff
@@ -157,6 +157,34 @@ def ensure_dir_exists(fullpath):
...
+        Environment variable name
+
+    Returns
+    -------
+    variable_directory : str
+        Path pointed to by the environment variable
+    """
+    variable_directory = os.environ.get(variable_name)
+    if variable_directory is None:
+        raise ValueError(("WARNING: {} environment variable is not set. "
```
Johannes-Sahlmann: I suggest to remove the `WARNING:` part of the message.

mirage/utils/utils.py
```diff
@@ -157,6 +157,34 @@ def ensure_dir_exists(fullpath):
...
+                          "This must be set to the base directory "
+                          "containing the darks, cosmic ray, PSF, etc "
+                          "input files needed for the simulation. "
+                          "These files must be downloaded separately "
+                          "from the Mirage package.".format(variable_name)))
+    if not os.path.isdir(variable_directory):
+        raise FileNotFoundError(("WARNING: The directory contained in the {} "
```
Johannes-Sahlmann: Same here. I suggest to remove the warning part.

mirage/utils/utils.py
bhilbert4: Removed.

mirage/utils/utils.py
bhilbert4: Removed

bhilbert4:
Arg. These changes cause Travis tests to fail because it now checks for the existence of the MIRAGE_DATA directory. I'll have to add an `offline` flag similar to what we have for the `yaml_generator` I guess.

Johannes-Sahlmann:
I should have waited for the checks to finish before approving ...

bhilbert4:
All tests now passing locally and on Travis!! I had to add 'offline' keywords to all the main files: **catalog_seed_image.py** **dark_prep.py**, **obs_generator.py**, **imaging_simulator.py** and **wfss_simulator.py**.

PR #258, opened 2019-02-11
wrap inverse kernel writer in save_intermediates if block

eteq:
just what it says on the tin: this was failing for me because it was trying to write into a place I didn't have permissions even though `save_intermediates` was off.  But with this change mirage now doesn't try to save the kernel and happily runs through to completion.

### DIFF
--- a/mirage/ramp_generator/obs_generator.py
+++ b/mirage/ramp_generator/obs_generator.py
@@ -1915,17 +1915,18 @@ class Observation():
             newkernel = np.copy(kern)
             newkernel[:, :, ys:ye, xs:xe] = realout1
 
-        # Save the inverted kernel for future simulator runs
-        h0 = fits.PrimaryHDU()
-        h1 = fits.ImageHDU(newkernel)
-        h1.header["DETECTOR"] = self.detector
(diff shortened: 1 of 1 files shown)


bhilbert4:
I think this is a step in the right direction. Thinking about it yesterday though, I was thinking that in this case, where Mirage is trying to save the kernel into the reference files directory, it would probably be best to also check for write permission into that directory. 

And also, in the case of local ST users, we probably don't want everyone trying to write into the single instance of the reference files directory....hmmmm. 

The point of saving this file is so that it can potentially be used in future Mirage runs. And for yaml files generated by the `yaml_generator` (which will probably be the most popular way to run Mirage) the generator looks into the reference file directory for the appropriate files. Maybe I should just find a way to disallow saving this file for those using the ST reference file directory. Not sure how to do that without hard coding the path on central store though...

eteq:
@bhilbert4 - if this is a step in the right direction but it's still not clear what the best direction is, perhaps we could merge this and then do the other directions in a follow-on PR?

bhilbert4:
Works for me.

Issue #251, opened 2019-01-16
Inputs from the APT file can cause Mirage to crash


KevinVolkSTScI:
I was trying to run a .yaml file generated by the APT to yaml interface in Mirage and the code told me that it could not open the parameter file.  (The message was "WARNING: unable to open" followed by the .yaml file name.  It is generated on line 2788 of catalog_seed_generator.py.)  There are two issues here in my opinion:

(1) the cause of the issue is a ":" in the label string of the observation.  I did not know that it was there as the observation is copied from one of the NIRISS CARS.  The APT to yaml conversion copies this label to the parameter file and the resulting line is

  observation_label: Confirmation: Imaging, NRM, GR700XD    # User-generated observation Label

This causes yaml to flag an error because of the second : on the line.  Since this parameter is not used in the code it is irritating that this causes the model to fail.  I would suppose that other such instances might happen in other parameters if a : is present.  The APT to yaml file conversion should screen out these cases in some way.

(2) When the code is unable to read the parameter file it calls sys.exit.  I think this is not the behaviour we want because when running a set of models from a series of parameter files the code will stop at this point whereas it would be better to exit in some other way so that the calculations can continue.  If I am running interactively in Python this kicks me out of the interpreter to the command line and I lose any variables that have been defined as well as having to reload the package when I go back into Python.  This is somewhat irritating, at least for me.  Is there a good reason to call sys.exit rather than just exit?  I think the latter will return one to the Python prompt.

KevinVolkSTScI:
I looked online and I see that sys.exit and exit both do much the same thing.  Nonetheless, if an error is raised in the code that will not cause the end of a Python session, so I suggest that the code raise an error condition rather than exiting.

bhilbert4:
You must have been running with a manually created/edited yaml file? Looking at the code in `yaml_generator`, there is already a check for colons in the proposal title. If they are there, they are replaced with a comma prior to writing out the yaml file.

In terms of avoiding this issue when reading in a manually-created yaml file, the only option I can think of would be a function that reads in the yaml file line by line and replaces/removes any colon after the first, before then moving on to loading the contents of the file into a yaml dictionary.

As for the sys.exit, I'll replace that with a properly raised exception.

KevinVolkSTScI:
[...]

PR #259, opened 2019-02-12
Add a check for colons in some text entries of yaml file

bhilbert4:
Any colons in the Proposal title will cause `yaml.load` to fail.

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -2773,19 +2773,52 @@ class Catalog_seed():
             qy = [float(s) for s in strinstrumentqy]
             self.countvalues = dict(zip(instrumentfilternames, instrumentmag15countrates))
             self.qydict = dict(zip(instrumentfilternames, qy))
-
         except:
-            print("WARNING: Unable to read in {}.".format(self.params['Reffiles']['phot']))
-            sys.exit()
+            raise IOError("WARNING: Unable to read in {}.".format(self.params['Reffiles']['phot']))
 
     def readParameterFile(self):
-        # read in the parameter file
+        """Read in the parameter file"""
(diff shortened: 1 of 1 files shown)


bhilbert4:
This work addresses #251.

@KevinVolkSTScI can you either send me the APT file you used when you were having trouble with the yaml files, or else try out this branch yourself on that file? So far, I made changes to `catalog_seed_generator` to look for and remove any colons in the proposal title string when loading the yaml file. I didn't make any changes to the `yaml_generator` yet, since that already appears to have a fix in it that should remove colons prior to writing the yaml file. I don't think that was added anytime recently though. But at least with this fix, any yaml files that do have colons in the title should be handled.

KevinVolkSTScI:
@bhilbert4 -- I will hunt down the APT file again and forward it to you.

KevinVolkSTScI:
I re-ran a test case with a recent version of mirage (installed around February 3) and the APT -> yaml conversion replaces the : in the title string by a , so the conversion runs without crashing.  Thus this problem was fixed sometime between when I tried in early January and when I last re-installed mirage.

bhilbert4:
That's great news. 
I suppose it's still worth keeping the changes I made last night, in case someone creates a yaml file manually and adds a colon to the title name. It seems like a low-probability situation, and the cost is that yaml files are essentially read in twice in the catalog seed generator now. But as you saw, the error message that comes from a yaml file with an extra colon is pretty tough to decipher. Do you agree?

KevinVolkSTScI:
I agree, keep the code in.  There are other kinds of errors one could make manually in a .yaml file that cause the same error (the other one I have managed to do is not leaving a space after the : in the title field when I changed it...) but we cannot guard against everything.  Having a colon in the title or elsewhere is not that usual and people might do that manually.

PR #260, opened 2019-02-12
Add Mirage-centric info to output fits file headers

bhilbert4:
This PR makes two general changes to the code.

1. In the case of imaging data, it ensures that the NOMXSTART, NOMXEND header keyword values are integers rather than floats. These are not used for imaging data, so this change is simply to be consistent with WFSS mode data, where the values are used by the disperser software.

2. Information about Mirage is now added to the headers of the fits files produced by Mirage. This includes the seed image, the cropped, reordered dark current object, and the final output simulated file. Added information includes the version of Mirage used, the name of the input yaml file, and information from the yaml file that applies to the specific output file.

Both of these changes were motivated by findings from Nor Prizkal as he was working on WFSS data. @NorPirzkal do you want to take a look at the info being added to these files and make sure it covers everything you were hoping to see in them?


bhilbert4:
Trying to clean up some of these pending PRs. @NorPirzkal is there any other info you would like to see in the seed images? See the bottom lines in `catalog_seed_image.py` for a list of what this PR is adding.

bhilbert4:
No response from Nor. Merging this and we can deal with any further tweaks in a separate PR.

PR #268, opened 2019-03-04
Fix CR rate: multiply by frametime

### LINKED ISSUE #267: Cosmic ray probability is scaled inversely to the frame time
I tried running a NIRISS model for the SUB64 subarray (64x64 pixels).  I noticed that the cosmic ray probability per pixel had a much higher value than expected.  From the log of the model I saw

Frametime is 0.05016
Base cosmic ray probability per pixel per second: 0.0012266746411483253

Looking in the code the obs_generator.py file has the following (on line 1795 in the version I am looking at

        self.crrate = self.crrate/self.frametime

I am certain that the rate need to be multiplied by the frametime, not divided by it.  Unless I am mistaken somehow, this is an easy thing to fix.  On the other hand the simulated cosmic rays in all the full frame models that have been made are at too low a rate.  I doubt that this is a real problem given the uncertainties in the cosmic ray rates from the Robberto report which these are based on.  @bhilbert4 do you agree that the frame time should be multiplied by the rate?

bhilbert4:
This PR fixes a typo in determining the cosmic ray rate. Previously the rate (CR/pixel/second) was being divided by the single frame exposure time. This changes it so that Mirage is multiplying by the frame time in order to arrive at the CR rate per pixel per frame.

Closes #267

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
     global:
         - MAIN_CMD='python setup.py'
         - CONDA_CHANNELS='http://ssb.stsci.edu/astroconda'
-        - CONDA_DEPENDENCIES='pytest=3.6 sphinx'
+        - CONDA_DEPENDENCIES='pytest sphinx'
         - PIP_DEPENDENCIES=''
         - PYTHON_VERSION=3.6
         - NUMPY_VERSION=stable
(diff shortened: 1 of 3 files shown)


bhilbert4:
I also moved the multiplication out of the function that retrieves the per second CR rate and up into the calling function. This is just because I liked the idea of having a well-contained function for retrieving the basic, per-second rate.

PR #271, opened 2019-03-04
[WIP] Update setup.py to avoid future conda vs pip installation collisions

bhilbert4:
I'm playing around in this PR to try and limit conda vs pip installation collisions, which just gave us some trouble in the previous PR (). Apparently astropy ci-helpers installs a bunch of packages via conda, but other packages (in `install_requires` in setup.py) are being installed via pip. In some cases the code is installed in different locations, which can lead to a case of python not being able to find a package. 

I'm testing out moving installations to `install_requires` and hoping that nothing else breaks.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -128,13 +128,14 @@ setup(
                   },
     install_requires=[
         'astropy>=3.1',
-        'numpy>=1.13',
-        'matplotlib>=1.4.3',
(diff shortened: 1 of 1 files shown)


bhilbert4:
setup.py is in good shape at the moment, so I'm closing this.

PR #273, opened 2019-03-05
AMI simulations tests

dthatte:
Tests to chaeck AMI simulations with Mirage.

Run createyaml.py, then run updateyaml.py, then run createsim.py.


mirage/seed_image/catalog_seed_image.py
```diff
@@ -2033,6 +2033,15 @@ def makePos(self, alpha1, delta1):
         delta2 = "%1s%2.2d:%2.2d:%7.4f" % (sign, decd, decm, decs)
         alpha2 = alpha2.replace(" ", "0")
         delta2 = delta2.replace(" ", "0")
+
+        # The following fixes a format issue; it is not clear why the signa
```
bhilbert4: Did you see a case where this was needed? This was in `catalog_seed_generator` a long time ago, but was removed when it was determined it wasn't needed.

KevinVolkSTScI:
Hi Bryan,

  This was needed in an earlier version when the + and – signs in the positions in the object catalogues came out “0+0” and “0-0”.   As noted in the comment with the code, I was never able to determine why the extra spaces were present because the format was just %1s and the input was either “+” or “-‘.

  At the moment it appears that the problem that motivated the change is not present and so these lines are not needed.

tests/test_apt_interface.py
```diff
@@ -23,15 +23,15 @@
...
-# importlib.reload( siaf_interface )
-# from mirage.yaml import generate_observationlist, yaml_generator
+from mirage.apt import read_apt_xml, apt_inputs
+from mirage.utils import siaf_interface
+import importlib
+importlib.reload( yaml_generator )
+importlib.reload( generate_observationlist )
+importlib.reload( read_apt_xml )
+importlib.reload( apt_inputs )
+importlib.reload( siaf_interface )
```
bhilbert4: Are these module reload statements necessary?

mirage/seed_image/catalog_seed_image.py
Johannes-Sahlmann: Following @KevinVolkSTScI comment I will remove these lines.

tests/test_apt_interface.py
Johannes-Sahlmann: only for debugging. removed them

PR #280, opened 2019-03-26
Choose correct pixel phase file from PSF library

### LINKED ISSUE #124: PSF positions are not exactly as requested due to code moving the peak of the PSF to the center of the PSF image
The WebbPSF input images for NIRISS do not always have the peak pixel of the 301x301 pixel PSF image at pixel (151,151) [using 1-based pixel numbering).  This characteristic of the PSF files is as it should be in cases where the star position is offset from the center of the pixel in the middle of the PSF image.  In the code within the catalog_seed_image.py script when a PSF image is read in the peak pixel is moved the center of the PSF stamp image (i.e. lines 1798 to 1800 where nxshift and nyshift are found).  When this is done, star positions may get moved off from where they should be by 1 pixel in x or in y (or both).  One should not offset the peak of the PSF file from WebbPSF in this way, in my opinion.  I had to take this shift out of the code to get a single star at the reference (RA,Dec) to be put at the correct positions in the various NIRISS sub-arrays.

bhilbert4:
This PR fixes a bug in `catalog_seed_image.py` where the wrong file was being chosen from the PSF library for a given pixel phase. The code was originally written such that a pixel phase of 0.5, 0.5 was defined as the center of a pixel. This is inconsistent with the definition used in `pysiaf` and the pipeline, where 0.0, 0.0 is the center of the pixel, and -0.5 and 0.5 are the edges.

This fix changes the pixel phase definition to agree with that used by Mirage's dependencies.

Closes #124 
@KevinVolkSTScI

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -1876,6 +1876,11 @@ class Catalog_seed():
                 print(j2, i2, l2, k2)
                 print('bad high')
 
+
+            print('i1, i2, j1, j2, k1, k2, l1, l2')
+            print(i1, i2, j1, j2, k1, k2, l1, l2)
+
+
             try:
                 psfimage[j1:j2, i1:i2] = psfimage[j1:j2, i1:i2] + webbpsfimage[l1:l2, k1:k2] * counts
                 # Divide readnoise by 100 sec, which is a 10 group RAPID ramp?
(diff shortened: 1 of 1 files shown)


bhilbert4:
I also added a warning for the case where the input PSF file has an even number of columns or rows. An odd number is preferred, but with an even number, Mirage will use the pixel just to the left and/or below center as the center of the array.

bhilbert4:
Tests from @KevinVolkSTScI show that the correct PSF files are now being selected.

Issue #274, opened 2019-03-07
Problem when using NIRISS .asdf distortion reference file for AMI


Johannes-Sahlmann:
It appears that there is a problem when using the .asdf distortion reference file for AMI simulations, in the sense that the sources do not appear in the expected location. When mirage uses pysiaf instead, the source locations are as expected.

Example code that can exhibit the problem is available on a development branch:
https://github.com/Johannes-Sahlmann/mirage/blob/sahlmann-ami-tests/examples/niriss_ami_example/niriss_ami_example.py

bhilbert4:
I noticed a similar situation last week when trying to make NIRCam TA subarray data. Pysiaf gave the correct position, while mirage did not. I thought at the time that it was a problem with the reference files, but since you're seeing it as well it seems like there must be a bug somewhere in Mirage.

Johannes-Sahlmann:
Good to know.

PR #285, opened 2019-04-11
[DO NOT MERGE]: Source locations within subarrays

### LINKED ISSUE #264: Incorrect positions from dist. reffile for subarrays
As I was creating some target acquisition data for some engineering tests, I found that the source position returned by the current distortion reference files (which are correct in terms of the round trip positions), were returning incorrect source locations. I was creating data using the apertures: `NRCA5_TAGRISMTS32`, `NRCA2_TAMASK210R`, and `NRCB5_TAPSIMG32`. In all cases I was using a catalog with a single point source placed at the RA, and Dec corresponding to the reference location of the subarray. When using the distortion reference files to calculate detector position, in all cases the sources were outside the subarray entirely. When switching to use `pysiaf`, the returned positions appeared correct. This is not necessarily Mirage-related, but investigate the current distortion reference files to see what the problem is.

bhilbert4:
This PR fixes a bug specific to subarray simulations.

In the case where sources were specified by RA, Dec values in the input catalogs and the distortion reference file (rather than pysiaf) was used to translate RA, Dec to x, y, the sources were being placed in the wrong location. This is because the distortion reference file translates to x, y in the full frame coordinate system. This was causing sources to be shifted from the subarray coordinate system.

In this fix, when the distortion reference file is used to calculate x, y values, the coordinates of the lower left corner of the subarray on the full detector are subtracted from the x, y values, in order to place them in the subarray coordinate system.

I have tested this fix with NIRCam and NIRISS examples and the results look correct.

closes #264 
Also most likely solves #274


bhilbert4:
@Johannes-Sahlmann, do you have a few minutes at some point to check and see if this solves the issue you were seeing with the subarray data?

Johannes-Sahlmann:
Did you test this with all NIRCam/NIRISS apertures, i.e. covering apertures with different DetSciYAngle and DetSciParity? 
I am wondering whether it would make sense to add a dedicated test for this, i.e. include the reference files that contain the same information as the SIAF and test against pysiaf?

bhilbert4:
This fix assumes that the reference files are doing the correct calculations, but that the results are given as (x,y) in the full frame coordinate system, rather than in the subarray coordinate system. In that case, all that's needed should be a shift in coordinates equal to the location of the subarray's lower left corner in the full frame coordinate system. 

I do think a dedicated test makes sense though.

bhilbert4:
Clearly a PR that's been open for a while. All the gridded PSF library updates were added when I rebased.

Anyway, all of the new tests are now passing. I've tested the updates on several NIRCam detectors as well as NIRISS, in order to cover different values of DetSciYAngle and DetSciParity. In all cases, regardless of whether the catalog inputs are in units of RA, Dec or pixel x, y, the sources are landing in the correct locations on the subarrays when the distortion reference file is used.

@Johannes-Sahlmann do you want to test this?

Johannes-Sahlmann:
I am nor sure that I am best positioned to test this as I am exclusively working with full SCA apertures. 
For the NIRISS AMI case, there is an additional complication in that the pipeline distortion reference file appears to be erroneous in its V3IdlYAngle value, which may complicate the test.

bhilbert4:
Replaced by #352. This branch was old enough that rebasing was causing headaches, so the changes were added to a new branch and merged.

PR #295, opened 2019-04-30
Reference file downloader

bhilbert4:
This PR adds a script that allows users to download the reference files needed to run Mirage. Downloads are from Box. Files are downloaded and decompressed/untarred and organized into a `mirage_data` directory. The user is then informed of how to set the `MIRAGE_DATA` environment variable so that Mirage will know where to look for the files.


Johannes-Sahlmann:
This looks like a nice solution to the problem. How do you get the files on box to be static and public?

bhilbert4:
@Johannes-Sahlmann it's a very manual and carpal tunnel syndrome-inducing process. After uploading the files to Box, it takes 5 or 6 clicks on each file to enable a static URL that anyone can access. The URLs are not programmatically discoverable, so they have to be hard-coded into the code.

Johannes-Sahlmann:
Then, what happens if a file has to be updated? Does the URL remain the same?

bhilbert4:
Yes, the URL stays the same. Box does keep track of file versions though. Presumably there is a way to recover the older version if needed.

Johannes-Sahlmann:
Looks like a great solution (despite the overhead in setting up)

bhilbert4:
Merging. With the downloader script added, users can now download the reference files, and Mirage is fully public.

Issue #98, opened 2018-05-31
Add support for time series data/transiting exoplanets


bhilbert4:
Kevin Stevenson mentioned the Batman package (https://www.cfa.harvard.edu/~lkreidberg/batman/) that can be used to create a simulated lightcurve for a transiting exoplanet.

Also may need to look into file formats, assuming that time series data are composed of many integrations that may need to span multiple files.

bhilbert4:
Moving this up to high priority, as there is a summer student coming in June to create TSO datasets.

bhilbert4:
Related issue: #294

PR #298, opened 2019-05-17
Add TSO mode simulations

bhilbert4:
This PR introduces code to enable TSO (time series observations) with Mirage. Specifically, this adds support for sources that vary in brightness with time during the exposure. Similar to what was done for the WFSS mode, users can specify an hdf5 file that contains a lightcurve for a particular target. Mirage reads in the lightcurve and varies the source brightness frame-by-frame based on the lightcurve. As with moving targets, this necessitates a 4D seed image, rather than the normal 2D seed image.

During development, I am working with lightcurves created using the [Batman](https://www.cfa.harvard.edu/~lkreidberg/batman/) repo.

I have also added some support for a new TSO catalog type, and modified the hdf5 code to be able to read in and deal with units for the lightcurves.

My goal is to also enable grism TSO observations in this PR. There are some small differences in this mode, since ideally the spectrum of the target will be varying with time along with the brightness. In that case, I think I will make some adjustments to allow users to input multiple spectra for a given target.

For the first TSO-specific commits though, I've focused only on the easier case of imaging TSO. 

Note that this branch is based off of the griddedPSF model branch (#134), because dealing with the PSFs is much easier in there, but I didn't have time to wait for that branch to be merged before beginning work on this branch.


bhilbert4:
Quick note to say that I have run some regular imaging and wfss observations through Mirage using this branch, and the results all appear unchanged from those constructed using the master branch.

bhilbert4:
Resolves #335  
Resolves #98 
Resolves #339

PR #301, opened 2019-05-29
Add explicit call to disperse method

bhilbert4:
This PR updates the wfss_simulator to include an explicit call to the `disperse` method. This is necessary due to an update to the NIRCAM_Gsim package in order to allow for cached results.


bhilbert4:
One line change, and my test case is working properly. Merging.

Issue #304, opened 2019-06-03
Problems installing MIRaGe OSX environment from file


laurenmarietta:
@shanosborne is trying to re-install MIRaGe. Originally, she was trying to just install with `pip install -e .` and ran into a problem because `jwst-background` wasn't specified in the `setup.py` file (which maybe it should, since it's pip-installable?).

I instructed her to try and set up a MIRaGe environment using the `mirage_environment_osx.yml` file instead, but then she got the following error:
```
Collecting package metadata: done
Solving environment: failed
ResolvePackageNotFound: 
  - nbconvert==5.3.1=py_1
  - qt==5.6.2=h9e3eb04_4
```
She was able to set up the environment successfully after I updated the environment file to say just:
```
  - nbconvert==5.3.1
  - qt==5.6.2
```
But we should maybe investigate updating the environment file so others don't run into this problem? Maybe specifying just the version numbers without the build string (e.g. `h9e3eb04_4`) would help.

bhilbert4:
Resolved, via #305 and #309

bhilbert4:
Oops. #310, rather than #309.

PR #305, opened 2019-06-04
Fix Travis and environment file; Update README

laurenmarietta:
While trying to help a colleague get the MIRaGe environment set up, I discovered a number of problems (including but not limited to #304), which I attempt to fix here. Problems and solutions include:
- I realized that **our environment files included way more packages than necessary**. Most of the packages listed were dependencies of other listed packages, which conda is smart enough to know to install anyway. I created an abridged environment file, `environment.yml`, which (as far as my testing has shown) works for both OSX and Linux systems, and updated the installation docs accordingly.
- Travis was **not using the environment files to build the test environment** - it was just installing with the dependencies listed in `setup.py`. This has been fixed.
- Travis was **not running all of the tests in the suite (!!)**. Only tests that were explicitly listed in the `PyTest` class within `setup.py` were being run - and it didn't seem to me that developers realized this, since only 3 test modules were listed there. I changed Travis to run all of the tests by default, and instead we can explicitly exclude tests with `pytest.mark.skipif` if we don't want them to run on Travis.
- Travis was **only running for Linux** tests. It now runs for both Linux and OSX.
- Since most of our tests had never actually been run with Travis, **many of them were failing**. I did my best to fix this, including:
    - Adding `pytest.mark.skipif(ON_TRAVIS)` to tests that require access to the mirage directory on central store
    - Removing the `test_psf_library` module, since it has been replaced by `webbpsf.gridded_library`
- I also **updated the README.md** to be more concise, since most of the information there has since been put into ReadTheDocs. (If anyone disagrees with this, please let me know and I can put it back!)

There are some tests in `test_catalog_generation` that are still failing - I'm going to tag @bhilbert4 to take a look at those, because it looks like there's an issue with the content of the tests rather than a code structure problem.


bhilbert4:
Thanks for all these fixes!! 
Ugh, those catalog generation tests. As far as I can tell, there is some problem with the Besancon model in astroquery. I've been getting that error message quite a bit lately, after the code working without a problem during development. I'm going to submit a question to the Besancon folks. In the meantime, maybe we should just mark the Besancon-related tests with `pytest.mark.skipif(ON_TRAVIS`.

Also, I have a fix for the yaml loader in #295, although there I used the `full_loader` rather than the `safe_loader`. The difference is probably not important for Mirage, I'm guessing. I can deal with merge conflicts when I merge #295.

laurenmarietta:
Okay, duly noted about the catalog problems. I marked both `test_for_proposal` and `test_get_all_catalogs` with:
```
@pytest.mark.skip(reason="Bug with the Besancon model in astroquery.")
```
so those tests won't run anymore, both on and off Travis.

And I'm also switching the yaml loaders back to the original - sorry, I didn't realize you were changing them in #295! There's no real need for that switch to happen in this PR and it was easy to put them back with PyCharm.

bhilbert4:
I think you probably also need to mark `test_besancon_generation` and `test_2mass_plus_besaoncon_convenience_function` with the `pytest.mark.skip`, as those also query the Besancon model.

laurenmarietta:
Okay, I added the extra skips. And it looks like the builds are passing now, so we're good to go! 😀I'm glad my months of experience fighting with CI across 3 different repos has finally paid off...

And please let me know if you have any questions about any of the changes - in theory I should be able to explain them by now.

PR #307, opened 2019-06-06
Correction of the cosmic ray event rates

KevinVolkSTScI:
The cosmic ray event rates in Mirage were too high by a factor of 10.73676 because they were per frame not per second.  There was also a transcription error in the SUNMIN/SUNMAX values in Volk's original code that is corrected here.  This change corrects the errors and provides documentation of where the values come from.


bhilbert4:
Closes #308

PR #311, opened 2019-06-10
Update obs_generator.py to get the CR rate correct.

KevinVolkSTScI:
As discovered by Chris Willott, the code was multiplying the cosmic ray rate per pixel per second by the frame time twice, once on line 938 and once on line 1021.  Hence the rates are too high for full frame and too low for small subarrays.  The new change removes the second multiplication by the frame time.

### DIFF
--- a/mirage/ramp_generator/obs_generator.py
+++ b/mirage/ramp_generator/obs_generator.py
@@ -1017,9 +1017,6 @@ class Observation():
         # Calculate the rate of cosmic ray hits expected per frame
         self.get_cr_rate()
 
-        # Multiply by the frametime to get probability per pixel per frame
-        self.crrate = self.crrate * self.frametime
-
         # Read in saturation file
         if self.params['Reffiles']['saturation'] is not None:
             self.read_saturation_file()


bhilbert4:
Do you have any opinion on what we should use for the default version of the CR rate? That is, the version written into the yaml files by `yaml_generator`? Right now it's set to be 1.5 times the SUNMIN rate (which I set back before SUNMIN was revised to be a higher rate than SUNMAX). It seems like this is probably too high to use as a default.

KevinVolkSTScI:
We should make the default SUNMAX now that this is the lower of the two values, and I would leave the default scaling at 1.0 personally.  If you want to leave it at 1.5 to get something in between SUNMIN and SUNMAX that is fine too.  The problem is that for many purposes the CR simulation is not needed.  Whatever the default is, the user needs to think about whether to change it.

bhilbert4:
That makes sense to me. Do you want to do that as part of this PR, or should I open a new one?

KevinVolkSTScI:
I do not know off hand where to change the defaults, so it may be faster if you make the change.

bhilbert4:
Sounds good. I can quickly change that in a new PR. I'll merge this one now.

PR #314, opened 2019-06-20
Failing stamp coords

bhilbert4:
This PR makes changes to ``catalog_seed_image`` to fix a bug in which sources that were located off or almost completely off the detector were not being treated properly. 

Updates were made such that if any of the coordinates that describe the location of the source within the aperture are `None`, then the source's stamp image is not added to the seed image.

Also `cropped_coords` was updated to make proper use of the `ignore_detector` flag. If `ignore_detector` is true, then coordinates can be for locations outside the aperture. If it is False, coordinates outside the detector are either reset to be at the aperture edge (in the case of the source falling partially within the aperture) or set to None, if the source is completely outside the aperture.


bhilbert4:
@laurenmarietta, the notebooks all work for me now, with these updates.

bhilbert4:
In the interest of getting the imaging example working for Maryam, I'm going to merge this now. Let me know if you see any more problems.

laurenmarietta:
(This did solve the problem I was seeing in #303. Thank you!)

PR #316, opened 2019-06-20
Update FGS linearized dark list creation

bhilbert4:
This PR introduces the correct search for FGS linearized darks within `yaml_generator` when it is run with `use_linearized_darks` set to `True`. These linearized darks are placed into the output yaml files.


bhilbert4:
@Johannes-Sahlmann, this worked for me in my test. Does it look ok to you?

PR #317, opened 2019-06-24
PSF selection for NIRISS NRM imaging

Johannes-Sahlmann:
This PR introduced a fix to allow mirage to generate NIRISS NRM imaging data with the new PSF selection scheme.


Johannes-Sahlmann:
Good idea!

PR #324, opened 2019-07-08
Add APT example notebook (using create_catalog)

### LINKED ISSUE #323: Add example notebook for nominal simulations from APT
Once #303 is merged, there will be notebook examples for how to simulate APT programs using non-nominal PSFs. However, there is no example for how to use APT to simulate a nominal observation. I imagine users would appreciate that!

laurenmarietta:
This PR:
- Rewrites the unnecessarily-specific OTE Commissioning example notebook to be more broad for any kind of APT program simulation, using the `create_catalog` functionality along the way
- Updates `yaml.generate_observationlist` to work with the output of `create_catalog`

Along the way, I discovered some bugs that I fixed, including:
- Updating `catalog.create_catalog` to work with FGS columns
- Updating `catalog.create_catalog` to successfully create galaxy catalogs (I'm very confused about this because I think I did this myself before, but the bugs I found made me suspicious that maybe it had never actually been running correctly?)
- Updating the dimensions in `seed_image.catalog_seed_image.cropped_coords` so as to never give a negative value for `l2`, which I was getting and was crashing things. It seemed to just be a +1 error.

Closes #323.


bhilbert4:
I was finally able to run the notebook all the way through (after some days with weird query results from GAIA), and everything worked for me. 

I had been concerned about the comment about `l2` above, but in looking over the code, the only change is to a comparison between stamp dimensions and `i1` and `j1` where a `+1` was removed. This change is correct.

Tests are passing, so I think this is ready to merge.

PR #330, opened 2019-07-18
Extended source countrates

### LINKED ISSUE #329: Count rates of extended sources
I tested the count rates in a seed image that includes both point sources and extended sources. The count rates for point sources seem fine, while the count rates for extended sources match the counts per frame instead of the counts per second.

@KevinVolkSTScI identify the problem to be in catalog_source_image.py, at the following lines:
           # First create the galaxy
            stamp = self.create_galaxy(entry['radius'], entry['ellipticity'], entry['sersic_index'],
                                       xposang*np.pi/180., entry['counts_per_frame_e'])

where entry['counts_per_frame_e'] should be entry['countrate_e/s'].

bhilbert4:
This PR fixes a bug in which the count rates of the galaxy and extended stamp images were being set to the number of counts per frame. The code has been updated to use counts per second instead, in line with the point sources and what Mirage expects.

Closes #329


bhilbert4:
@camipacifici, should I merge this?

bhilbert4:
Note that these changes are contained within #347, which contains some additional updates. I'll close this PR in favor of #347.

PR #332, opened 2019-07-23
Remove oversampling squared flux scaling in gridded library

bhilbert4:
This PR changes the flux scaling that is done to the gridded PSF library that is read in. Previously, there was a discrepancy between the assumption in `photutils` that a gridded PSF library has a total signal level of `oversampling^2`, while libraries produced by `webbpsf` had a total signal of 1.0. In that case, Mirage would read in a Webbpsf-generated library and scale the PSF by a factor of oversampling^2.

Webbpsf has now been updated to scale the signal in the library when it is created, so this scaling is no longer necessary in Mirage.

See [webbpsf PR 311](https://github.com/spacetelescope/webbpsf/pull/311) for details.

Along with this update, the gridded PSF libraries that Mirage uses will have to be updated.

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -2064,14 +2064,8 @@ class Catalog_seed():
                                                           coord_sys='full_frame',
                                                           ignore_detector=ignore_detector)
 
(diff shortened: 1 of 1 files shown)


bhilbert4:
Closing, as this is identical to #356

Issue #333, opened 2019-07-23
Ensure setup.py and environment.yaml agree


laurenmarietta:
As @mperrin just discovered, there are inconsistencies between the package's `setup.py` file and the `environment.yaml` file. It should be easy enough to resolve these - I think the `environment.yaml` should be up to date and thus `setup.py` should just be modified accordingly.

PR #334, opened 2019-07-25
Update setup.py to be consistent with environment file

bhilbert4:
This PR updates the dependencies listed in setup.py to be equivalent to those in the environment file, which was updated recently.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,15 +113,15 @@ setup(
                              'config/*.*']
                   },
     install_requires=[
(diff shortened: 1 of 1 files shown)


bhilbert4:
Closes #333

bhilbert4:
Travis is still failing with these updates with what looks like a problem somewhere between astropy/photutils.

bhilbert4:
This PR is no longer necessary. setup.py and environment files have been updated.

PR #342, opened 2019-08-07
Support use of WLP4 filter and weak lens PSFs

mperrin:
- Add WLP4 to a list of NIRCam filters in two different places
 - Ignore weak lenses +8 and -8 and the DHS gratings when finding lists of filter names. 

Related to #340  and #341.  This still won't work on its own until the rest of #340 is taken care of.

### DIFF
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -850,6 +850,10 @@ def get_filters(pointing_info):
     """Return a dictionary of instruments and filters contained within a
     pointing dictionary from an APT file.
 
+    This function is aware that sometimes filters are installed in pupil wheels,
+    and also that some wheels contain non-filter items such as the NIRCam
+    weak lenses. It will return a list of those items that are indeed spectral bandpass filters.
+
     Parameters
     ----------
     pointing_info : dict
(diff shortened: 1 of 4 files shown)


mirage/psf/psf_selection.py
```diff
@@ -291,7 +291,7 @@ def get_library_file(instrument, detector, filt, pupil, wfe, wfe_group,
             file_filt = header['FILTER'].upper()
 
             try:
-                file_pupil = header['PUPIL_MASK'].upper()
```
mperrin: Note, I don't understand how this has been working before, since PUPIL_MASK has never been a keyword that WebbPSF writes; in fact it's not even a valid keyword at all (too many characters). I think it has always been failing back to CLEAR via the exception case, whenever this function is invoked currently.

mirage/apt/apt_inputs.py
```diff
@@ -878,6 +882,9 @@ def get_filters(pointing_info):
 
             filter_list = list(set(short_pupils))
             filter_list.remove('CLEAR')
+            for wfsc_optic in  ['WLP8', 'WLM8', 'GDHS0', 'GDHS60']:
+                if wfsc_optic in filter_list:
+                    filter_list.remove(wfsc_optic)
 
             filter_list.extend(list(set(long_pupils)))
```
bhilbert4: I think it makes sense to do something similar to `long_pupils` and remove GRISMR and GRISMC.

mirage/psf/psf_selection.py
bhilbert4: Yeah. I must have copied the parameter name from webbpsf without thinking about it.

PR #345, opened 2019-08-08
Fix sign errors in PC matrix

### LINKED ISSUE #343: WCS PC matrix sign error
Following up on what @laurenmarietta reported regarding creating mosaics with MIRAGE outputs, I believe (from testing) all that needs fixing are the signs for the PC matrix elements, which can be found in /utils/set_telescope_pointing_separated.py. They should be as follows:

    fheader['PC1_1']  = np.cos(pa_aper_deg * D2R)
    fheader['PC1_2'] = np.sin(pa_aper_deg * D2R)
    fheader['PC2_1']  = -np.sin(pa_aper_deg * D2R)
    fheader['PC2_2'] = np.cos(pa_aper_deg * D2R)

I am curious, however, as to how the pipeline assigns WCS information and whether the mosaic from these outputs is correctly laid out. I don't know if Lauren did that test, but we should check.

bhilbert4:
This PR fixes two sign errors in the calculation of the PC matrix in the `set_telescope_pointing_separated.py` module. These sign errors were causing errors in mosaic creation when using Mirage-created data.

Closes #343


bhilbert4:
@Skyhawk172 are you interested in testing this before merging?

bhilbert4:
Further discussions with @jdavies-st have shown that the equations in Mirage before this change agree with what is used in the JWST pipeline, where `vparity` for NIRCam has a value of -1.
https://github.com/spacetelescope/jwst/blob/master/jwst/lib/set_telescope_pointing.py#L1243

The changes in this PR make the PC matrix equations inconsistent with those in the calibration pipeline.

Skyhawk172:
Interesting. I'll have to think about how to reconcile what we're seeing when stitching the images together into a mosaic. Could it be QUIP's handling of the WCS?

mperrin:
QUIP is just a modified version of Ginga so I would expect it's doing totally standard things with the WCS... 

In any case, I think the right next step is to try running the output files from MIRAGE through the pipeline, and _then_ trying them in QUIP.

jdavies-st:
There's some small discussion here as to the reason for the signs on the terms:

https://github.com/spacetelescope/jwst/pull/1855#issuecomment-378408162

One could try running the images through `reproject` as well as QUIP.  And `montage` while you're at it.

As for the output mosaics made in the JWST pipeline from many input NIRCam detector images, this output frame is computed from the input orientation and scale taken from the PC matrix and CDELTn from the first input NIRCam image.  Same is true of resampling a single NIRCam detector image.

So whatever `mirage` is doing, it is at least consistent with the JWST pipeline.  That said, there could be a mistake somewhere in each, so it would be good to double check this.  I.e. make sure that the position angle has the right sign as well.

pllim:
Unfortunately, mosaic is not very well tested on the Ginga side. I still would advise using Ginga's Mosaic plugin on some real exposures with rotation (e.g., HST or HLA) and see if it produces the correct result. It might very well that you have caught a bug; can never know for sure until you try it on something other than theoretical JWST data.

And if you want to see how Ginga actually calculates it...

Code to grab WCS: https://github.com/ejeschke/ginga/blob/1ab3d1bf2ace3f444d1ecb2251ee79017ca942e1/ginga/util/wcs.py#L212

Code to do transformations: https://github.com/ejeschke/ginga/blob/master/ginga/trcalc.py

Code that ties the math to the images: https://github.com/ejeschke/ginga/blob/1ab3d1bf2ace3f444d1ecb2251ee79017ca942e1/ginga/AstroImage.py#L527

Skyhawk172:
As suggested earlier, I ran the MIRAGE output images all the way through Image2Pipeline and stitched them together with QUIP. The result is unchanged from my initial finding, i.e. the PC matrix elements need to be "fixed" in order for the images to display properly in the mosaic.  Need to investigate further...

ejeschke:
@pllim, @Skyhawk172, there was a bug that affected mosaicing PC-matrix (WCS) images in Ginga. It's fixed in master.  Could you update your installation of `stginga` and test QUIP now?

jdavies-st:
Could you provide a link to the PR fix?

pllim:
I think it is ejeschke/ginga@0c2785652dae09f6a1dac872f47bf853f05421f2

jdavies-st:
So I'm assuming the code worked fine for WCS with a CD matrix, but the bug was there for those (like JWST) that use a PC matrix and CDELTs instead of the CD matrix?

pllim:
Yes, I think so. Nothing was wrong with HST/ACS images and I saw no PC stuff in the header there.

jdavies-st:
Right.  ACS and WFC3 use a CD matrix where the CDELTs are folded into the affine transformation matrix of the WCS.  👍 

The world makes sense!  Relieved it was not a bug in the `jwst` package.  😅

pllim:
Still, would be nice for TEL to test QUIP with dev Ginga to be sure. If it is indeed a Ginga bug that is now fixed, you can close this PR *without* merge and also finally close #343 !

ejeschke:
I should point out that this bug only affected the Ginga quick mosaics, displaying a PC-matrix type image in the usual way was not affected.

Skyhawk172:
Great! I will test this new fix as soon as possible. I will report back as soon as I'm recovered from my laptop completely dying on me.

Skyhawk172:
Okay, I was able to run QUIP with Ginga 3.1.dev51 and STGinga 1.2dev1 and everything is behaving as I expected.  I used MIRAGE images that were processed through the JWST pipeline (thanks to Marcio!).

Using only a smaller set of images, I was also able to compare mosaics generated with QUIP as well as with the JWST pipeline (level 3) and they match nicely, suggesting that QUIP/Ginga are doing the right thing.

pllim:
Thanks for checking, @Skyhawk172 ! Does that mean this PR can be closed *without* merged and #343 closed as well?

Skyhawk172:
Yes, I believe these two issues can be closed without merge. Nothing to change in MIRAGE.

Issue #344, opened 2019-08-08
Header values of RA_V1, DEC_V1 are usually not correct


Johannes-Sahlmann:
The values of RA_V1, DEC_V1 are set to the target's RA, Dec which is incorrect in almost all cases because the target is usually located within an instrument aperture:
e.g. here https://github.com/spacetelescope/mirage/blob/ef140bd0e8cc86e7dec4b8dbc7cf34360bc9f673/mirage/ramp_generator/obs_generator.py#L2588

Instead RA_V1, DEC_V1 should be computed on the basis of the target coordinates and the target's V2,V3 location in combination with the specified `rotation`. It should probably be made more explicit what `rotation` refers to. Currently it is defined as " y axis rotation (degrees E of N)". It should be clarified if this is ROLL_REF or PA_V3.

bhilbert4:
Solved in #346

PR #346, opened 2019-08-08
Fix pointing info and NIRISS PSF wings

Johannes-Sahlmann:
This PR introduces a fix to the values of RA_V1 and DEC_V1 header values.

It also fixes a problem with the selection of NIRISS PSF wing files. This may have untested side-effects when simulating data with NIRCam. I verified that the code works for NIRISS and FGS.

### DIFF
--- a/mirage/psf/psf_selection.py
+++ b/mirage/psf/psf_selection.py
@@ -274,6 +274,9 @@ def get_library_file(instrument, detector, filt, pupil, wfe, wfe_group,
     pupil = pupil.upper()
     wfe = wfe.lower()
 
+    # set default
+    file_wfe = ''
+
     # handle the NIRISS NRM case
     if pupil == 'NRM':
         pupil = 'MASK_NRM'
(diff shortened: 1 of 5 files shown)


mirage/psf/psf_selection.py
```diff
@@ -287,7 +287,10 @@ def get_library_file(instrument, detector, filt, pupil, wfe, wfe_group,
 
             # Compare the header entries to the user input
             file_inst = header['INSTRUME'].upper()
-            file_det = header['DETECTOR'].upper()
+            try:
+                file_det = header['DETECTOR'].upper()
+            except KeyError:
+                file_det = header['DET_NAME'].upper()
```
mperrin: Hmm, DET_NAME vs DETECTOR is probably our fault from writing inconsistent header keywords in different parts of WebbPSF. What keyword will DMS use for this? We can make sure WebbPSF is consistent from now on.

Johannes-Sahlmann:
I checked by simulating NIRISS data that the pointing information is now correct. This ready for review.

mirage/psf/psf_selection.py
bhilbert4: DMS uses `DETECTOR`

bhilbert4:
Closes #344

Johannes-Sahlmann:
OK. I'll have a look at the failing tests.

Johannes-Sahlmann:
Tests are passing. @bhilbert4 can you please merge?

PR #347, opened 2019-08-12
Normalize PSF to convolve with galaxy and extended image stamps

### LINKED ISSUE #329: Count rates of extended sources

bhilbert4:
This PR fixes a bug that was causing galaxy and extended images to have signal levels that were too low. This was due to the PSF that is convolved with the galaxy/extended stamp images. Since we are using just the "core" of the PSF for the convolution, and the full PSF is normalized to 1.0 at the telescope entrance pupil for an infinite aperture, the total signal in our "core" is several percent below 1.0. This was then causing the total signal in the convolved galaxy/extended image to be several percent lower than expected. 

We experimented with two solutions:

- Using a PSF with a larger area for convolution
- Re-normalizing our PSF "core" to 1.0 prior to convolution

Using a larger PSF solved the problem, but at the expense of increasing the computation time by a factor of 2-2.5. We therefore settled on the second option, and now renormalize the PSF prior to convolution. 

Closes #329


mirage/psf/psf_selection.py
```diff
@@ -351,7 +356,8 @@ def get_library_file(instrument, detector, filt, pupil, wfe, wfe_group,
             # If so, add to the list of all matches
             if match:
                 matches.append(filename)
-        except KeyError:
+        except KeyError as e:
+            print(e)
```
mperrin: Is just a print sufficient for this error? Would it be better to use warnings.warn() instead?

mirage/seed_image/catalog_seed_image.py
```diff
@@ -3076,22 +3080,27 @@ def make_galaxy_image(self, file):
...
+                                       xposang*np.pi/180., entry['countrate_e/s'])
 
             # If the stamp image is smaller than the PSF in either
             # dimension, embed the stamp in an array that matches
             # the psf size. This is so the upcoming convolution will
             # produce an output that includes the wings of the PSF
             galdims = stamp.shape
 
-            # *******OR SHOULD WE ALWAYS CONVOLVE WITH A LARGER PSF? IF PSF IS SMALL,
-            # THIS WILL ARTIFICIALLY PUSH SIGNAL TOWARDS THE CORE OF THE GALAXY, I THINK.
```
mperrin: This comment here (which is removed in this PR) is correct I think.  Using the smaller PSF normalized to 1 will keep more light near the core of the extended source, compared to the more rigorous approach that uses the full convolution including the wings. Whether this is a problem or not will depend on the relative sizes of the photometry aperture versus the extended source...

mirage/seed_image/catalog_seed_image.py
```diff
@@ -3390,7 +3403,7 @@ def make_extended_source_image(self, extSources, extStamps):
         for entry, stamp in zip(extSources, extStamps):
             stamp_dims = stamp.shape
 
-            stamp *= entry['counts_per_frame_e']
+            stamp *= entry['countrate_e/s']
```
mperrin: Is this an separate unrelated bug that was fixed as part of this PR? Seems unrelated to PSF normalization.

mirage/seed_image/catalog_seed_image.py
```diff
@@ -3076,22 +3080,27 @@ def make_galaxy_image(self, file):
...
             psf_dimensions = np.array(self.psf_library.data.shape[-2:])
             psf_shape = np.array((psf_dimensions / self.psf_library_oversamp) -
                                  self.params['simSignals']['gridded_psf_library_row_padding']).astype(np.int)
+
+
+            # Find the PSF size to use based on the countrate
+            #print('\n\nTEST OF INCLUDING WINGS IN PSF TO CONVOLVE WITH GALAXY\n\n')
+            #psf_x_dim = self.find_psf_size(entry['countrate_e/s'])
+            #psf_shape = [psf_x_dim, psf_x_dim]
+            #print('\n\nTEST OF INCLUDING WINGS IN PSF TO CONVOLVE WITH GALAXY\n\n')
```
mperrin: Should these commented-out lines be removed from this PR, or is it intentional to keep them?

mirage/psf/psf_selection.py
bhilbert4: That's a good idea, since this will catch problems with files that may not be used during a given run of Mirage.

mirage/seed_image/catalog_seed_image.py
bhilbert4: Yes, this is a separate bug that I had fixed in a different branch but needed here as well in order to test for the correct source brightnesses.

mirage/seed_image/catalog_seed_image.py
bhilbert4: I tend to err on the side of leaving anything that may be useful in the future (for example, if we changed our minds and decided that it would in fact be better to use the full PSF for convolution) in the code. But it does make things more messy. I'll pull this out.

mirage/seed_image/catalog_seed_image.py
bhilbert4: Hmm. Maybe I'll leave this comment in the code, as a warning in case it is found to be a problem.

bhilbert4:
My tests of this fix, using galaxy sources placed at various locations across the detector, show differences between the input and output count rate that are <0.1% for NIRISS and <0.9% for NIRCam. The difference depended on the exact location on the detector, similar to what was seen in #134. Output count rates were determined using simple aperture photometry with a radius of 300 pixels on seed images where the background signal was 0.

bhilbert4:
@mperrin do you have any objection to merging this? @camipacifici has been simulating galaxies using Mirage recently and should alert us to any residual problem with count rates.

mperrin:
Fine with me! Go ahead and merge.

 I agree, if anyone encounters issues later that require sim accuracy better than your test results above, we can address that later.

Issue #281, opened 2019-03-29
Discrepancy in the file naming convention between MAST and Mirage.


dthatte:
There is a discrepancy in the file naming convention between MAST and Mirage. 

For MAST data the file names corresponding to two different dither pointings (same filter, ngroup, nint) have the same activity number and different exposure numbers. The exposure numbers are reset when the activity number changes. I have the following files from MAST (from November 2018 NIRISS SI Rehearsal #1 (SIR1) campaign program 793) for an observation of a source with three filters at two different dither positions.  The activity numbers for F480M, F430M and F380M are 02, 04 and 06 respectively. In each activity the exposure numbers are 00001 & 00002 corresponding to two dithers.

**2 dithers with same ngroup,nint for F480M**  
jw00793016001_03102_00001_nis_uncal.fits
jw00793016001_03102_00002_nis_uncal.fits  

**2 dithers with same ngroup,nint for F430M**     
jw00793016001_03104_00001_nis_uncal.fits    
jw00793016001_03104_00002_nis_uncal.fits 

**2 dithers with same ngroup,nint for F380M**
jw00793016001_03106_00001_nis_uncal.fits
jw00793016001_03106_00002_nis_uncal.fits

For data simulated with mirage, for each filter, ngroups,nint combination the exposure number remains the same for all dither pointings but the activity number is different. The activity numbers never reset. For 45 exposures (fits files) I have 45 activity numbers.
I think this is not correct since exposures are part of an activity. 

In the example above all exposures for F480M are numbered 00001, all exposures for F430M are numbered 00002 and all exposures for F380M are numbered 00003. The activity numbers increase from 01 to 0c without reset.

**4 dithers with same ngroup,nint for F480M**  
jw00793001001_01101_00001_nis_uncal.fits
jw00793001001_01102_00001_nis_uncal.fits
jw00793001001_01103_00001_nis_uncal.fits
jw00793001001_01104_00001_nis_uncal.fits

**4 dithers with same ngroup,nint for F430M**          
jw00793001001_01105_00002_nis_uncal.fits
jw00793001001_01106_00002_nis_uncal.fits
jw00793001001_01107_00002_nis_uncal.fits
jw00793001001_01108_00002_nis_uncal.fits

**4 dithers with same ngroup,nint for F380M**
jw00793001001_01109_00003_nis_uncal.fits
jw00793001001_0110a_00003_nis_uncal.fits
jw00793001001_0110b_00003_nis_uncal.fits
[...]

PR #350, opened 2019-08-13
Fix filenames with respect to sequence id and exposure number.

### LINKED ISSUE #349: File names of parallel exposures are wrong
The sequence id of parallel exposures has to be unequal to 1, but mirage currently sets it always to 1. 
The sequence id for the first parallel instrument should be 2.

Johannes-Sahlmann:
This PR provides a preliminary fix to the sequence id and exposure number encoded in the output file name.
It does not solve the issue of setting the activity id correctly.

Closes #349 

Relates to #281

### DIFF
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -302,15 +302,19 @@ class AptInput:
             if isparallel:
                 self.exposure_tab['sequence_id'][j] = '2'
 
-        # set exposure number (new sequence for every combination of seq id and act id and observation number)
-        temp_table = Table([self.exposure_tab['sequence_id'], self.exposure_tab['exposure'], self.exposure_tab['act_id'], self.exposure_tab['obs_num']], names=('sequence_id', 'exposure', 'act_id', 'obs_num'))
+        # set exposure number (new sequence for every combination of seq id and act id and observation number and detector)
+        temp_table = Table([self.exposure_tab['sequence_id'], self.exposure_tab['exposure'], self.exposure_tab['act_id'], self.exposure_tab['obs_num'], self.exposure_tab['detector']], names=('sequence_id', 'exposure', 'act_id', 'obs_num', 'detector'))
 
         for obs_num in np.unique(self.exposure_tab['obs_num']):
             for act_id in np.unique(temp_table['act_id']):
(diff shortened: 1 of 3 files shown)


Johannes-Sahlmann:
Thanks @bhilbert4 I think I solved the detector-related issue.

Johannes-Sahlmann:
This originates in a part of `yaml_generator` that I did not write nor modify, and I suspect it is caused by the change to the `activity_id` setting. Maybe we should have a separate issue that can be tackled when we address the activity id?

bhilbert4:
Sounds good to me. I haven't had any luck yet tracking down a description of how the activity ID is populated.

bhilbert4:
From my inquires, it sounds like activity IDs are generated by OPGS. They are present in Visit files but nowhere else that I've found. It seems like we'd need some functions that look at things like dithers/slews and probably other things in order to correctly calculate the activity ID for a particular exposure. 

So in the meantime, I guess the question is, is it better to set the activity ID to 01 for everything, or to go back to incrementing for each exposure, which is closer to correct, but not quite.

mperrin:
Activity IDs are part of the structure of the OSS visit files. They’re clearly labeled in those files, but may require some amount of reverse-engineering to recreate independent of that. For instance, slews and guide star acquisitions count as activities so those can increment the activity number without any exposure being taken. 

My recommendation would be let’s talk to Ilana Dashevsky to see if there is a good document summarizing how OPGS creates and assigns there. For each different APT template we may need to get a copy of a visit file created with that template and build some lookup table of what happens for that template...

mperrin:
(I wrote my comment just above at the same time as you were writing; sorry if it’s a bit duplicative.)

Visit files are not ITAR sensitive. They can be and are posted to the outer space wiki, and I don’t see any reason examples couldn’t be posted here too.

Johannes-Sahlmann:
I suggest that we keep the activity_id set to 1 until we figure it all out. 
JWST .vst files are not public and should therefore not be posted here (we had that disucssion for https://github.com/spacetelescope/jwst-visit-parser, too).

Johannes-Sahlmann:
So, should we merge this?

bhilbert4:
I think this is ready to merge, yes. Tackling the activity ID is going to be a whole separate effort.

PR #352, opened 2019-08-13
Source locations within subarrays -- redo

### LINKED ISSUE #264: Incorrect positions from dist. reffile for subarrays

bhilbert4:
This is a redo of #285, which is old enough that I'm having trouble rebasing to master. I didn't trust the long list of commits after rebasing, especially when the changes related to this PR are relatively minor. So this is a new copy of that PR. 

From the original:

This PR fixes a bug specific to subarray simulations.

In the case where sources were specified by RA, Dec values in the input catalogs and the distortion reference file (rather than pysiaf) was used to translate RA, Dec to x, y, the sources were being placed in the wrong location. This is because the distortion reference file translates to x, y in the full frame coordinate system. This was causing sources to be shifted from the subarray coordinate system.

In this fix, when the distortion reference file is used to calculate x, y values, the coordinates of the lower left corner of the subarray on the full detector are subtracted from the x, y values, in order to place them in the subarray coordinate system.

Closes #264
Also most likely solves #274


bhilbert4:
I've repeated testing. Pytest passes using the new `test_ra_dec_to_x_y.py` file. I also ran the catalog seed generator twice on each of three detectors: NIRISS, NRCA1, and NRCB1. For each detector and using the same point source catalog, I created a seed image first using the distortion reference file to calculate source locations, and then a second time using pysiaf to calculate the source locations. In all detectors, the sources were in the same positions in both cases, as expected.

I tested with these detectors in order to cover various values of DetSiYAngle and DetSciParity.

@Johannes-Sahlmann can you think of any other tests I should try? If not, I'm tempted to merge this fix since it is relatively simple.

Johannes-Sahlmann:
I think this looks fine and ready to merge.

PR #359, opened 2019-09-05
Niriss config

KevinVolkSTScI:
The code changes here are for allowing the user to put any allowed NIRISS filter name in the FILTER parameter of the .yaml file and put CLEAR in the PUPIL parameter in the .yaml file, and the code will track whether the filter of that name is actually in the filter wheel or whether it is in the pupil wheel.  The block of code for this needs to be put in both in catalog_seed_image.py and obs_generator.py as the .yaml file is read in twice rather than just once as I would have expected.  One could make a small function and have the code in one place.  That may be preferable to the current situation, but I would need some suggestion as to where the code should be put in that case.
I tested the code changes on my machine and they worked.  However, as seems to regularly be the case I ran into a git problem trying to push the changes back so I had to copy the blocks of code in the git editor.  Hence while the code changes should work I have not been able to test them from the git repository.

### DIFF
--- a/mirage/ramp_generator/obs_generator.py
+++ b/mirage/ramp_generator/obs_generator.py
@@ -2257,12 +2257,12 @@ class Observation():
         try:
             with open(self.paramfile, 'r') as infile:
                 self.params = yaml.safe_load(infile)
-                if self.params['Inst']['instrument'].lower() == 'niriss':
-                    newfilter,newpupil = utils.check_niriss_filter(self.params['Readout']['filter'],self.params['Readout']['pupil'])
-                    self.params['Readout']['filter'] = newfilter
-                    self.params['Readout']['pupil'] = newpupil
         except FileNotFoundError as e:
             print("WARNING: unable to open {}".format(self.paramfile))
+        if self.params['Inst']['instrument'].lower() == 'niriss':
+            newfilter,newpupil = utils.check_niriss_filter(self.params['Readout']['filter'],self.params['Readout']['pupil'])
+            self.params['Readout']['filter'] = newfilter
+            self.params['Readout']['pupil'] = newpupil
 
     def read_saturation_file(self):
         """Read in saturation map from fits file"""
--- a/mirage/ramp_generator/obs_generator.py
+++ b/mirage/ramp_generator/obs_generator.py
@@ -44,7 +44,7 @@ import pysiaf
 
 from mirage.ramp_generator import unlinearize
 from mirage.reference_files import crds_tools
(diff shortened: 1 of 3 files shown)


KevinVolkSTScI:
I have made changes here at the github level but need to test them.  I will try to do that tonight, if git allows me to pull down the revisions.  Since there were issues trying to post changes from my computer to github there is a good chance that it will give me problems trying to go the other way with the first set of changes still pending.

bhilbert4:
This won't quite work the way it is at the moment. Since check_niriss_filter is located outside of the Observation and Catalog_seed classes in obs_generator.py and catalog_seed_image.py, it will not have access to the self.param variable. So you'll have to set up check_niriss_filter such that you pass the filter and pupil names into it, and have it return the (possibly modified) filter and pupil values, which you can then assign to the proper self.param entry within obs_generator and catalog_seed_image.

mirage/ramp_generator/obs_generator.py
```diff
@@ -2257,6 +2257,10 @@ def read_parameter_file(self):
         try:
             with open(self.paramfile, 'r') as infile:
                 self.params = yaml.safe_load(infile)
+                if self.params['Inst']['instrument'].lower() == 'niriss':
```
bhilbert4: This looks better. But I think it would be better to move these new lines of code below the try/except block. This way any problems with check_niriss_filter won't cause a FileNotFoundError to be raised.

mirage/utils/utils.py
```diff
@@ -153,6 +153,47 @@ def calc_frame_time(instrument, aperture, xdim, ydim, amps):
...
+    pupil nothing is done, but the user can just put the filter name in the filter parameter and CLEAR in the pupil 
+    parameter, in which case this routine sorts out which value actually goes where.  Hence for example one can have a 
+    parameter file with FILTER = F277W and PUPIL = CLEAR or FILTER = F090W and PUPIL = CLEAR for imaging and either will 
+    produce the desired result although the actual pairings would be F277W/CLEARP and CLEAR/F090W respectively.
+    
+    The code also corrects CLEARP to CLEAR if needed.
+    
+    Parameters:
+    
+    filter:   a string variable, assumed to be the self.params['Readout']['filter'] value from the .yaml file
```
bhilbert4: Not super-important, but 'filter' is actually a built-in function in python. https://www.w3schools.com/python/ref_func_filter.asp
I try to avoid using 'filter' as a variable name, in favor of 'filter_name' or 'filter_value'. It does still work with 'filter', but it can be a little confusing to people reading the code.

KevinVolkSTScI:
The code passes my checks and should be OK to merge.

PR #361, opened 2019-09-13
Fix switched channel values in CRDS dictionary

bhilbert4:
This PR fixes a bug in the generation of the dictionary sent to CRDS when searching for the best reference files. Previously the SHORT and LONG channels were switched from what they should have been. This only affected searches for the distortion reference file.


bhilbert4:
Trivial fix. My test case is now working. I'm going to merge this now.

PR #362, opened 2019-09-16
Optional user inputs to yaml generator

bhilbert4:
This PR makes changes to the optional user inputs into yaml_generator. Specifically, it changes the inputs for all source catalog types to be consistent, and requires a dictionary with one catalog per APT target, rather than one catalog per observation, as it was previously. See the yaml_generator documentation page (mirage/docs/yaml_generator.rst) for details.

Other input changes include allowing the user to specify:

- Cosmic ray library and scale for each observation
- PAV3 for each observation
- Background rates for each instrument and observation
- The date of each observation

Along with the code changes, I have updated the yaml_generator documentation page as well as the relevant tests. I have also updated the imaging mode example notebook. These new options are shown in the notebook, and the notebook has been reorganized a bit into a more logical workflow. I plan to do the same to the WFSS and moving targets notebooks this week.

Does anyone have some time in the next couple days to test this? @NorPirzkal I know you were excited about specifying catalogs per target rather than per observation.

Closes #213

PR #363, opened 2019-09-16
Add flat field to list of reffiles returned by CRDS query

bhilbert4:
Small addition that adds flat field files to the list of reference files that CRDS is queried for. These files are then added to the input yaml file. Previously the flat field file was hardcoded to None in the yaml files.

### DIFF
--- a/docs/yaml_generator.rst
+++ b/docs/yaml_generator.rst
@@ -78,7 +78,7 @@ There are currently a number of parameters which can only be set using the ``par
 JWST Calibration Reference Files
 ++++++++++++++++++++++++++++++++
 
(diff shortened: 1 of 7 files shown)


bhilbert4:
This PR also fixes a bug in the reference file override dictionary where some values were not being pushed into the yaml files correctly due to mismatches in the CRDS key name and the Mirage key name for particular reference file types.

bhilbert4:
Tests are now passing here. I'm tempted to merge this without much of a review since it is just adding another type of reference file to the already-merged list of reference files for CRDS to look up and that the user can specify. Unless someone wants to take a look at this in the next couple days? Given the short timescale though (Mirage is set to go public on Monday!), I think #362 is more important for reviewing. @KevinVolkSTScI @Johannes-Sahlmann

bhilbert4:
Tests are all passing. Merging. I'll deal with any bugs in separate PRs.

PR #364, opened 2019-09-17
Fix bug in niriss filter check when grism is in the filter parameter

bhilbert4:
This is a simple fix to the niriss_filter_check function, for cases where the input filter name is not in either the filter wheel filter list nor the pupil wheel filter list. The most common example of this will be WFSS observations where the Grism will be in the input filter name (since the grisms are in the filter wheel).

With this fix, if the input filter is not in either of the wheel lists, the input filter and pupil names are returned unchanged.


bhilbert4:
@KevinVolkSTScI does this look ok to you?

KevinVolkSTScI:
Yes this change is fine.  I did not consider what would happen with someone putting the grism name in the FILTER parameter, but that is where the GR150 grisms are situated.  Thank you for finding this oversight on my part.

PR #365, opened 2019-09-18
Remove pysiaf pre delivery dir

bhilbert4:
This PR changes instances of pysiaf.Siaf for NIRCam to be the same as those for the other instruments. Previously Siaf instances for NIRCam were created by looking in pysiaf's pre-delivery directory, as updated coefficients were stored there. However, in order to be consistent with the current set of distortion reference files (which use the old coefficients), which removes the possibility of creating the data with one set of distortion coefficients and then running the pipeline with another set, we move NIRCam back to the old coefficients in pysiaf. Also, in the latest version of pysiaf, the pre-delivery directory is empty.

This has implications for level 3 pipeline processing of Mirage data. NIRCam mosaics will not look right because the old distortion coefficients have large round trip (x, y -> RA, Dec -> x, y) errors around the edges of the detectors.

### DIFF
--- a/mirage/utils/siaf_interface.py
+++ b/mirage/utils/siaf_interface.py
@@ -23,7 +23,6 @@ import numpy as np
 
 import pysiaf
 from pysiaf import iando
-#from pysiaf.constants import JWST_DELIVERY_DATA_ROOT
 from ..utils import rotations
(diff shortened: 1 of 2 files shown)


bhilbert4:
Tests are now passing. I removed NIRCam from the round trip test in test_ra_dec_to_x_y.py

PR #370, opened 2019-09-24
Rename self.pararms --> self.params

eslavich:
This is just a little thing I noticed while I was cruising through the code.  It looks like this was intended to replace self.params with the expanded paths, but because of the typo it is writing to a new attribute.


bhilbert4:
Good catch! Thanks!

PR #375, opened 2019-10-01
Fix installation

jotaylor:
The current `environment.yaml` no longer works now that STScI does not support the astroconda-dev channel. I removed that channel and subsequently moved the `jwst` installation from via conda to via pip, matching the version that was previously in the env file. I also added an explicit `crds` import since it isn't done by default with pip's installation of `jwst`.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -25,4 +25,4 @@ dependencies:
 - pip:
     - jwst-backgrounds>=1.1.1
     - pysiaf==0.3.0
-    - git+https://github.com/spacetelescope/jwst@9d5eb8c#egg=jwst
+    - git+https://github.com/spacetelescope/jwst@0.13.8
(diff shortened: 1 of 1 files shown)


bhilbert4:
Thanks Jo. I'm doing the same thing for one of my other repos today and figured that Mirage would come next. My one comment is that the jwst version previously in the environment file was a minimum version. It probably makes more sense to have Mirage use a stable version of the pipeline. Do you agree? 

I updated my other repo to use the 0.13.8 release using: `git+https://github.com/spacetelescope/jwst@0.13.8`

Maybe Mirage should go that route as well?

jotaylor:
Agree, good suggestion! I updated my branch to reflect this.

bhilbert4:
Looks good to me! Thanks! Hacktoberfest += 1 :)

PR #378, opened 2019-10-03
Modernization of seed image from fits (mosaic) image

### LINKED ISSUE #372: Update cropping and blotting code for new subarray def file
It looks like `fits_seed_image.py` was never updated to reflect the newer subarray definition file format, so this code is no longer working. 

Update and use pyisaf in place of what was previously in the subarray definition file.

bhilbert4:
This PR contains updated code for the seed image from fits functionality. That script had been left in a very primitive state, and Mirage has evolved quite a bit since then. These changes bring the code up to the current state of Mirage. This is in response to #372 

Given a distortionless image (e.g. a mosaic from CANDELS), this code will take a yaml file, crop an area of the mosaic roughly equal that to be covered by the new simulation, and then call the jwst package's resample method to resample the input onto the NIRCam/NIRISS pixel scale and introduce the proper distortion. This output can then be 1) used directly as a seed image, or 2) used as one entry in an extended source catalog when creating a seed image.

Closes #372


bhilbert4:
Just to complete the documentation on what the code does: between the cropping and resampling step, the code will create a matching PSF kernel (using photutils) which it will then convolve with the cropped image, in order to change the PSF from that of the input mosaic to the JWST instrument being simulated. 

Currently it will only do this convolution if the JWST PSF FWHM is larger than the FWHM of the input mosaic. This is to avoid sharpening the image during convolution. This is limiting, in that the JWST PSFs, especially at the shorter wavelengths, have quite small FWHM values, making it difficult to use data from other instruments for the mosaic. We may revisit this issue to try and provide more flexibility.

I have also created a new notebook in the examples directory showing two ways to use this code.

@swara13, I'm going to merge this code now so that you can play with the new notebook and see if it is helpful for the work you were doing recently. Feel free to send me comments or open github issues for any problems.

PR #380, opened 2019-10-07
Add grism files and temp nircam distortion files to downloads

bhilbert4:
This PR adds the grism-related reference files to the downloader script. These files were accidentally omitted in the previous version of the downloader, and are necessary for creating simulated WFSS data.

I have also added in the yet-to-be-released NIRCam distortion reference files. These files contain the correct distortion coefficients, whereas the SIAF is currently working with older, slightly-incorrect coefficients that do not round-trip (x, y -> RA, Dec -> x, y) correctly. This update currently does not change the yaml generator to look for these new files automatically. Users would have to manually override the distortion files by including them in an input dictionary when the yaml_generator is called. I think this is preferable, since the correct coefficients will make it into SIAF at some point, at which time the yaml generator would have to change again if it were to change now.

### DIFF
--- a/docs/reference_files.rst
+++ b/docs/reference_files.rst
@@ -14,13 +14,15 @@ After installing Mirage, these reference files can be downloaded using the *down
 
   from mirage.reference_files import downloader
   download_path = '/path/into/which/files/are/downlaoded/'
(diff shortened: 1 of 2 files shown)


bhilbert4:
Merging this now to see if it solves the help call received by @dthatte

PR #382, opened 2019-10-09
Update reference file versions in test yaml

eslavich:


eslavich:
(@bhilbert4 harvested the useful bits of this and the rest isn't needed, so I'm closing the PR)

PR #383, opened 2019-10-09
Fix non-Travis tests

bhilbert4:
Several tests that are not run on Travis were failing. This PR fixes those tests. 

- The reference files in the NIRISS example yaml files were set to 'crds' so that there is no need to update the reference file names each time a new reference file is delivered.

- The unnecessary catalog input was removed from test_filename_generation.

- In test_yaml_generator, the Mirage environment variable is set back to its original value if the test is not being run on Travis.


bhilbert4:
Travis tests are passing and the failing tests are now passing locally. Merging.

PR #384, opened 2019-10-09
Address concerns raised by flake8

eslavich:
@bhilbert4 I was running flake8 on my S3 branch and noticed that it found some issues in code from master.  This fixes most of them, there is just one that I don't know how to handle, which I'll point out in a comment on this PR.

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -15,6 +15,7 @@ import os
 import copy
 import re
 import shutil
+from yaml.scanner import ScannerError
 
 import math
 import yaml
@@ -3350,7 +3351,7 @@ class Catalog_seed():
         x_pos_ang = self.calc_x_position_angle(pixelv2, pixelv3, pos_angle)
         print('extended position angle:', x_pos_ang)
         print('scipy rotate seems to be failing with no error raised')
-        rotated = rotate(stamp_image, x_pos_angle, mode='nearest')
+        rotated = rotate(stamp_image, x_pos_ang, mode='nearest')
         print('rotated shape: ', rotated.shape)
         return rotated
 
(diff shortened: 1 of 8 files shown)


mirage/seed_image/fits_seed_image.py
```diff
@@ -375,6 +377,7 @@ def save_seed_image(self, model):
         dim = len(self.seed_image.shape)
         if dim == 2:
             units = 'e-/sec'
+            # TODO(eslavich): Where should this come from?
             yd,xd = arrayshape
```
eslavich: This is the line that I don't know for sure how to fix.  Should I just replace `arrayshape` with `dim`?

bhilbert4: I wouldn't worry about fits_seed_image.py, nor blot_image.py. I'm in the middle of basically completely rewriting those. I put them together for a particular case for someone a long while ago and haven't touched them since, so they are in serious need of updating.

mirage/seed_image/catalog_seed_image.py
```diff
@@ -3350,7 +3351,7 @@ def rotate_extended_image(self, stamp_image, pos_angle, right_ascention, declina
         x_pos_ang = self.calc_x_position_angle(pixelv2, pixelv3, pos_angle)
         print('extended position angle:', x_pos_ang)
         print('scipy rotate seems to be failing with no error raised')
-        rotated = rotate(stamp_image, x_pos_angle, mode='nearest')
```
bhilbert4: Huh. I wonder if this is why I was having problems with `rotate`...

bhilbert4:
These changes all look good, thanks! I should have a completely new version of fits_seed_image.py and blot_image.py next week sometime.

mirage/seed_image/fits_seed_image.py
eslavich: Okey dokey, I'll just remove my comment

PR #386, opened 2019-10-11
Add reference pixel masking

### LINKED ISSUE #385: Reference pixels have source signal in dispersed seed images
It looks like in the case of WFSS observations that the dispersed seed image has signal from sources+background in the reference pixels. Direct seed images use a mask to keep the reference pixels at zero in the seed image. We need to apply a similar mask to the dispsersed seed image when it comes back from the disperser. 

This will probably require a call to pysiaf to get aperture corners and dimensions, since some subarrays have reference pixels on one or more sides, while others do not. 

This bug was found because the WFSS background for F356W falls off dramatically right around the amp 3/4 boundary. This was causing a large difference in the mean signal in the reference pixels between amps 3 and 4, which was in turn causing a large difference in the mean signal after reference pixel subtraction.

bhilbert4:
This PR fixes an issue where reference pixels in the seed images contained signal from astronomical sources. A mask has been created that zeros out the reference pixels in the seed image when in modes other than WFSS. A similar mask is applied to dispersed seed images coming out of the disperser.  Direct seed images going into the disperser are not masked because the disperser needs to know about the astronomical signal in areas outside the science pixels.

I've tested this on direct seed images in full frame and subarray mode, and the seed images now look correct. I still need to test WFSS observations.

Closes #385

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -257,11 +257,33 @@ class Catalog_seed():
            (aperture_suffix not in ['FULL', 'CEN'])):
             self.seedimage, self.seed_segmap = self.pad_wfss_subarray(self.seedimage, self.seed_segmap)
 
-        # Save the combined static + moving targets ramp
-        self.saveSeedImage()
-
+        # For NIRISS POM data, extract the central 2048x2048
         if self.params['Inst']['mode'] in ["pom"]:
             self.seedimage, self.seed_segmap = self.extract_full_from_pom(self.seedimage, self.seed_segmap)
+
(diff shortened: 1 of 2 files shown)


bhilbert4:
Tests on a full frame WFSS exposure look good as well. The seed image has reference pixels zeroed out, and the rate image that comes out of running the detector1 pipeline no longer has the amp-dependent mean signal level that @NorPirzkal first saw.

bhilbert4:
Test on a NIRISS WFSS exposure also looks good. I'm going to merge this so that @NorPirzkal can try re-running his simulations where he first noticed the problem.

PR #388, opened 2019-10-18
Bug for sources just outside the aperture

bhilbert4:
This PR fixes a bug in the catalog seed generator when there is a source located just outside the fov of the aperture. Sources very far from the aperture were (and continue to be) successfully filtered from the source list. Also, sources landing partially within the aperture were being treated correctly. But sources completely outside the aperture by only a small distance were not being treated correctly. The code was generating `None`'s for these sources' x, y locations, but then these `None`'s were not being treated correctly later in the code. These fixes ignore sources with coordinates of `None`.

I have tested this fix using a yaml file and source catalog provided by @NorPirzkal, using which he first noticed the error. I also created WFSS and imaging test cases, stepping sources over the edges of the aperture, all of which were treated appropriately.


bhilbert4:
Merging this so that @NorPirzkal can try re-running his simulations.

PR #399, opened 2019-11-05
WFSS background scaling

### LINKED ISSUE #367: Background level not scaled properly in WFSS dispersed images
I created different simulated images running `wfss_simulator.WFSSSim` and specifying different levels of background in the yaml files, but the outputs are all the same. So I went back to the code and found that in mirage/wfss_simulator.py the background is included at `disp_seed.finalize(Back=background_file)`. This function lives in NIRCAM_Gsim/grism_seed_disperser.py and could accept a `BackLevel` argument, which is however not specified in mirage/wfss_simulator.py. Hence, the images all have the same background, while they should reflect what is specified in the yaml file.

### LINKED ISSUE #376: Background in seed images to be dispersed should be zero
Given that the background in dispersed seed images is added by the disperser software (see #367 ) the direct images that are fed into the disperser should have no background signal at all. Currently Mirage does not force this to be true.

We should add something to wfss_simulator to make sure the background is zero. Keep in mind though that a background level for the dispersed seed image is still needs to be an input and most likely non-zero. 

So it seems like the best solution is to temporarily update the yaml files before creating the direct seed images to have a background level of 0, and then return to the original yaml file for determining the background in the dispersed image.

### LINKED ISSUE #377: WFSS simulator should create single direct seed from wfss yaml file
In the case where a user runs wfss_simulator and provides multiple yaml files, if create_continuum_sed is True, Mirage will currently create a seed image from the first input yaml file only and then move on.

But in cases where the observation details (e.g. background) are different from yaml to yaml, using the first yaml from the list may be incorrect. Better in this case would be to create the direct seed image from the WFSS mode yaml file.

This should be an easy fix. The wfss_simulator already knows the name of the WFSS mode yaml file and has it saved in self.wfss_yaml.

Just make sure the Mirage does the right thing in producing the seed image from a yaml file with mode set to WFSS. Also keep in mind #376. The background for this file should be set to zero temporarily while creating the direct seed image. The dispersed background will be created and added following the method described in #367

bhilbert4:
This PR makes changes to the background signal in WFSS observations. 

- For NIRISS, the existing background files are scaled according to the user-input "low", "medium", "high" levels. Currently no scaling is done.

- For NIRCam, a 2D background image is computed on the fly using `jwst_backgrounds`. Users can specify an observation date, or use "low", "medium", "high". 

- The order of the yaml files in `wfss_simulator` has also been updated such that the WFSS mode file will always be first. This ensures that the direct seed image will be created from that file.

Resolves #367 
Resolves #376 
Resolves #377


bhilbert4:
@KevinVolkSTScI with this update to the WFSS background scaling, NIRISS WFSS observations will always have a background signal equal to the appropriate pre-computed background image, scaled by low/medium/high. Should we allow users to enter a number for the background rate, in which case a constant background of <number> DN/sec/pixel will be added? It doesn't make a whole lot of sense to me to allow that option, with the possible exception of zero, if there is some case where someone wants to create data with no background added.

KevinVolkSTScI:
I would prefer to allow the user to enter a number and scale to that, the same as for imaging backgrounds.  One may want either a low or a high background compared to the standard models but not want to provide an image for the background (i.e. say observing near the Orion Nebula...).

bhilbert4:
Ok, I can add a block for that. To summarize then, the background options for WFSS are:

NIRISS:
bkgdrate = a number
The number is treated as the background value in ADU/sec/pixel to add to the seed image. All pixels get the same value. (Identical to the imaging case)

bkgdrate = low/medium/high
The pre-computed background image is read in, scaled to the requested level, and added to the seed image.

NIRCam:
bkgdrate = a number
Same as NIRISS case above

bkgdrate = low/medium/high
The spectrum corresponding to the input level is chosen from jwst_backgrounds and fed into the disperser, which creates a 2D dispersed background image on the fly

'simSignals': 'use_dateobs_for_background' = True
The spectrum corresponding to the observation date is retrieved from jwst_backgrounds and fed into the disperser, which creates a 2D dispersed background image on the fly

Does that sound reasonable for NIRISS?

KevinVolkSTScI:
When I said one may not want to use an image for the background I meant an image for the undispersed background.  Given an input number for the spectral background ADu/s/pixel the WFSS background image should be scaled to that number, the same way LOW/MEDIUM/HIGH are.

bhilbert4:
For WFSS observations, the background is added by the disperser, so the undispersed seed image, which is created and saved prior to calling the disperser, will always have zero background. For the dispersed seed image, you are saying that the pre-computed background image should be scaled such that the mean (median? max?) of the background image is equal to the input number?

KevinVolkSTScI:
Above you say that in the case of bckdrate bring "low", "medium" or "high" that the template image is read in and scaled to the signal level.  The template image is from CV3 observations where the continuum source illuminated the grisms (the simulated image from the disperser is a bit different than the observed CV3 image, probably because the position of the imaging field of view in the POM is a little different than assumed in the disperser calculations).  Given an input count rate for the background the same type of scaling should be done.  I have to check how the template images were normalized.  Originally they were normalized to 1 in the area where the full signal was received, and you could just scale by the count rate.  I think Nor at some point wanted the images scaled to "medium" signal, but I am not 100% sure of what was done in the end.

bhilbert4:
Ok, makes sense.

You told me a week or two ago that the NIRISS background files are scaled to "medium", so that is what I assumed when I wrote the code. Let me know if that's not true. With a low/medium/high entry, jwst_backgrounds is used and returns a number. I then ratio that number to the number jwst_backgrounds gives for "medium", and I multiply the background image by that ratio.

So in the case where the user provides a number, I can do the same thing. I will ratio the number to the number jwst_backgrounds gives for medium, and then mulitply the background image by that ratio.

KevinVolkSTScI:
I will re-check but they are likely scaled to medium as per Nor's request.

bhilbert4:
The names of the NIRISS background files are *medium_background.fits.  So I'm assuming that is what they are scaled to.

One potential confusion with allowing a number to be input comes from the fact that the "medium" background value depends on the pointing. 
My test case uses F090W and a random pointing of RA=12.2 Dec=12.2. jwst_backgrounds says that "medium" for this pointing is about 0.5 ADU/sec/pixel. But the values in the background file are ~1.1 - 1.2. So when I enter 10 as my requested background value, the ratio of 10./0.5 = 20. The background image is then multiplied by 20, so that pixels have background signals of ~22 - 24 ADU/sec/pixel. This is different than the input value of 10 and could make users confused.

I have added  the couple lines necessary to allow for numerical values for the NIRISS background, and I've tested all the options for NIRISS. So if you're happy with this then it is ready to be looked over (if you want to), or merged. Let me know what you think.

KevinVolkSTScI:
The values I used to normalize the WFSS backgrounds are in electrons/second and these are from the old ETC low/medium/high spectra:

# Calculated background count rates in electron/second for WFSS
# total rates per pixel for orders -1 to 3
#
# Filter  JWST standard   ETC low    ETC medium    ETC high     Scattering
F090W     5.356419e-01 6.598314e-01 1.169831e+00 2.184065e+00  4.376572e-01
F115W     5.243016e-01 7.120550e-01 1.262420e+00 2.356930e+00  4.397160e-01
F140M     1.978292e-01 3.097863e-01 5.492275e-01 1.025403e+00  1.857715e-01
F150W     3.991473e-01 6.500614e-01 1.152510e+00 2.151730e+00  3.921011e-01
F158M     1.989966e-01 3.368594e-01 5.972244e-01 1.115005e+00  2.061193e-01
F200W     3.229331e-01 5.543794e-01 9.828732e-01 1.835018e+00  4.155141e-01

Hence the discrepancy is not as bad as it looked.  The images need to be divided by the above ETC medium values and then multiplied by the input background in ADU/second.

bhilbert4:
Looking at the F090W background file, the sigma-clipped mean (1.1695229) matches up nicely (0.00027%) with the ETC medium value above. So would you say it's safe to simply normalize the background image by it's sigma-clipped mean value and then scale? I just hesitate a bit to hard-code the numbers above, just in case the background files change in the future and new values are needed.

KevinVolkSTScI:
The relation between the background value and the mean/median/mode of the image depends on where the different orders are with respect to the direct image.  For F090W the first and second order spectra are close to the direct image position and one therefore does not see much variation in signal across the image when different orders come and go.  For other filters this is not the case and you will have problems if you try to use a statistic over the whole frame.  It may work OK if you do statistics on say the central 200x200 pixels for the scaling.  Try looking at the F200W image and see if this works OK.  If it is good there, it should work for the other filters where the spectral offsets are smaller.

bhilbert4:
With F200W, I still get sigma-clipped means (in both R and C files) within 0.00025% of the ETC value when I work on the entire image. When I cut down to the central 200x200 pixels, the sigma-clipped mean goes up to the point where it differs by ~2%

KevinVolkSTScI:
I just did a clipped mean on the images (in IRAF, but that should not affect things too much) and the values that came back were all close to the "target" values so it seems this will be fine.

bhilbert4:
Excellent! I'm waiting for the tests to finish running. Once they are all passing, are you ok with this being merged?

KevinVolkSTScI:
I was not able to look at all the code changes, but the ones related to the NIRSS scaling seem to be OK.
This week when I installed the latest version of Mirage to do the NIRISS WFSS simulations I saw that the niriss_zeropoints.list has not been updated with the change of the F_lambda values to CGS units.  I must have forgotten to file a change request for the file.

bhilbert4:
My tests all look good. Travis tests are passing. I'm going to merge this.

This PR allows WFSS background entries of:

NIRISS:
   number: background image is scaled such that the mean value is this number
   low/medium/high: background image is scaled to match the requested level using background values from jwst_backgrounds

NIRCam:
   low/medium/high: Spectrum corresponding to the requested level is obtained from jwst_backgrounds. Disperser creates a 2d dispersed image from this spectrum
   observation date: Spectrum corresponding to the requested observation date is obtained from jwst_backgrounds. Disperser creates a 2d dispersed image from this spectrum

PR #400, opened 2019-11-05
Change units for F_lambda in Guider/NIRISS zeropoints files

KevinVolkSTScI:
I have changed the units of the F_lambda entries in guider_zeropoints.list and niriss_zeropoints.list from W/m^2/miceron to erg/s/cm^2/Angstrom.  All the values are divided by 10 in the new versions  compared to the old versions of the files.


KevinVolkSTScI:
Hi Bryan,

  OK I changed the entry in the test file.  It looks like it is now in the change request, but if not I can issue another one.

Kevin

bhilbert4:
Looks good to me. I'll let the tests run, and assuming they pass, this will be ready to merge.

bhilbert4:
Oops. i forgot to look farther down in the test file. The FGS value on line 36 also needs to be updated.

KevinVolkSTScI:
OK I made that change as well.

Kevin

PR #402, opened 2019-11-07
Overlapping wfss source types

### LINKED ISSUE #401: Problem when dispersing overlapping sources
I wanted to create a very crowded scene with 13 sources.
Screen Shot 2019-11-06 at 3 12 02 PM
The seed image looks perfectly fine, but when I try to use the disperser, it crashes with an error saying that source 15 is not in the list, while in principle the code should only count up to 13 sources.

The crash happens in `obervations.py`, where the code identifies the number of sources doing a list of the set of the ravel of the segmentation map: `all_IDs = np.array(list(set(np.ravel(self.seg))))`.
Screen Shot 2019-11-06 at 3 10 00 PM
Some sources are added together when the stars and galaxies segmentation maps are added and the counts go up to 16 missing 4 and 14.

I see two problems here:
- Summing up the stars and galaxies segmentation maps is not probably the best idea when having overlapping sources;
- Also when galaxies overlap, there does not seem to be a way to keep track of which sources contribute to a specific pixel. The segmentation map seems to keep only a single number in overlapping regions.

bhilbert4:
This PR tweaks the way that the WFSS simulator code works, in order to calculate the correct dispersed signal in areas of the detector where sources **with different source types (e.g. a star and a galaxy)** overlap.  Previously the direct seed image and segmentation map were fed into the disperser. However, in cases where objects of different types overlapped, the segmentation map was being improperly calculated. In that case the object numbers from the two objects were summed. This was later interpreted as a new object number by the disperser. 

With this fix, we now save the seed image and segmentation map from each type of object separately. Each of these is then sent to the disperser, and the resulting dispersed seed images are then combined. In this way, the signal from a star and a galaxy that fall on the same pixel will be accurately computed.

This fix does *NOT* address the problem of overlapping sources of the same type (e.g. two overlapping galaxies). In that case, the segmentation map keeps the value from only one of the two objects, meaning that only one object's signal will fall on those pixels. A fix for this will come later, as it requires modifications to the disperser software.

Resolves #401

### DIFF
--- a/mirage/psf/psf_selection.py
+++ b/mirage/psf/psf_selection.py
@@ -488,7 +488,7 @@ def _load_itm_library(library_file):
         # Convert to HDUList and create library
         phdu = fits.PrimaryHDU(data, hdr)
         hdulist = fits.HDUList(phdu)
-        library = to_griddedpsfmodel(hdulist, ext=0)
+        library = to_griddedpsfmodel(hdulist)
 
         return library
     else:
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -468,7 +468,21 @@ class Catalog_seed():
         return padded_seed, padded_seg
 
     def saveSeedImage(self, seed_image, segmentation_map, seed_file_name):
-        # Create the grism direct image or ramp to be saved
+        """Save the seed image and accompanying segmentation map to a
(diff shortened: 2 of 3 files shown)


bhilbert4:
This fix ran successfully in my tests as well as @camipacifici 's notebook, which illustrated the original problem. Merging this now.

PR #404, opened 2019-11-08
Extended source rotations

bhilbert4:
This PR turns the rotation for extended sources back on. Previously, there were cases where it seemed that scipy.rotate was silently failing. I didn't see any failures in my tests here.

- I turned the rotations back on

- Removed the relevant block of code from the (too large) try/except block

- Adjusted the rotation angle. This is because calc_x_position_angle() was designed around the galaxy inputs. However, the definition of position angle is different between galaxies and extended sources. For galaxies it describes the position angle that the galaxy will have in the seed image, while for extended sources it is the position angle of the source within the provided stamp image.

 My limited set of test cases seemed to work fine.


bhilbert4:
Merging this and we can deal with any bugs down the road.

PR #407, opened 2019-11-19
Frame time corrections

bhilbert4:
This PR makes two small adjustments to the calculation of the frame time.

1. Correct the order of x,y inputs to calc_frame_time() in catalog_seed_image(). Previously x and y were in the wrong order. This switch was only noticed when creating data using the grism stripe 256 aperture, which is 2048 x 256. For square apertures, this correction does not change the calculated frame time.

2. Update FGS timing model. FGS frame times were updated according to the model provided by Stansberry. This includes changing the padding for subarrays 32x32 and smaller.

### DIFF
--- a/mirage/ramp_generator/obs_generator.py
+++ b/mirage/ramp_generator/obs_generator.py
@@ -870,7 +870,7 @@ class Observation():
                 f_match = self.params['Readout']['filter'] == fw_positions['Name']
                 p_match = self.params['Readout']['pupil'] == fw_positions['Name']
             elif self.instrument.upper() == 'NIRCAM':
-                if self.detector[-1] == '5':
+                if '5' in self.detector[-1] or 'LONG' in self.detector.upper():
                     channel = 'LW'
                 else:
                     channel = 'SW'
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -149,7 +149,7 @@ class Catalog_seed():
         # calculate the exposure time of a single frame, based on the size of the subarray
         self.frametime = utils.calc_frame_time(self.params['Inst']['instrument'],
                                                self.params['Readout']['array_name'],
-                                               self.nominal_dims[0], self.nominal_dims[1],
+                                               self.nominal_dims[1], self.nominal_dims[0],
                                                self.params['Readout']['namp'])
         print("Frametime is {}".format(self.frametime))
 
--- a/mirage/utils/utils.py
+++ b/mirage/utils/utils.py
@@ -141,19 +141,23 @@ def calc_frame_time(instrument, aperture, xdim, ydim, amps):
     elif instrument == 'fgs':
         xs = ydim
         ys = xdim
-        colpad = 6
-        if 'acq1' in aperture.lower():
-            colpad = 12
-        rowpad = 1
-        if amps == 4:
+        colpad = 12
+
+        if namps == 4:
+            rowpad = 1
             fullpad = 1
         else:
+            rowpad = 2
             fullpad = 0
+
+        if ((xdim <= 32) & (ydim <= 32)):
+            colpad = 6
+            rowpad = 1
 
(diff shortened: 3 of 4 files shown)


bhilbert4:
@KevinVolkSTScI does this timing pattern change for FGS agree with your expectations?

KevinVolkSTScI:
According to the Guider pages (https://jwst-docs.stsci.edu/observatory-hardware/fine-guidance-sensor#FineGuidanceSensor-Track), COLPAD is 6 for read-out modes other than 128x128 pixel ACQ1.  hence the full frame read time is given as 10.6138 seconds.  For NIRISS, and also for NIRcam as far as I know, COLPAD is 12 and the full frame read time is 10.6377 seconds.  The code is not giving the correct Guider frame times.  It needs to set COLPAD of 6 for the Guider in all cases but xdim = 128 and ydim = 128.  Right now it sets COLPAD = 6 only when xdim and ydim are 32 or smaller.  That is probably correct for NIRISS but we have no such sub-arrrays so it does not affect anything.  The Guider full frame read-out used to have COLPAD = 12 and was the same as NIRISS but that got changed at some point.

KevinVolkSTScI:
Another minor point, fullpad is zero for the Guider it seems.  The code has fullpad = 1 in that case.

bhilbert4:
Ok, I'll have to check with Paul G. John's IDL code has a note that updates were applied based on information from Paul, but this seems to disagree with what's in JDox.

KevinVolkSTScI:
I think what you coded is what the Guider frame times used to be, so maybe the last round of changes to the microcode were not communicated to everyone.  Pierre Chayer should know for certain what the current status is of the timing.

bhilbert4:
@KevinVolkSTScI, given the conversation with Pierre today, do you agree that these changes are correct for FGS?

KevinVolkSTScI:
Pierre said that the JDox description is correct.  That does not seem to correspond to what is coded unless I have misunderstood the code change page.  Do you reproduce the full frame value of 10.613820 seconds?  I think for full frame your code has colpad = 12 and that would give a frame time of 10.74 seconds or so, about 1% longer.

bhilbert4:
Ok, made a few updates, under the assumption that JDox is indeed correct for the full frame value. The function now calculates frame times that match what is in Pierre's spreadsheet. I also added some test functions in a new file, in order to better keep an eye on this.

bhilbert4:
Pierre said that colpad is 12 for subarrays larger than 64x64, so I believe the current code should work for all FGS subarrays regardless of whether they are CAL, ACQ1, etc.

Looks like the updated code caused a test to fail. I'll fix that in a new PR. Also, I also accidentally removed some lines I had in the code that swapped the x and y dimensions for NIRISS and FGS since their fast readout directions are opposite that of NIRCam (according to the note I left in the code). This was done so that the same equation could be used for all instruments.  It does make sense to have that in there, right? It seems like FGS doesn't have any non-square apertures, so it doesn't really matter there. Are NIRISS's only non-square apertures the STRIPE apertures for SOSS? In that case the axes flip shouldn't matter for NIRISS either since Mirage doesn't make SOSS data.

KevinVolkSTScI:
There are a number of non-science calibration arrays that are strips, mostly for WFSS mode.  We need to be able to use those in Mirage.  We may need to simulate SOSS mode subarrays too from a scene image using AWESIMSOS to get the GR700XD spectrum so that also should be supported.  In general since the sub-arrays are in the SIAF file users can select them and we want the correct frame times.  One way to be explicit is to have variables like nfast and nslow, and assign these to x or y at the start of the subroutine depending on the instrument.  But swapping the input (x,y) values when needed is OK as well.

The way I read the Guider JDox pages COLPAD is 6 for all sub-arrays except full frame and 128x128 including the strip modes used in acquisition.  What is in the code now is fine for the subarrays that you are supporting.  If Mirage ever supports the acquisition mode then this would have to be clarified.

PR #410, opened 2019-11-19
Get correct detectors for given aperture from yaml_generator

bhilbert4:
This PR includes changes to have the yaml_generator create files for the correct detectors for all apertures. Currently it blindly assumes that all detectors for a given module are used. But in some cases (e.g. SUB400P, SUBGRISM256), only a subset of the module's detectors are used.

Very early in the processes at the moment. Creating a PR so that I don't forget to work on this after getting some higher priority stuff out of the way.


bhilbert4:
Tests are passing, and my example cases using the (non-WFSC) apertures are giving good results. It's working well for the SUBXXXP, SUBGRISM, and DHSPIL apertures, as well as the standard SUBXXX and FULL apertures.

Coronagraphy apertures have been left in the NIRCam subarray definition file, but these are not currently used by Mirage. The other type of aperture that is not yet correctly defined in there are the target acquisition apertures. Mirage does not yet support TA observations. This should be fixed, at least for the grism time series TA apertures, in #298 

In the mean time, I'm going to merge this so that I can then bring the updates into the TSO branch.

bhilbert4:
Travis Linux tests all pass. OSX tests seem to be hanging. Going to merge since everything passed before updating the CHANGES.rst file.

PR #411, opened 2019-11-20
Re-tweaked FGS timing params after more discussion

bhilbert4:
This PR makes some additional adjustments to the FGS timing model after further discussions with the FGS team.

- colpad set to 12 for all but the apertures smaller than 64x64

- Added back in the x/y switch for NIRISS and FGS, which I mistakenly removed in #407 

- Added a few more frame time calculation tests, for non-square apertures

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -21,8 +21,8 @@ dependencies:
 - scipy>=1.1.0
 - sphinx>=2.1
 - webbpsf>=0.8.0
-- yaml>=0.1.7
+- yaml==0.1.7
 - pip:
     - jwst-backgrounds>=1.1.1
-    - pysiaf==0.3.0
+    - pysiaf>=0.6.1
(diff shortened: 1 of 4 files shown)


bhilbert4:
One of the test_deployment tests was failing on Travis due to the combination of yaml 0.2.2 and pyyaml 5.1.2. I found a resolved issue for this on pyyaml, but given that it's resolved I don't understand why this is a problem, nor why it just started showing up now. 
https://github.com/yaml/pyyaml/issues/320

Either way, I pinned the yaml version at 0.1.7 for the moment, in which case the test passes.

bhilbert4:
@KevinVolkSTScI any objections to merging this now? The function (which Paul G modified to fit his understanding of the parameters) returns frame times that match those in the PRD subarrays file that John Stansberry pointed out to me yesterday.

Issue #408, opened 2019-11-19
Improper scaling of input spectra


bhilbert4:
When a normalized spectrum is input for an object and Mirage is asked to scale the spectrum to a given magnitude in a given filter, the current way that the scaling is done is not optimal. At the moment, Mirage essentially assumes that the input spectrum has been normalized to 1.0 at the pivot wavelength of the filter in question. It then takes the requested magnitude, converts it to units of f_lambda, and multiplies the spectrum by that value. 

If the spectrum happens to be normalized in some other way (such as normalized to 1.0 at the peak, wherever that happens to be), then the rescaling will be incorrect.

Mirage should be updated such that during renormalizing, Mirage divides the input spectrum by its value at the pivot wavelength, and then multiplies by the f_lambda value calculated from the input magnitude. This way, the absolute values of the input spectrum won't matter. 

This raises the question then of how the user should notify Mirage that the spectrum should be rescaled. Currently Mirage rescales any spectrum that has units of 'normalized' in the hdf5 file. With this proposed change, any spectrum, even those with real flux density units, could potentially be rescaled. Perhaps keeping the 'normalized' units is best, since the user will have to explicitly set this.

bhilbert4:
Input spectrum rescaling is now done via synphot, with its renormalize() function.

bhilbert4:
This was corrected in #412. Closing.

PR #412, opened 2019-11-20
Correct scaling of input spectra

bhilbert4:
This PR changes the way that input spectra (from an hdf5 file or a dictionary) are renormalized if requested by the user. Previously Mirage was multiplying by the flux density at the pivot wavelength of the filter being used. This was incorrect.

This changes the renormalization to use synphot. The input spectrum is read in, along with the throughput curve of the filter. Synphot's renormalize function is then used. This works in units of photon/s/cm2/𝐴˚

Basically, synphot normalizes by:
new_spectrum = input_spectrum * renorm_value * (stdflux / total_flux)
where total_flux is the integration of the input_spectrum times the filter_bandpass, and
stdflux is the integration of either Vega or a constant spectrum times the filter_bandpass

By working in units of photons, the input spectrum is correctly normalized regardless of whether it contains emission/absorption lines in the filter bandpass.

In addition, this new strategy means that users can provide an input spectrum in any units, and the rescaling will be done correctly. Previously Mirage assumed that the input spectrum was normalized to a maximum value of 1.0. This is no longer true. As long as the user specified 'normalized' as the units of the input spectrum, it will be correctly scaled to the requested magnitude.

Perhaps I should change 'normalized' to something else, although that would break backward compatibility. 

Note that as of this initial commit, the renormalization process is functional (but not tested for correctness) for rescaling to vegamag values. It has not been tested at all for rescaling to AB mag nor ST mag values. It might also be worth adding an option for an input spectrum in counts in the case of an FGS observation. This is looking forward to a time when users can input spectra for imaging simulations, which is not yet possible.


bhilbert4:
Resolves #408

bhilbert4:
@KevinVolkSTScI I think this branch is ready to merge. Input spectra rescaled to magnitude X in photmetric system Y are giving the same count rate as that calculated by magnitude_to_countrate() for magnitude X and photometric system Y.

I tweaked the renormalization for NIRISS Vegamag cases to account for the assumption that Vega has a magnitude of 0.02. I have decided to push off the work on FGS/Sirius for now, after my quick attempt wasn't working. Currently input spectra are only allowed for WFSS mode, so FGS isn't relevant. At some point I'd like to allow for input spectra to be used for imaging mode sources. Even in that case I don't think it makes sense to worry about rescaling input spectra, so it may be that we don't need to worry about supporting FGS in these functions.

PR #415, opened 2019-11-27
2MASS catalog to vegamag. FGS seed image filenames exclude filter

### LINKED ISSUE #413: Incorrect photometric system for 2MASS search results
Line 135 in get_catalogs.py sets the photometric system to be AB mags for the query results from 2MASS. This should actually be set to 'vegamag'.

### LINKED ISSUE #414: Seed image file name for FGS exposures
from @obi-wan76 
In catalog_seed_image:

# Set FGS filter to "N/A" in the output file
# as this is the value DMS looks for.
if self.params['Readout'][self.usefilt] == "NA":
    self.params['Readout'][self.usefilt] = "N/A"

This is giving an error  when writing the fits files because the change happens after the ptsrc_seed_image.fits is created so basically:
First:
Point source image and segmap saved as /ifs/jwst/tel/wfr4_mirage_sims/psf_check/OTE-28_2MASS_WISE/jw01165001001_01101_00004_g1_uncal_NA_ptsrc_seed_image.fits
Then, change the filter to "N/A" and when trying to look for the seed fits:
FileNotFoundError: [Errno 2] No such file or directory: '/ifs/jwst/tel/wfr4_mirage_sims/psf_check/OTE-28_2MASS_WISE/jw01165001001_01101_00004_g1_uncal_N/A_seed_image.fits'

Seems like the best solution is to drop the reference to the filter in the filename when creating FGS data.

bhilbert4:
This PR fixes two small bugs:

- In get_sw_catalog(), which is the 2MASS catalog search in get_catalog.py, the photometric system in the output source catalog is changed from abmag to vegamag.

- In catalog_seed_image, when saving seed images for FGS simulations, the filter name is excluded from the seed image file name. Since FGS has no filters, it doesn't make sense to have "NA" in every filename. In addition, Mirage changes the filter value to be "N/A", to agree with what CRDS expects, but this was then causing Mirage to look for a filename with "N/A" in it, which was a problem. With this change, the filter substring is removed completely.

Resolves #413 
Resolves #414

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,17 @@
+Not Yet Tagged
+==============
+
+Catalogs
+--------
+
+- Change photometric system in catalog output from 2MASS query from ABmag to Vegamag (#415)
+
+Seed Image
+----------
+
+- Remove filter substring from seed image output file name in the case of FGS simulations (#415)
+
+
 1.1.1
 =====
 
--- a/mirage/catalogs/get_catalog.py
+++ b/mirage/catalogs/get_catalog.py
@@ -134,7 +134,7 @@ def get_sw_catalog(target_coords, search_radius=15 * u.arcmin):
         # Save shortwave catalog and header comments to file
         comments = [
             '',
-            'abmag',
+            'vegamag',
             'Catalog for ramp_simulator.py. created from 2MASS catalog',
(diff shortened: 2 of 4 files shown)


mperrin:
👍 Thanks Bryan!

PR #418, opened 2019-12-19
Fix number of dithers for nrc parallel to nis wfss case

### LINKED ISSUE #417: NUMDTHPT invalid for NIRISS WFSS+Nircam Imaging
For a NIRISS WFSS observation with NIRCAm imaging in parallel, it looks like the total number of dither points is not being saved correctly. The jwst datamodel issues a warning when saving the data, saying that 2-POINT-LARGE-NIRCam is not a valid entry for NUMDTHPT. This error goes back to the APT reader I believe, which should be stripping the '2' out of that name and saving that in the total_primary_dither_positions entry.

bhilbert4:
This PR makes a small change to the data read in from an APT observation where NIRCam imaging is used in parallel to a NIRISS WFSS prime observation. In this case, APT saves the name of the dither pattern in the PrimaryDither xml tag (other modes place the number of dither positions here). This change checks to see if the value is the name of the dither pattern and if so, extracts the number of dithers from it.

Note that in this case, the dither pattern names are N-POINT-<SIZE>-NIRCam, where N is a number and SIZE is SMALL, MEDIUM, or LARGE. This code update places that string in the primary dither pattern name, however the JWST datamodels currently do not recognize this as a valid value for the primary dither pattern. A JIRA ticket has been filed about that.

Resolves #417


bhilbert4:
Simple fix. Tests passing. Merging now.

PR #424, opened 2020-01-07
Switch versioning from relic to setuptools_scm

bhilbert4:
This PR switches over to use setuptools-scm for versioning. Prior to this, mirage was using relic for versioning.

### DIFF
--- a/mirage/dark/dark_prep.py
+++ b/mirage/dark/dark_prep.py
@@ -39,7 +39,6 @@ from mirage.utils import read_fits, utils, siaf_interface
 from mirage.utils.file_splitting import find_file_splits
 from mirage.reference_files import crds_tools
 
-MIRAGE_VERSION = mirage.__version__
 
 # Allowed instruments
 INST_LIST = ['nircam', 'niriss', 'fgs']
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -0,0 +1,5 @@
+# pyproject.toml
+[build-system]
+requires = ["setuptools>=41.2.0",
```
jdavies-st: This version of `setuptools` is too strict.  That looks like the latest one, but it will work with earlier versions as well.  You can probably not specify a version for `setuptools` or `setuptools_scm` and it will be fine.

mirage/dark/dark_prep.py
```diff
@@ -34,14 +34,12 @@
...
-
+import mirage
 from mirage.utils import read_fits, utils, siaf_interface
 from mirage.utils.file_splitting import find_file_splits
 from mirage.reference_files import crds_tools
 
-from mirage import version
-
-MIRAGE_VERSION = version.__version__
+MIRAGE_VERSION = mirage.__version__
```
jdavies-st: You could also do something like
```
from mirage import __version__ as mirage_version
MIRAGE_VERSION = mirage_version
```
or dispense with the global variables and just use `mirage.__version__` everywhere it is needed.

jdavies-st:
You may also want to do a grep of `relic` and find out if you are doing any .gitignore or anything else with it (excluding it from packaging).

pyproject.toml
bhilbert4: Removed version number for `setuptools_scm` and lowered the version number for `setuptools` to match that in the `jwst` repo (38.2.5)

mirage/dark/dark_prep.py
bhilbert4: Good idea. I switched everything to use `mirage.__version__`

bhilbert4:
Also removed a reference to relic in MANIFEST.in. I don't see any other mentions of it.

jdavies-st:
After you've merged this, you'll want to delete the `relic` directory in your local checkout too.  👍

Issue #429, opened 2020-01-09
Some subtle issues in NIRISS imaging or WFSS simulations.


KevinVolkSTScI:
Chris Willott has pointed out that the background parameter "bkgdrate" is interpreted as ADU/second/pixel in imaging simulations but it is interpreted as electrons/second/pixel in the WFSS mode.  The comment in the imaging .yaml exampl files for this parameter is "#Constant background count rate. Number (ADU/sec/pixel) or 'high','medium','low' similar to what is used in the ETC", as in the example file imaging_test_for_notebook.yaml for example.  Apparently Chris is assuming that this is true also for WFSS parameter files.  I see that the example WFSS .yaml files such as wfss_f250m_test.yaml have a comment that the value of bkgdrate is in electrons/second/pixel.  The question is whether this should just be highlighted in the documentation or whether the code should be changed to make the units the same for both cases?  It would seem to be better to make the units of bkgdrate the same in all modes.  

Related to this issue, for NIRISS there is an issue in the gain reference file in that the values are "tiled" in 20x20 pixel blocks.  This is not a concern for the pipeline where the gain is only used in the noise statistics.  When Mirage uses the gain map for WFSS simulations it introduces small discontinuities in the spectra if these cross a tile boundary, as most of them do.  One could fix this in Mirage by using a mean gain value rather than the gain map to convert ADU/s to electrons/s.  If the code goes both ways as in imaging where the background is in ADU/s and is multiplied by the gain reference file and the frame time to get electrons/frame and then converted back to ADU for the output the tiling mostly disappears.  It would disappear entirely if the non-linearity were not applied.  For WFSS the background is in electron/s/pixel and then the gain map is only used to convert to ADU for the output ramp, and so the imprint of the gain tiles is present in the simulated frames.

Personally I would rather have Mirage use the mean gain value from NIRISS the gain reference file for all cases where the values need to be converted from ADUs to electrons, since I am not certain that the structure seen in the NIRISS gain map is actually real--one does not see analogous variations in the CV3 flat field images, and the gain should really be one value per channel not dependent on the position within the area using a given output amplifier.  It may be the gain measurements for NIRCam are better than what we have for NIRISS.  Chris tells me that he has changed the Mirage code to use the median gain rather than the gain map for NIRISS to get rid of the tile effect in the simulations.

[...]

PR #430, opened 2020-01-10
Adjustments to bring backgrounds in line with ETC

### LINKED ISSUE #426: Mirage backgrounds for NIRISS appear to be incorrect and inconsistent between the imaging and WFSS  modes.
The background values added to the seed images for NIRISS imaging and NIRISS WFSS spectral observations are not consistent with the values from the ETC or the values generated directly from jwst_background.  The ratio of the backgrounds in imaging and WFSS spectral observations are also not consistent with what is expected or what a simple scaling of one value to the other would produce.  For the ETC the WFSS backgrounds are always about a constant factor smaller than the imaging background for the same set-up.
  I have used the ETC to get background values for imaging and WFSS modes.  It is difficult to determine exactly how many pixels are in the signal aperture for WFSS mode, although it looks like it is 1 pixel times twice the defined aperture "height", but for imaging the size can be directly calculated.  Comparison of these values for a few sky positions have shown that the Mirage values from the current version (installed 30 December) are not the same as the ETC values and are always much smaller.
  For one particular case, the proposed new NIRISS focus field at position RA = 85.22458 degrees and DEC = -69.5225 degrees the Mirage imaging background values in some of the filters are

F090W    medium    0.1360 ADU/s/pixel
F115W     medium    0.1875
F150W    medium     0,1677
F200W    medium    0.2239

and for the wfss mode the values are

Filter       Low         Medium       High
F090W   0.2834    0.3068       0.3235    electron/s/pixel
F115W    0.2788    0.3028        0.3164
F150W   0.2360    0.2556        0.2656
F200W   0.1941     0.2091        0.2159

One would expect that the count rate for the background in WFSS mode will be about 0.8 times the value in imaging mode.  With a gain value of 1.612 the imaging background values are generally smaller than the WFSS equivalents, although for the F200W case the imaging value is a little larger than the WFSS value.  But all these values are much lower than the ETC values which are 

Filter        Low       Medium      High
F090W   1.545        1.709        1.772    electrons/s/pixel
F115W    1.560        1.705        1.768
F150W   1.223         1.316        1.365
F200W   1.125         1.208        1.253

The Mirage WFSS background values are roughly a factor of 5 too small.
  The other oddity besides that the WFSS background is higher than the imaging one is that the ratio values between the filters are not the same for imaging and WFSS modes.  The ETC values for the F090W and F115W filters are fairly close, but in the Mirage case these are not similar.
  I wrote a code to use jwst_background.jbt to get the background spectra at a given position and then integrate over the NIRISS response to get the expected background signal values.  My code agrees with the ETC values to about 0.5%.  Hence I think the issue is in how Mirage is calculating the background values.

bhilbert4:
This PR makes some adjustments to the way background levels are calculated, in order to bring those backgrounds in-line with the ETC. 

- Backgrounds for NIRISS WFSS are now explicitly 80% of those for imaging mode

- When calculating the background value from a background spectrum, Mirage integrates the spectrum over the filter bandpass, and divides this by the integral of the filter bandpass itself. This should give acceptable results unless the background spectrum is not smooth. Currently that is never the case in jwst_backgrounds.

- Fixed a bug in Grism TSO mode where the background was being added to the seed image twice

- Updated Grism TSO mode to create a dummy point source catalog in the case where the user provides only a grism TSO catalog. This is needed because the background is calculated and added to the background sources.

- calculate_background moved into backgrounds.py so that it is accessible to catalog_seed_image as well as wfss_simulator, grismt_tso_simulator, etc.

Resolves #426


bhilbert4:
@KevinVolkSTScI I think this solves the issue of the backgrounds not matching those calculated by the ETC. Do you have a chance in the near future to grab this branch and try out the cases you describe in #426? I inserted a few pytest functions to compare results to ETC results, and those are passing, but it couldn't hurt to test a few more cases.

bhilbert4:
Note also that this PR solves the first issue mentioned in #429. For WFSS mode, if the user provides a number in the `bkgdrate` field, Mirage now interprets this number as ADU/sec/pixel, just as is done for imaging modes.

KevinVolkSTScI:
Yes I will take a look at it soon.

KevinVolkSTScI:
I installed the branch and ran one test of WFSS background.  Setting the background value to 1.0 still gives a background level of 1.0 in the dispersed seed image.  This appears to be the same as before.  I only got one test run in before an astropy error appeared.  Something seems to be wrong in the conda installation so I am trying to fix that now.

bhilbert4:
Ah ok. So when a number is input for the background rate, you want that to be the rate in the undispersed image?

KevinVolkSTScI:
When the imaging background for NIRISS is 1 ADU/second/pixel the WFSS background should be around 1.3 electron/s/pixel.  When I ran the test case the signal level of the background was 1 electron/s/pixel the same as before.  Unfortunately conda seems to have gotten into a bad state somehow and i cannot currently run Mirage at all to test the code further.  I am trying to recover but it looks like I need to remove the entire conda installation and install conda from scratch.

KevinVolkSTScI:
I must have done something wrong in my installation of the branch...both using bkgdrate = low and bkgdrate = 1.0 produce the same results as for the 30 December version of Mirage.  I will try again to get the proper version and install it.

KevinVolkSTScI:
I have run a couple of the test models but the backgrounds remain unchanged.  For the position I was looking at (85.22458, -69.5225) the bkgdrate = low value is about 0.236 electrons/s/pixel and when I set the bkgdrate to 1.0 I get 1 electron/s/pixel out.  I must be doing something wrong as github refuses to download the changes in your branch niriss-backgrounds, and then when I copy over the code from the changed version of catalog_seed_image.py it seems to not use the new file in the calculations.  I can see the change for the integration over the filter profile and dividing the background rate by this value in the file, line 4312, 

filt_integ = np.trapz(filt_thru, x=filt_wav)

but it does not seem that this code is executed when I run the model.  I do not see anything in the code that scales the bckdrate by the gain in the case where the parameter is a number.  I just see the value being read and assigned to the parameter value.  So I cannot seem to test the code.  I do not know what I am doing wrong in the installation of the code.  I always seem to have issues with github.

bhilbert4:
It seems like you're looking at an older version of the branch. I changed the code to change MJy/sr to electrons/sec before integrating.

bhilbert4:
In the latest version of the code, if I run a NIRISS imaging case in F090W with the background set to low, I get 1.59 e-/pix/sec (0.99 ADU/pix/sec). 

For a WFSS observation with the F090W filter and the background set to low, I get a value of about 1.29 e-/sec/pixel, which is about 80% of the 1.59 e-/sec/pix value in imaging mode.

For a WFSS observation with the F090W filter and the background set to 6.0 in the yaml file (which is assumed to be in ADU/pix.sec), I see background values of about 9.6 e-/sec/pixel in the dispersed seed image, which is as expected given the gain value of 1.612 e-/ADU. 

One subtlety that occurred to me yesterday: When starting from an APT file, the user can specify one background level per observation. A WFSS observation will be composed of both the dispersed and the direct images. The code as it is will put the same background value in the yaml files for all exposures. But these input values are treated differently depending on the mode. For example, a background value of 1.0 in an imaging yaml will result in 1 ADU/pix/sec being added to the seed image. A background value of 1.0 in a WFSS yaml file will result in a mean background value of 1 ADU/pix/sec in the dispersed seed image. But a background of 1 ADU/pix/sec in the dispersed seed image implies that the background in a direct image would be 1.0 / 0.8 = 1.25 ADU/pix/sec, which is inconsistent with the value in the actual direct image. Do you think it makes sense to treat the background value in all input yaml files as the background in ADU/pix/sec in an undispersed scene?

I think this is only an issue for NIRISS, since NIRCam WFSS mode only allows background values of low/medium/high, or gets the background based on date.

KevinVolkSTScI:
Yes I had thought about this.  On one hand setting the bgkdrate value always to the imaging value would mean that the inputs would be consistent between the different .yaml files.  On the other hand the output background will not match the input background in the dispersed images.  I think on balance it may be better to make the value the imaging value and multiply by 0.8 in the dispersed case, and just have some documentation of this somewhere to alert users about this detail.  When we are on-sky with NIRISS we may then need to revise the ratio of 0.8.

bhilbert4:
The other option would be to have the yaml_generator multiply the input background value by 0.8 and populate the background entry in the wfss mode yaml files with that value. I haven't looked at the yaml_generator to see how easy or difficult this would be, but my impression is that it wouldn't be too hard.

The other side of the coin would be users who are manually creating wfss mode yaml files. I don't know how common this will be compared to those generating yamls from APT files. But in this case, would it make more sense for those users to specify the direct or dispersed background rate?

bhilbert4:
And of course this only applies when numbers are given for the background. If the user specifies 'low', then the resulting background values will be different depending on mode.

KevinVolkSTScI:
I am thinking in terms of people doing manual editing of the .yaml files, or maybe using a code, once the initial set of .yaml files are made with yaml_generator.  It is nice to have the files from .yaml_generator to start with, but I would think people are likely to want to run additional simulations for a variety of backgrounds or exposures without going through the APT file every time.  I tend  to do this type of thing in testing.  Maybe this is not what most people will do.

KevinVolkSTScI:
For use of high/medium/low backgrounds in NIRISS WFSS one gets the number from imaging times 0.8.  So in that sense the same input produces backgrounds with a set ratio and it may therefore be more consistent to do the same when a number is used.

bhilbert4:
That makes sense. I'll put in a multiplication by 0.8 for WFSS observations then, and adjust the documentation to make clear that the input background level is that for direct images.

bhilbert4:
@KevinVolkSTScI let me know if you get conda working and/or if you think this is ready to merge.

bhilbert4:
Actually, I spent some time this morning updating the installation documentation. So if you want to start over and create a new mirage environment, here is what I put in the documentation:

```
    conda create -n mirage python=3.6 -y
    conda activate mirage
    cd /your/dir/containing/mirage
    pip install -e .
```
After that you can try grabbing this branch again.

KevinVolkSTScI:
Hi Bryan,

  After some looking at “stack overflow” I know how to get the right branch so I will test this today.

Kevin

KevinVolkSTScI:
I ran a few test cases and the code seems to be working.  When I enter a background value of 1.0 in the bkgdrate parameter the background value averages out to 1.284 electrons/second/pixel in the dispersed seed image.  When I use "low", "medium", and "high" I get values of 0.99, 1.07, and 1.12 for this particular sky position.
However I did see a problem.  When I have "config" for the "filter_throughput" parameter the code does not seem to replace this value by the proper throughput file name.  Instead the string "config" is passed to the call to read_filter_throughputs on line 71 of backgrounds.py.  The code then fails because this file does not exist in my current directory.  When I put in the actual file name for the filter I want to use in the WFSS simulation the code works as expected and reads in the filter throughput file.  One does not want to have to locate the filter throughput file and put it in by hand.  I do not know what is causing the code not to change the value of "config" for this parameter since I have other parameters in the refFiles section of the yaml file set to "config" such as filtpupilcombo that I think is being used in the code.

bhilbert4:
filter_throughput is called out specifically in "wfss_simulator", while the other entries are not. I added a check to the line in "wfss_simulator", so "config" is now an acceptable input.

bhilbert4:
Oops. I accidentally put the updates for switching over multiplication by the pixel area map to the flat field here, instead of #431 as I had intended. Too many active branches. Anyway, flat field updates are in, and my tests are all passing. Note that NIRCam's grism flat reference files are currently dummies (all 1.0's), and so even though Mirage is multiplying by the flat, it's not changing the data at all. Flats will be updated after launch.

bhilbert4:
@KevinVolkSTScI just a note that I plan on merging this branch, as well as #431  this afternoon, unless you have major concerns.

PR #431, opened 2020-01-13
Use mean gain values rather than gain reffiles

bhilbert4:
This PR addresses issues brought up in #429 

I have updated the code to use mean gain values rather than gain reference files. The binned gain reference files were causing discontinuities in images after multiplying by the gain. This was seen in NIRISS and also possibly in NIRCam WFSS observations.

Changes have not yet been made for the NIRISS occulting spots...


bhilbert4:
@KevinVolkSTScI, I'm making updates now to multiply the seed image by the flat field. For WFSS mode, do we want to multiply the direct seed image by the flat field before dispersing? That seems backwards for features other than the occulting spots in the flat field file, right? I would think that ideally Mirage should take the direct seed image, add in the effects of the occulting spots, disperse the seed image, and then multiply by the flat field (excluding occulting spots). 

Also, the occulting spots are visible in the WFSS background file. Does this mean we don't need to multiply the background image by the flat field?

KevinVolkSTScI:
The WFSS background images have the effect of the spots in them and one does not need to (and in fact definitely should not) multiply by the flat field image.  For NIRISS WFSS we do not have monochromatic grism flats, which we should have done if we had been able to find the time prior to delivery of the instrument, and the grism flats are just "patched" versions of the imaging flats where the obvious POM features including the occulting spots have been patched over.  The hope is that the flat will allow for the pixel to pixel variations properly.  It is not going to work for the distortion effects from the grism, or the illumination variations across the detector in the dispersed images.  That is going to be a limitation on the data reduction for WFSS mode.
  For imaging mode one wants to multiply by something like the filter flat field image.  One actually wants only the POM part of the flat field structure, not the detector pixel part of the flat field image, for the seed image to be dispersed.  That will put the spots in the right places (more or less...there are some subtle issues in the predicted dispersed background compared to the observed one, but in general they are similar).  But, we do not have a good way to separate out the detector part from the POM part of the flats, so we just have to live with imprinting some of the detector structure (i.e. the cross-hatching) onto the dispersed images.  If it is just imaging the full flat field can be used because there is no offset of light from one area of the detector to another as in the grism cases.

bhilbert4:
@volk, @NorPirzkal ok, I'm going to proceed with slightly different methods for NIRCam and NIRISS. Let me know if this sounds ok.

Imaging mode: same for both instruments. The flat field will be multiplied into the seed image, which includes all sources and the background.

WFSS mode:
NIRISS: 
- Multiply the flat field into the seed image when it contains the sources only (no background).
- Disperse, and add in background from background file
- No flat field applied after dispersion

NIRCam:
- Do not multiply flat into the direct seed image
- Create background image from input spectrum
- Disperse, and add background
- Multiply flat field into dispersed seed image

KevinVolkSTScI:
We have to see what the results look like, but that should work for NIRISS.

NorPirzkal:
I would like all of us to specifically call the pixel to pixel flat-field, a pflat. There is a need to also have an illumination flat-field, which has nothing to do with the detector. I discussed with with Bryan and think that we need to allow for a pflat to be applied after the dispersion is performed. Optics issues such as coronagraph support is something that affect what the grisms sees and can be also approximated by a “flat-field”. In this case it is more like a 2D transmission filter but it is essentially another image we multiply by at some stage. That stage should be before we disperse anything and it should be applied to any seed image where such an effect needs to be taken into account.
Mirage therefore should have both options available, but let’s be careful to not confused pflat with things like gain corrections and transmissions.

On Jan 21, 2020, at 4:26 PM, Bryan Hilbert <notifications@github.com<mailto:notifications@github.com>> wrote:


@volk<https://github.com/volk>, @NorPirzkal<https://github.com/NorPirzkal> ok, I'm going to proceed with slightly different methods for NIRCam and NIRISS. Let me know if this sounds ok.

Imaging mode: same for both instruments. The flat field will be multiplied into the seed image, which includes all sources and the background.

WFSS mode:
NIRISS:

  *   Multiply the flat field into the seed image when it contains the sources only (no background).
  *   Disperse, and add in background from background file
  *   No flat field applied after dispersion

NIRCam:

  *   Do not multiply flat into the direct seed image
  *   Create background image from input spectrum
  *   Disperse, and add background
  *   Multiply flat field into dispersed seed image

bhilbert4:
pflat definitely needs to be multiplied in to the data, since one of the pipeline steps divides the data by the pflat. 

Anyway, given the looming deadline, I'm going to fit in these changes in probably a less-than-optimal way. But I do like the idea of going to a model that has separate pflat and lflat options eventually, assuming that there will be a need for it at some point (i.e. no one has lflats at the moment).

KevinVolkSTScI:
Hi Nor,

  Yes I agree with the terminology.  However the pipeline currently has no mechanism for distinguishing the two flats, and for NIRISS imaging we only have the combined flat including both components so we are somewhat hampered in such discussions.  Originally I was arguing that the pipeline should have the separate pixel flats and illumination flats as inputs to be applied, but this was not adopted and there is only the one combined flat field image used.  One can of course have both types in Mirage where needed.  One also has the problem of separating the pixel area effects from the pixel to pixel detector variations and from the illumination changes.  All of these are in the “flat field” images used by NIRISS even though each of these is from a different cause.

Kevin

NorPirzkal:
I think that we should not be limiting  ourselves by the pipeline design. I would like to see Mirage being able to generate as good of a simulatikn of every effects we might encounter. Whether the pipeline handles it properly is another matter and being able to see how well the pipeline can approximate things, or fail, is the best way to move forward.

Sent from my iPhone

On Jan 21, 2020, at 17:55, Kevin Volk <notifications@github.com> wrote:

﻿Hi Nor,

Yes I agree with the terminology. However the pipeline currently has no mechanism for distinguishing the two flats, and for NIRISS imaging we only have the combined flat including both components so we are somewhat hampered in such discussions. Originally I was arguing that the pipeline should have the separate pixel flats and illumination flats as inputs to be applied, but this was not adopted and there is only the one combined flat field image used. One can of course have both types in Mirage where needed. One also has the problem of separating the pixel area effects from the pixel to pixel detector variations and from the illumination changes. All of these are in the “flat field” images used by NIRISS even though each of these is from a different cause.

Kevin

bhilbert4:
Oops. After all this, I accidentally put the flat field updates into #430 rather than here. 

This PR contains only the update to use mean gain values rather than gain images.
Resolves #429

PR #433, opened 2020-01-14
Besancon catalog function updates

### LINKED ISSUE #427: Update Besancon-related code
The team that maintains the Besancon model have significantly changed the way that users run the model and get their results. Basically, astroquery is broken (I reported this to them several months ago). With the updated model interface, a user must:

1) create an account on https://model.obs-besancon.fr/modele_home.php
2) log in and either populate a webform by hand, or download the python script created by the team and call that from the command line.
3) When the results are ready, you get an email with a link to an ftp site from which you can get the table of your results

I could conceivably create a modified version of their python program that would allow users to submit their query programmatically, but I don't see a way around having to click the ftp link in the email and download the results. 

So, I think the Besancon-related code in Mirage should be updated such that it takes a user-provided catalog conforming to the format of the table retrieved from the ftp site. From that point on, I don't think Mirage will need any changes. In this way, if Mirage users have some other method of creating or obtaining a table of stars with V, V-J, V-K, etc, they can still use the Besancon-related code in Mirage to turn that table into a Mirage-formatted catalog with source magnitudes that have been converted to the JWST filters of interest.

bhilbert4:
The manner in which the Besancon model is queried has recently changed. The query via astroquery is broken. Users must now get an account at https://model.obs-besancon.fr/modele_home.php. Queries are now done using a client provided by the group that maintains the model. The query is sent to the server, which then sends the user an email when the results are ready. The user must then download the results using an ftp link in th email.

I have adjusted the Besancon-related functions in create_catalog.py to deal with this new workflow. The functions have been made a little more general, such that a Mirage user can now translate any catalog with source VJHK magnitudes into JWST filter-based magnitudes. The catalog need not be from the Besancon model. 

The besancon() function is now a wrapper around the call to the command-line tool written by the Besancon model team.

I have also modified the match_model_magnitudes() function to use array-based calculations, which has sped up the conversion from VJHK to JWST magnitudes by a factor of several hundred.

Still to do: 
Include the Besancon client tool in the repo, for convenience.
Modify functions that combine Besancon catalogs with catalogs from other sources.
Update the appropriate notebook 

resolves #427


bhilbert4:
Also, in the read_standard_magnitudes() function of create_catalog.py, there is a list of standard_filters. The second and third entries in that list are both ‘Johnson H’. The first of these should probably be ‘Johnson J’. Kevin will confirm.

bhilbert4:
Comparison of the filter list in `read_standard_magnitudes()` with that in config/magslist_bosz_normal_mirage.new shows that the first of the two "Johnson H" values should indeed be "Johnson J". I have fixed this.

bhilbert4:
The latest commits have taken care of everything on the 'to do' list. The updated notebook is functioning correctly.

PR #434, opened 2020-01-17
Update for pip install

bhilbert4:
This PR makes adjustments to setup.py to support installing Mirage from Pypi. Entries from environment.yml have been copied into `install_requires` within setup.py. 

Documentation on installing Mirage has been updated to describe how to install from Pypi, as well as how developers can get the latest development version (by cloning and installing).


bhilbert4:
Add ipython and notebooks to the environment.

PR #444, opened 2020-01-29
Small but important bugfixes

bhilbert4:
The TSO simulator notebook was missing an import statement for the recently added MEAN_GAIN_VALUES. This PR adds that import statement. 
 
It also removes an unused line from the TSO example notebook that was causing the `multi_spec_plot` function to crash. Not sure how these problems made it through, since I ran the notebook so recently.


bhilbert4:
Add a missing "self." in front of `add_segmentation_image` for extended objects so that the segmentation image is correctly updated.

bhilbert4:
Resolves #445

Issue #457, opened 2020-02-13
Tilt-to-pixel scaling issue in segments_psfs.py


Skyhawk172:
While generating mosaics using unstacked segments we noticed that the SW and LW mosaics were different. In particular, the segment offsets on the LW data looked consistently scaled up from the SW data.  See attached.

We narrowed it down to the fact that the tilt-to-pixel scaling was only valid for the SW detectors. The fix should be easy to make, which I will provide as soon as possible.

Note that we confirmed that nominal MIRAGE SW and LW mosaics agree with each other.

![SW_vs_LW](https://user-images.githubusercontent.com/9323123/74470317-5227dc80-4e6c-11ea-9c18-4c41eb36d502.gif)

bhilbert4:
Resolved via #462

PR #462, opened 2020-02-19
Add Optional telescope boresight offset & Fix tilt scaling issue between SW and LW segment offsets

Skyhawk172:
Fixed XTILT/YTILT header keyword description to micro radians.

Fixed 'get_segment_offset': docstring originally suggested returned values for
offsets were in pixels, but it turns out they are in arcseconds.  Because of
that confusion, the SW pixel scale was applied to LW data resulting in a scaling
of the offset.  This is now fixed and does not require pixel scale knowledge.


Skyhawk172:
This is in response to https://github.com/spacetelescope/mirage/issues/457

mirage/psf/segment_psfs.py
```diff
@@ -59,6 +59,10 @@ def generate_segment_psfs(ote, segment_tilts, out_dir, filters=['F212N', 'F480M'
     fov_pixels : int, optional
         Size of the PSF to generate, in pixels. Default is 1024.
 
+    boresight: list, optional
+        Telescope boresight offset in V2/V3. This offset is added on top of the individual
```
bhilbert4: Are users going to know that the input boresight offset values should be in units of arcmin? Or should that info be added here to the documentation?

mperrin:
I agree, it seems like a good idea to add a label about the units for the offset. Especially because the code seems to use a mix of microradians, arcseconds, and arcminutes in different places! (Charles, do you think it might be worth switching the offset to be in arc seconds, so there’s one fewer different units mixed in? I’m OK with it either way, as long as it’s clear in the doc strings.)

Skyhawk172:
I thought I had put the arcminutes in the docstring. I'll add it.

PR #463, opened 2020-02-28
Non nominal PSF inputs

bhilbert4:
This PR updates the WFSC aspects of Mirage in order to support OTEs 1-7. Development was done by @Skyhawk172 and @bhilbert4 

- The unstacked mirror segment notebook and non-nominal PSF notebook have been updated to work with the latest changes in Mirage. 

- Tweaked the code responsible for creating the PSF libraries that go with these notebooks in order to speed up execution time. 

- A bug was fixed that was resulting in the improper normalization of the mirror segment PSFs when ITM data are used.

- Fixed a bug in yaml_generator that was miscounting the number of exposures/activities in the APT file for cases where the non-default PSF libraries are used. This only affects WFSC simulations.


bhilbert4:
A new WFSC notebook was also added to the set of example notebooks. There are now 3 WFSC notebooks in the repo.

1.  Unstacked_mirror_simulations_use_examples.ipynb:
            - Unstacked segments, no mirror moves.
            - Use for OTE-01
2. Unstacked_config_with_mirror_moves.ipynb:
            - This is the image-move-image scenario with unstacked segments
            - It will create the various folders for each exposure before/after a mirror move
            - Use for OTE-03 and OTE-04 (and OTE-06)
3. Nonnominal_PSF_simulations_use_examples.ipynb:
            - This is also for the image-move-image scenario but for when we are using one input containing all the segments, instead of creating a library of PSF for each single segment as in bullet 
            - Use for OTE-07 and onward.

bhilbert4:
@Skyhawk172 and I agree that these updates are now functioning as expected. This can be merged into the development branch.

PR #467, opened 2020-03-18
Add APERNAME value

bhilbert4:
This PR adds code to populate the APERNAME header keyword in the simulation files output by Mirage.  The aperture name has always been used by Mirage. This update just adds that name to the APERNAME keyword. This keyword is not used by the ``jwst`` calibration pipeline, even though it is in the datamodel schema. Folks working on WFSC simulations requested this keyword in order to help with their data analysis.

PR #469, opened 2020-03-19
Adjust tests to get around Travis-astroquery issue

bhilbert4:
For some reason, Travis tests started failing in #467 and #468 within the test where astroquery is used to query 2MASS. Tests pass locally, but on Travis there were errors with SSL. This PR reworks the failing tests to: 

1. Skip test_2mass_catalog_generation(), since the only point of this test is the 2MASS query itself
2. Rework test_catalog_combination() to read in an existing catalog rather than querying 2MASS in order to get a catalog to test the catalog combination step.

### DIFF
--- a/tests/test_catalog_generation.py
+++ b/tests/test_catalog_generation.py
@@ -89,17 +89,6 @@ def test_2mass_catalog_generation():
 def test_catalog_combination():
     """Test the combination of two existing catalogs into one
     """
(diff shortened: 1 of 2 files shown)


bhilbert4:
Also skipping a test that downloads and reads in an ITM PSF library, as the download was timing out. This test can be run locally for now.

PR #471, opened 2020-03-23
Updates to documentation, env, and testing for python 3.7

bhilbert4:
This PR updates Mirage to use python 3.7 as the default version. Documentation for environment creation has been updated from python 3.6 to 3.7. Travis testing is now done using python 3.5, 3.6, and 3.7 (previously only 3.6).  The environment file has also been updated to specify python 3.7.

Tests have shown undiagnosed problems when running with 3.8, which is why it's not included here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,9 @@ compiler: gcc
 sudo: false
 
 python:
-    - 3.5
-    - 3.6
-    - 3.7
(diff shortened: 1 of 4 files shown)


bhilbert4:
The changes here have already made it into the repo in various other PRs since this one was created. No need to merge this.

PR #473, opened 2020-04-08
Parallelize segment PSF computations

mperrin:
This PR revises `mirage.psf.segment_psfs.generate_segment_psfs` to parallelize the calculation  of PSFs for the 18 segments. This is straightforward via use of `multiprocessing.Pool`, plus some rearranging to split out the calculation work into a separate helper function that can be called per each segment. 

This results in significant performance increases, depending on the hardware capabilities and CPU count of whatever machine it's running on. 

Incidentally this also adds an `nlambda` option, so that the number of wavelengths used in the calculation can be tuned, rather than hard-coded at 10.


mperrin:
Tests by @Skyhawk172 show a >10x speedup compared to the single process version, when running on telserv3.

bhilbert4:
@Skyhawk172 @mperrin would you try running pytest on your branch? The tests passed on Travis using linux, but I keep getting timeout errors with macOS. I'm assuming this is more a problem with Travis than with the repository. I don't want to hold up this PR if you want the updates merged if this is a problem with Travis.

Skyhawk172:
Just running test_segment_psfs fails for me because the pre-calculated offsets in the parametrized_test_data.yaml file (using the psf library on central store) don't match what the function get_segment_offsets returns.  We did have to flip a sign in that function not too long ago but I don't remember changing the test values. But what puzzles me is why we haven't caught that earlier and why Travis passes on linux... I'll have to investigate.

mperrin:
@bhilbert4 FYI - I added a second commit to this PR, not related to parallelization directly but also improving the segment PSFs calculation. Seemed better to add to this PR than to make a separate one. 

I saw the Travis job froze or hung, and agree that's weird. For now I'm trying to restart that one job manually just to see. But since it appeared to be hanging in a completely unrelated test, I agree that this seems most likely to be some upstream Travis problem. 

That said @Skyhawk172 is right, we'll also need to update the test data for the other test using the Central Store data at some point (though that test just gets skipped on Travis so it's clearly not part of the issue here).

mperrin:
Travis build timed out / error again. I do think it's likely something not related to this branch. But it's also not critically urgent that we merge this right away, so we can take some more time to investigate. But I have very little idea how to proceed with investigating what's up with Travis..

bhilbert4:
@mperrin try pulling the master branch again. I just increased the Travis timeout threshold in a separate PR and the OSX tests had time to pass. I just merged that branch into master.

mperrin:
Rebased onto current master, and now watching the tests run again. 

(Pretty hefty set of tests if they take 30 minutes to run! Or is it just that the Travis mac builds run on relatively slow hardware?)

bhilbert4:
The latter is true. In addition, prior to a couple weeks ago the OSX tests were running much faster than they are now. I'm not sure what changed.

mperrin:
Excellent, looks like all tests are passing now!

bhilbert4:
Anything else to add, or should this be merged? @Skyhawk172, did you want to add the changes to the test that is skipped on Travis that we were looking at a couple weeks ago?

mperrin:
@bhilbert4 - @Skyhawk172 and I just added a commit to fix the test that's skipped on Travis but runs locally. I've just ran the test suite and verified that everything in `test_segment_psfs.py` passes here for me. 

As it's passed on Travis too, we believe this is good to merge.

bhilbert4:
Excellent! Merging now.

PR #478, opened 2020-05-05
Dark entries in yaml_generator if no darks are present

### LINKED ISSUE #477: Mirage fails if raw darks are not present
@dancoe points out that even if the user requests linearized darks, Mirage still runs get_dark(), which looks for the raw as well as the linearized darks. If the raw darks are not present, then Mirage will crash when attempting to generate a list of random index numbers equal to the length of the list of darks. 

The code should probably be updated such that it doesn't bother looking for raw darks if the user requests linearized darks, and vice versa.

The crash is in get_dark(), on the line: 
rand_index = np.random.randint(0, len(files) - 1)

bhilbert4:
This PR addresses a bug where the yaml_generator crashed in the case where only the linearized dark current inputs were present. The yaml_generator searches for both raw and linearized input darks regardless of which type the user requests. In the case where someone has downloaded only one kind of darks, this would lead to the yaml_generator crashing when the other type of darks was not present.

With this fix, yaml_generator still searches for both types of darks, but returns None if no dark file is found.

Resolves #477

### DIFF
--- a/mirage/yaml/yaml_generator.py
+++ b/mirage/yaml/yaml_generator.py
@@ -661,9 +661,15 @@ class SimInput:
         if self.use_linearized_darks:
             self.info['dark'] = [None] * len(darks)
             self.info['lindark'] = lindarks
+            if set(lindarks) == set([None]):
+                raise RuntimeError(("ERROR: Linearized darks requested, but no linearized dark files "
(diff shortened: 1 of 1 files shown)


bhilbert4:
I've tested these fixes for both the linearized and raw darks, and everything works as expected. Now, if the user requests a type of dark whose reference files do not exist, rather than returning a cryptic error statement, Mirage will raise an exception and inform the user that the requested type of dark files are not present in the reference file directory.

Issue #486, opened 2020-05-20
Update xml parser to grab FGS exposures in Global Alignment template


bhilbert4:
The xml reader currently searches only for the NIRCam exposures in the global alignment template. However, the observing program also includes FGS exposures. The global alignment template reader in read_apt_xml.py needs to be updated to find and add these FGS exposures. 

Once the exposures have been added to the exposure dictionary, be sure that the information from the pointing file is correctly applied to the two instruments.

bhilbert4:
This will be a little more complex than I first thought. I added a few lines to get Mirage to read the FGS exposures in the GA template. But I ran into 2 issues.

1. The APT reader at the moment makes assumptions about the number of exposures associated with each exposure type (ADJUST2, CORRECT+ADJUST, etc). This information doesn't look to be in the xml file, so we should double check that these values are correct. It does appear that these numbers can be inferred from the number of rows in the pointing file, but the xml is read in prior to the pointing file.

2. Since the pointing file contains rows only for NIRCam, Mirage is getting confused when reading in the pointing file and trying to match up rows to the list of exposures from the xml file. So it seems that we will need to add some specialized bit of code to the pointing file reader to add rows for FGS.

mperrin:
For point 1 I was looking at that yesterday and the numbers are not correct - I fixed this in my local copy and didn't PR it yet. Here's what I recommend: 

```
        if GA_iteration == 'ADJUST1' or GA_iteration == 'CORRECT':
            # 3 NIRCam and 1 FGS images
            n_exp = 4
        elif GA_iteration == 'ADJUST2' or GA_iteration == 'CORRECT+ADJUST':
            # 5 NIRCam and 2 FGS
            n_exp = 7
        elif GA_iteration == 'BSCORRECT':
            # 2 NIRCam and 1 FGS
            n_exp = 3
```

Somewhat lower in the same function we'll want something like this to get the exposures assigned correctly: 

```

        # Repeat for the number of exposures
        for j in range(n_exp):
            # Add all parameters to dictionary
            if j==2 or j==5:
                # This is an FGS image as part of GA
                
                ## TODO update this for FGS parameters instead of NIRCam
                tup_to_add = (pi_name, prop_id, prop_title, prop_category,
                              science_category, typeflag, mod, subarr, pdithtype,
                              pdither, sdithtype, sdither, sfilt, lfilt,
                              rpatt, grps, ints, short_pupil,
                              long_pupil, grismval, coordparallel,
[...]

PR #488, opened 2020-05-21
WFSC Global Alignment fixes

mperrin:
Fixes in support of #486

- [x] Improve parsing to include FGS images as well as the NIRCam images
- [x] Generalize segment PSF generation code to be able to make FGS PSFs as well as NIRCam ones.


mirage/apt/read_apt_xml.py
```diff
@@ -927,16 +926,40 @@ def read_globalalignment_template(self, template, template_name, obs, prop_param
...
                 long_pupil = 'CLEAR'
 
-        # Repeat for the number of exposures + 1
-        for j in range(n_exp + 1):
+        for fgs_conf in ga_fgs_configs:
+            fgs_grps = fgs_conf.find(ns + 'Groups').text
+            fgs_ints = fgs_conf.find(ns + 'Integrations').text
+
+        # TODO read guider detector requirement from GuideStarID special requirement in this case
+        guider_det_num = 1  # PLACEHOLDER ONLY
```
kjbrooks, line 934: We do this in MAGIC so if you are interested you can find APT query here: https://github.com/spacetelescope/jwst_magic/blob/3b9f9933cb4bcdb49f8874cd45e1a2972b023a87/jwst_magic/masterGUI.py#L1263
Specifically lines 1333-1355:
`# Pull: List of Observations for the CAR > Specific Obs > Special Requirements Info (sr)
        namespace_tag = '{http://www.stsci.edu/JWST/APT}'
        observation_list = tree.find(namespace_tag + 'DataRequests').findall('.//' + namespace_tag + 'Observation')
        observation = observation_list[obs_number - 1]  # indexes from 1
        sr = [x for x in observation.iterchildren() if x.tag.split(namespace_tag)[1] == "SpecialRequirements"][0]

        # Try to pull the Guide Star information
        try:
            gs = [x for x in sr.iterchildren() if x.tag.split(namespace_tag)[1] == "GuideStarID"][0]
        except IndexError:
            self.lineEdit_normalize.setText('')
            LOGGER.error("Master GUI: This observation doesn't have a Guide Star Special Requirement")
            raise ValueError("This observation doesn't have a Guide Star Special Requirement")

        # Pull out the guide star ID and the guider number
        gs_id = [x for x in gs.iterchildren() if x.tag.split(namespace_tag)[1] == "GuideStar"][0].text
        guider = [x for x in gs.iterchildren() if x.tag.split(namespace_tag)[1] == "Guider"][0].text

        # Account for if the guider is written as "guider 1" or "guider1"
        if not isinstance(guider, int):
            guider = guider.lower().replace(' ', '').split('guider')[1]

        LOGGER.info('Master GUI: APT has been queried and found guide star {} and guider {}'.format(gs_id, guider))
`

mirage/psf/segment_psfs.py
```diff
@@ -170,8 +183,11 @@ def generate_segment_psfs(ote, segment_tilts, out_dir, filters=['F212N', 'F480M'
 
     # Define the detector list to loop through
     if detectors == 'all':
-        detectors = ['NRCA1', 'NRCA2', 'NRCA3', 'NRCA4', 'NRCA5',
-                     'NRCB1', 'NRCB2', 'NRCB3', 'NRCB4', 'NRCB5',]
+        if instrument.upper()=='NIRCAM':
+            detectors = ['NRCA1', 'NRCA2', 'NRCA3', 'NRCA4', 'NRCA5',
+                         'NRCB1', 'NRCB2', 'NRCB3', 'NRCB4', 'NRCB5',]
+        if instrument.upper()=='FGS':
+            detectors = ['FGS1', 'FGS2']
```
kjbrooks, line 190: I think this can just simply be: `inst.detector_list`, right?

bhilbert4:
Looks like there is a failing test in test_apt.py, as it is parsing tests/test_data/NIRCam/NIRCamTest.xml. It looks like it is not finding the FGS detector number in there. It could very well be that the format of the xml file has changed since this test xml was created. Can you confirm?

kjbrooks:
This is likely my fault since I added in the ability to grab the FGS detector from special requirements. It should be backwards compatible so it shouldn't be failing so I probably messed up somewhere. I will look into it.

kjbrooks:
Looking into this, if this test passed previously, then it looks like the test xml file is, in fact, different. If you weren't getting an error from `apt.read_apt_xml.get_guider_number` then the NIRCam template previously included the FgsExternalCalibration template, which is no longer does. @mperrin do you have any thoughts on why this file would have changed?

bhilbert4:
Unfortunately, the xml file format changes without warning pretty often. This is why some of the simulator developers have asked for a new, stable output file. Until then, we have to make adjustments when the format changes.

kjbrooks:
Okay, would you then recommend changing the test or changing the `get_guider_number` function so it will run even if there is no guider information in the APT file?

bhilbert4:
I'm not familiar enough with the xml format of the GA template to know offhand what would be easier. Was the guider number present somewhere else in the file previously? Or not there at all? 

It seems like updating the test (maybe just open the xml in the latest version of APT and resave/export?), would be fairly easy, with the only side effect that it will force people to use the latest version of APT and possibly convert their older xml files?

kjbrooks:
Looking through the xml file, while there is no FgsExternalCalibration template in this xml file, there is a special requirement with a guide star and guider specified for all but two of the observations. With the code that I put in, that means that it will find an FGS detector most of the time but not those cases. 

The problem I am having here is that I don't understand how this xml file is being created in the first place. This does not depend on GA, right? It looks like it should be running all the templates which would mean that the FGSExternalCalibration template would be included. I am still getting familiarized with Mirage so I apologize if these are silly questions.

mirage/apt/read_apt_xml.py
```diff
@@ -987,16 +986,39 @@ def read_globalalignment_template(self, template, template_name, obs, prop_param
...
+                # This is an FGS image as part of GA
+
+                # Add FGS exposure to the dictionary
+                tup_to_add = (pi_name, prop_id, prop_title, prop_category,
+                              science_category, typeflag, 'N/A', fgs_subarr, pdithtype,
+                              pdither, sdithtype, sdither, 'N/A', 'N/A',
+                              'FGSRAPID', fgs_grps, fgs_ints, 'N/A',
+                              'N/A', 'N/A', coordparallel,
+                              i_obs, j + 1, template_name, 'FGS', obs_label,
+                              target_name, tracking)
```
kjbrooks, line 1011: Based on my sleuthing, I have come to the conclusion that previously, there were no FGS  observations in the GA template but now that there are, (due to this section?), yam.create_inputs in the test means that for the FGS observations the code enters the elif statement on line 551 of apt_inputs which then tries to find the guider information. The problem is that this APT file does not have any information about which guider to use. 

These are the options as I see them: A) we can add in a guide star special requirement for the GA observations with FGS as the instrument being used, or B) we can try to make it skip this or put in a dummy guider. 

To me, the option A is the best since it will actually be testing the code, specifically the changes that I made. 

The problem: I do not know how this file is generated and where to make the necessary changes. @mperrin or @bhilbert4, do either of you know how this file gets generated and if it's possible to edit it?

bhilbert4: Is it possible to open the xml file in APT? If so, I think that then re-exporting it (after adding a guide star special requirement if necessary) will probably put it in the current format. 

Will GA observations always have a guide star special requirement? If not then the new code should probably be modified to handle that.

kjbrooks: I was just able to open it so I can try to edit it. 

Yes, the GA observations must have a guide star special requirement.

kjbrooks:
I have updated the NIRCamTest.xml file. Tests pass locally. 🤞

bhilbert4:
Tests passed on linux machine...

bhilbert4:
Now that tests are passing, is this ready to merge?

mperrin:
I think so, yes thanks!

PR #492, opened 2020-05-29
Global alignment pointing file updates

bhilbert4:
This PR makes an adjustment to the exposure dictionary in the yaml_generator for global alignment observations. All exposures in the APT pointing file are listed under the NRCA3_FULL aperture. For the purposes of simulating the data, the FGS exposures need to be switched to the appropriate FGS exposure, and RA, Dec values updated. 

This PR relies on updates in #488. It requires the name of the GA_iteration to be present in the exposure dictionary, as well as the number of the FGS detector used.

- [x] Update aperture names
- [x] Update pointing information


mirage/utils/constants.py
```diff
@@ -62,6 +62,16 @@
 # to when it is not in the beam
 NIRISS_GRISM_THROUGHPUT_FACTOR = 0.8
 
+# Lists of the NIRCam/FGS exposure order associated with each type of
```
mperrin, line 65: Note, typo in spelling of 'ORDER' ("OREDER"). 
A simplification is possible here, in that the order of exposures is always the same; just smaller subsets are executed depending on which flavor of GA observation. In particular, the "CORRECT+ADJUST" and "ADJUST2" types (which are identical) are the full set of 7. Everything else is just a truncation of that.

mperrin:
@shanosborne  - any chance you might be able to look over the SIAF pointing conversions between NIRCam and FGS in this MIRAGE PR?  This is for updating the MIRAGE parameter yaml files for global alignment so that they correctly show the pointing of the FGS at the guide star, relative to the NIRCam observation that actually defines the pointing.

bhilbert4:
See line 924 in apt_inputs.py for the pointing changes.

shanosborne:
So I don't completely understand the context of what you're trying to do in this PR. But looking at line 924-926 it looks like you are trying to get the coordinates of NIRCam A3's V2Ref and V3Ref points in the FGS ideal frame (for any of the fgs apertures). So you get the "center" of NIRCam from FGS's perspective.

I plotted up what you're doing, where the ideal values you're saving out `basex, basey` are the blue dot on the right panel (with values (151, 172)). 

So I can't really say if that's what you want, but I can confirm that that's what you're doing. 

image

bhilbert4:
Thanks @shanosborne. We're basically trying to manually adjust the information in an APT pointing file. The file shows where NRCA3 is pointing, and from that we are calculating where FGS is pointing. Based on your comment and @mperrin's testing, this does seem to be doing the right thing.

PR #493, opened 2020-06-02
Save NIRISS POM seed image

bhilbert4:
This PR adds code to save the oversized POM seed image in NIRISS "pom" observations as a separate fits file. The POM image is 2322 x 2322 pixels, and covers the entire pick off mirror area. The seed image is saved in a *_pom_seed_image.fits file, and the array is exposed as self.pom_seed (and self.pom_segmap)

In addition this PR corrects a bug where point sources were not showing up in the expanded area of the array (beyond the central 2048x2048 detector pixels). Galaxy and extended sources were correctly showing up in this region, but point sources were being cropped/removed if they fell off the detector itself. The same behavior was seen in NIRCam and NIRISS WFSS simulations, which also use the extended area for the direct image that is passed to the disperser. This has now been corrected.

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -380,9 +380,17 @@ class Catalog_seed():
 
             # For NIRISS POM data, extract the central 2048x2048
             if self.params['Inst']['mode'] in ["pom"]:
+                # Expose the full-sized pom seed image
+                self.pom_seed = copy.deepcopy(self.seedimage)
+                self.pom_segmap = copy.deepcopy(self.seed_segmap)
+
+                # Save the full-sized pom seed image to a file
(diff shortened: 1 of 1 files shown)


mirage/seed_image/catalog_seed_image.py
```diff
@@ -2457,7 +2465,7 @@ def make_point_source_image(self, pointSources, segment_number=None, ptsrc_segma
 
             scaled_psf, min_x, min_y, wings_added = self.create_psf_stamp(
                 entry['pixelx'], entry['pixely'], psf_x_dim, psf_y_dim,
-                segment_number=segment_number
+                segment_number=segment_number, ignore_detector=True
```
bhilbert4, line 2468: This is the key fix to get the point sources to appear in the extended area around the detector.

PR #494, opened 2020-06-02
Add convenience functions for computing PA V3 values for roll_angle input

mperrin:
Adds functions to compute nominal V3 position angle for a given target on a given date. These rely on the [JWST GTVT](https://github.com/spacetelescope/jwst_gtvt) to calculate the nominal roll angle on some given date. 

Documentation updated to describe these functions and show example code.  

Thanks to @mfixstsci for help with the GTVT, including some tweaks to that codebase to enable this usage.

### DIFF
--- a/docs/yaml_generator.rst
+++ b/docs/yaml_generator.rst
@@ -208,19 +208,35 @@ In order to use the Mirage default value (roll angle = 0), simply do not provide
 
 ::
 
-    pav3 = None
+    roll_angle = None
 
-To specify a single roll angle to be used in all observations, supply a single number.
+To specify a single roll angle to be used in all observations, supply a single number in decimal degrees.
 
 ::
 
-    pav3 = 34.5
+    roll_angle = 34.5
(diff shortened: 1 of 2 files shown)


mperrin:
I think it doesn’t need to be a required dependence, since it’s only needed for this one function (and it’s optional to use that function too). 

But that said, it’s a simple install via either conda or pip so not a big deal to add as a dependency if you prefer.

bhilbert4:
Makes sense to me. Maybe just add a note in the documentation where it's called that you'll need to install jwst_gtvt first?

mperrin:
Yes that was my intent with this sentence I added in the docs:
> These rely on the `JWST general target visibility tool <https://github.com/spacetelescope/jwst_gtvt>`_, which you must have installed to use this feature.

Is that not clear enough? I can rephrase if you have any suggestions.

bhilbert4:
That looks good to me.

PR #495, opened 2020-06-03
[WIP]: Enable weak lenses in catalog creation and simulations

bhilbert4:
This PR adds code to allow for weak lens+filter simulations, specifically by updating the catalog creation functions to interpolate correctly when working from WISE/GAIA/2MASS query results.


mperrin:
@bhilbert4 - I know you're doing this primarily motivated by the TSO observing modes but this is also going to be useful for some of our WFSC simulations, so I'm definitely 👍 for this. 

(And I was going to suggest having the WLM8 lens also just use the same columns as WLP8 but then I actually read the PR and saw you're already doing that. :-)

bhilbert4:
Putting together a list of places in the code that may be affected by the addition of weak lenses:

- galaxy_background within create_catalog.py cannot use the weak lenses until config/goodss_3dhst.v4.1.jwst_galfit.cat is updated to contain them.

- get_all_catalogs makes the magnitude column names and then simply returns them. Input filter names are put into a dictionary though, which is passed into johnson_catalog_to_mirage_catalog
filter names are also passed into combine_and_interpolate

- johnson_catalog_to_mirage_catalog calls make_filter_names to make magnitude column names. These are passed into transform_johnson_to_jwst. There is also a line in here that contains filt.split('_'). This may need to change to support the weak lenses.

- In transform_johnson_to_jwst, filter names are passed to crossmatch_filter_names

- add_magnitude_column in the catalog generator also assumes a filter name, so this may need to be updated. Would continue to work as-is if the input filter name is something like 'wlp8_f090w'
Also filter_check.

- catalog_seed_image: need to grab the correct zeropoint from the zeropoint file. Need to modify select_magnitude_column, which currently grabs the filter name from the yaml file and constructs the magnitude column names from that only.

bhilbert4:
self.usefilt in catalog_seed_image needs to be rethought as well.

bhilbert4:
- The zeropoints file needs to have the weak lens entries added, and fluxcal_info needs to be updated to find the new entries.

bhilbert4:
Plan to move forward:
Updates are split into two broad categories:
1. catalog creation updates
2. seed image creation updates

1. Catalog creation updates:
Top-level change in get_all_catalogs.
User needs to input a list of filters with which to populate the catalogs. Currently that list is (e.g. ["F090W", "F323N"]). For filters in the pupil wheel, like F323N, the filter wheel filter is implied (e.g. F322W2). We could keep this behavior as the default, but also allow the user to input a more complex string like "F322W2/F323N" or "WLP8/F090W". In this case the code will split the string into pupil and filter values, and create catalog column names such as "nircam_wlp8_f090w_magnitude". This way the user can specify weak lenses, as well as non-standard combinations (like F323N/F356W, which is used for WFSC).

2. Seed image creation updates
Read in catalog column names. Currently the code supports only a single optical element in the name ("nircam_f323n_magnitude"), and the pupil value is assumed. Keep this default, but also allow a case where there are two optical elements in the name. Along with this, the zeropoint file will need to be updated.

bhilbert4:
Updated galaxy_background for the new filter/pupil format. New entries in our 3D HST catalog for the weak lenses can be done in a separate PR. For now, it will return an exception if you ask for a galaxy sample using the weak lenses.

Issue #483, opened 2020-05-19
Mirage crashes for large volume imaging mode sims


bhilbert4:
First reported by Eliot Malumuth, the observation_generator fails in the case of an imaging mode simulation with lots of groups. He's working on PID 1067, which in observation 357 has a full frame RAPID exposure with 20 integrations * 3 groups per integration. This puts the output file at a size above the file splitting limit.

The problem in this case is related to the file splitting work that was inserted as part of the TSO mode simulations. When running in TSO mode with a large simulation, the seed image is split into multiple segments and the list of filenames containing the segments is passed to the observation generator. This is because the seed image in this case is a 4D array of signals, rather than signal rates. This allows a source with a signal that is varying in time. Similarly, dark_prep creates multiple dark current segment files. The two sets of files are then combined in the observation_generator.

However, in imaging mode, the seed image is a single 2048x2048 (for full frame) array that is passed directly to the observation generator. So in the case of an imaging mode observation that is above the file splitting limit, the observation generator gets confused because the seed image is an array, but in terms of file splitting it is assumed to be a list of files, and the dark input is also a list of files, etc. 

So, the observation generator needs to be updated to support the file splitting for all modes.

bhilbert4:
File splitting for imaging/WFSS modes enabled with #506. File-splitting is still not in place for non-sidereal simulations, nor sidereal simulations with moving targets. Code changes for these will be more invasive, so I'm moving them to a new issue in order to get the changes for sidereal simulations in place.

PR #506, opened 2020-06-25
Add file splitting to imaging mode data

bhilbert4:
This PR adds file splitting for imaging mode observations. File splitting was already implemented for imaging TSO and grism TSO observations, but it is also possible for other modes to have observations that are large enough to trigger file splitting. This has been seen in several WFSC commissioning programs. 

These changes allow for file splitting in imaging data with sidereal tracking. 

- [x] Test updates with imaging observations that are both over and under the splitting limit

- [x] Test with WFSS observations both over and under the limit.

- [x] Test TSO imaging mode to make sure these changes haven't caused any errors

- [x] Test TSO grism mode to make sure these changes haven't caused any errors

These changes will not affect non-sidereal observations, nor sidereal observations that contain non-sidereal targets (e.g. background asteroids). Implementing file splitting for those modes will be much more complicated and invasive, and at the moment is lower priority.

### DIFF
--- a/mirage/dark/dark_prep.py
+++ b/mirage/dark/dark_prep.py
@@ -391,30 +391,6 @@ class DarkPrep():
         for ref in rlist:
             self.ref_check(ref)
 
-    def file_splits(self):
-        """Check to see whether the final dark file will need to be
-        split into segments due to large file size
-        """
-        # Let's declare the equivalent of a 100 group full frame ramp
-        # as the upper limit for observation file size
-        pixel_limit = 2048 * 2048 * 100
-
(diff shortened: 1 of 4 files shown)


bhilbert4:
Ugh. The failing test is the Gaia query with an error code of 500 (internal server error). Maybe we'll just wait a bit on this and see if whatever that problem is goes away, since I don't think it's related to Mirage at all.

bhilbert4:
Resolves #483 
I'll create a separate issue for adding file splitting to non-sidereal and moving target data.

Issue #472, opened 2020-04-06
Mirage seed image failed for faint objects


mtakahiro:
When I run the following,

```
test_yaml_files = 'yamls/new_focus_field_f150w_ghost.yaml'

imseeds = []
pfile = test_yaml_files
print('Running catalog_seed_image for {}'.format(pfile))
cat = catalog_seed_image.Catalog_seed(offline=False)
cat.paramfile = pfile

cat.make_seed()
imseeds.append(cat.seed_file)
print(cat.seed_file)

```

Mirage crashed and returned this error message; 

```
AttributeError                            Traceback (most recent call last)
<ipython-input-20-16b606c09bf3> in <module>
----> 1 cat.make_seed()
      2 imseeds.append(cat.seed_file)
      3 print(cat.seed_file)

~/anaconda3/envs/mirage/lib/python3.6/site-packages/mirage-1.1.1.dev38+gcfbe662d-py3.6.egg/mirage/seed_image/catalog_seed_image.py in make_seed(self)
    221         if self.params['Telescope']['tracking'].lower() != 'non-sidereal':
    222             print('Creating signal rate image of synthetic inputs.')
--> 223             self.seedimage, self.seed_segmap = self.create_sidereal_image()
    224             outapp = ''
    225

~/anaconda3/envs/mirage/lib/python3.6/site-packages/mirage-1.1.1.dev38+gcfbe662d-py3.6.egg/mirage/seed_image/catalog_seed_image.py in create_sidereal_image(self)
   1553
   1554             # translate the extended source list into an image
-> 1555             extimage, ext_segmap = self.make_extended_source_image(extlist, extstamps)
   1556
   1557             # Multiply by the pixel area map

~/anaconda3/envs/mirage/lib/python3.6/site-packages/mirage-1.1.1.dev38+gcfbe662d-py3.6.egg/mirage/seed_image/catalog_seed_image.py in make_extended_source_image(self, extSources, extStamps)
   3486                 # Make segmentation map
   3487                 indseg = self.seg_from_photutils(stamp[l1:l2, k1:k2] * entry['countrate_e/s'],
-> 3488                                                  entry['index'], noiseval)
   3489                 segmentation.segmap[j1:j2, i1:i2] += indseg
   3490         return extimage, segmentation.segmap

[...]

PR #507, opened 2020-06-26
Make segmentation map thresholds a user input

bhilbert4:
This PR updates the threshold value that is used when adding sources to the segmentation map. Previously the threshold value was based on a guess of some fraction of the noise in the observation. However, this didn't take into account observation programs with the ultimate goal of combining N exposures in the level 3 data. Instead of the one-size-fits-all approach, this update allows the threshold value as a user input in the yaml files.

The default value is 0.031 ADU/sec, as determined by some experimentation by Nor with a very deep observation program example. If the new threshold keywords are not present in the yaml file, Mirage will use this default value. Otherwise, users can provide their own value, along with the units that value is in. Allowed units are: ADU/sec, e/sec, MJy/sr, FLAM, and FNU. All are converted to ADU/sec upon entry.

### DIFF
--- a/docs/example_yaml.rst
+++ b/docs/example_yaml.rst
@@ -44,6 +44,7 @@ Below is an example yaml input file for *Mirage*. The yaml file used as the prim
 	  pixelflat_: None
 	  illumflat_: None           #Illumination flat field file
 	  astrometric_: crds         #Astrometric distortion file (asdf)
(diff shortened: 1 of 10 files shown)


bhilbert4:
Resolves #472

bhilbert4:
Hmmm. Somehow this has the file splitting PR commits (#506 ) in it. Not sure how I managed to do that.

PR #510, opened 2020-06-30
Remove Grism files from reference file collection

bhilbert4:
Grism-related reference files are stored in separate GitHub repositories. Currently there is a copy of these reference files stored in Box as part of the Mirage reference file collection. This is easy for users, but there is a danger of the two copies of the grism reference files becoming out of sync when updates are made. 

Therefore, this PR removes the grism reference files from the Mirage reference file collection, removes them from the downloader, and updates the reference file installation instructions to have users install the grism reference files directly from the other GitHub repositories.

https://github.com/npirzkal/GRISM_NIRCAM
https://github.com/npirzkal/GRISM_NIRISS


bhilbert4:
The motivation behind this is that the NIRISS Grism reference files were recently updated to change the orders from A, B, C to 0, +1, +2. The disperser software was also changed to expect the new order names.

bhilbert4:
Resolves #510

Issue #511, opened 2020-07-03
TSO mode requires background source catalog


bhilbert4:
Grism TSO mode in Mirage appears at the moment to require having a background source on the detector. If there are no background sources that land within the aperture, or if the point source catalog is set to None, then Mirage crashes. This requirement should be made optional.

bhilbert4:
Fixed in #512

PR #512, opened 2020-07-06
Make background sources optional in grism TSO simulations

bhilbert4:
When creating a grism TSO simulation, Mirage currently fails if there is no background source catalog (e.g. point source, galaxy, extended) listed in the input yaml file. It also fails if none of the background sources from those catalogs are within the requested aperture.

This PR makes updates so that the background catalogs are not required, and if no sources within those catalogs happen to be in the FOV, then Mirage will handle that situation.


bhilbert4:
Resolves #511

Issue #484, opened 2020-05-19
Correct scaling of extra large galaxy stamp images


bhilbert4:
There is currently a bug in the catalog seed generator for galaxies that are larger than the field of view of the detector. Currently the largest galaxy stamp image Mirage will create is 2048x2048. It creates an instance of the desired galaxy using astropy's Sersic2D class, and the galaxy itself is created using the model and the grid of pixels for the stamp image. This stamp is properly scaled for the given parameters. However, when Mirage then scales the galaxy to the requested magnitude, it normalizes the stamp to have a total signal of 1, and then multiplies by the total signal corresponding to the requested magnitude. This is a problem if a significant fraction of the galaxy falls outside the stamp, because the stamp will contain the total signal corresponding to the requested magnitude when if fact it should contain that value minus whatever fraction is associated with the part of the galaxy that is outside the stamp. In other words, the galaxy will be scaled to be too bright. 

I believe the only case where this is a problem is for galaxies whose stamp image should be larger than 2048x2048. Mirage currently sets the stamp image length as the smaller of 100 * the half-light radius or 2048 pixels. The galaxy is then created and scaled, and the stamp image is cropped in order to contain 99.95% of the signal. This was done in order to prevent even small galaxies (with large Sersic indexes) from having extra large arrays that were slowing down the calculations. I believe that as long as the stamp image is 2048x2048 or less, then the stamp will contain 99.95% of the expected signal, and the scaling will be correct.

@dancoe came up with a recipe for dealing with these large/bright galaxies whose stamps may be larger than the detector. I'll play with the notebook he gave me to see if the solution can be easily integrated and give better results.

PR #516, opened 2020-07-14
Galaxy stamp scaling and sizing

bhilbert4:
This PR updates the way galaxy stamp images are created and scaled. The goal is to make Mirage faster, and to correct for slightly incorrect signal scaling in the cases of stamps with a very large area.


bhilbert4:
Resolves #484

bhilbert4:
Tests are showing this new method of scaling and sizing galaxy stamp images cuts computation time roughly in half from what it was. With these new changes, Mirage makes use of the equation governing the amount of flux within a given radius for a Sersic profile to estimate the stamp size needed to contain an acceptable amount of flux (currently set to 99.95% of the total object signal in the source catalog). Previously, Mirage was blindly making very large stamps, as a way to be sure they contained enough signal. With this change the stamp sizes decrease, speeding up execution time. In addition, in many cases the Sersic profile is correctly scaled upon creation. In some cases (high Sersic index, high ellipticity, galaxy center close to the center of the pixel), Mirage does have to rescale the stamp, as the scaling done by Sersic2D is incorrect. As described in an email from Larry Bradley:

"When astropy models are discretized/evaluated, the values that are returned represent the point samples of the model at the input pixel locations.  In general this does not preserve total model flux (it doesn’t take into account how the model changes within a pixel bin).  To exactly preserve total model flux, one would need to integrate the model over the pixel bins.  The difference is thinking of a pixel as a point sample (i.e., at the center of the pixel) vs. thinking of a pixel as a bin."

In this case, since Mirage has previously calculated the stamp size such that it contains the required percentage of total signal, rather than integrating the model over the pixel bins, we simply rescale the 2D Sersic profile to match the total signal given in the source catalog.

PR #517, opened 2020-07-15
[WIP]: Grism TSO simulation: make sure 2D background is added in all cases

bhilbert4:
This fixes a problem in the logic of #511 . With that update, the 2D dispersed background was not being added to the simulation in cases where there were no "background" (i.e. non-TSO) sources. This fixes that so that the background is added regardless of whether there are sources besides the TSO source.


bhilbert4:
Needs to work for:
- [ ] Case where no background source catalogs are given
- [ ] Case where background source catalog is present, but no background sources are on the detector
- [ ] Case where background source catalogs are given and there are sources present on the detector

PR #523, opened 2020-07-20
Separate flats and transmission files

### LINKED ISSUE #521: Mirage is multiplying seed images by the flat field twice
@micaelabagley has found that Mirage multiplies simulations by the flat field reference file twice.

Within `catalog_seed_image.py`, the various pieces of the seed image are multiplied by the flat field before being combined into the final seed image. But then in `obs_generator.py` the flat field is multiplied again, just after using the seed image to create the exposure with all its integrations and groups.

So which one should we remove? I think we don't want to multiply the seed image by the flat. The idea of the seed image in general is that it doesn't contain detector effects (except for distortion, which needs to be included). For WFSS simulations, the seed image is provided to the disperser. The flat should be applied after dispersing since it is the dispersed light that hits the detector and is subject to the pixel-to-pixel sensitivity variations.

Is there any reason we want/need the flat applied in the seed image?

https://github.com/spacetelescope/mirage/blob/master/mirage/ramp_generator/obs_generator.py#L1239

bhilbert4:
This PR separates flat fields and the newly-defined transmission files. Previously, for WFSS simulations, NIRISS had to have a flat field applied to the data prior to dispersing because the flat field file contained transmission information (e.g. pixels covered by the occulters were set to zero.) However this is not the correct way to add flat field effects to the data. The flat field effects should be added after dispersing, while the transmission information should be added prior to dispersing. By separating the two, everything can be done correctly. 

Previously for NIRCam WFSS simulations, the flat field was not applied until after dispersion. However, the flat that was being used was the placeholder grism flat in CRDS. This flat has all pixels set to 1.0, because during data reduction/calibration, the data are not flat fielded until you have a 1D extracted spectrum, but the pipeline needs a flat field file to divide into the data prior to that point. However, using the placeholder file means that flat field effects were not being added to the WFSS simulations. This PR switches to using the imaging mode flats to add the flat field effects. This assumes no wavelength dependence on the flats, which is ok for now. 

Finally, for simulations with all modes, the flat field was being applied multiple times. This PR fixes that, so that the flat is only applied once.

- [x] Separate transmission info from flats
- [x] For NIRCam, using imaging mode flats for WFSS sims. For NIRISS, use grism-specific flats
- [x] Multiply transmission file to each seed image, do not multiply flats
- [x] Test to be sure CRDS returns the expected flats for nircam and niriss, wfss and imaging
- [x] Multiply flats within observation generation only
- [x] Check to be sure flats are not applied in wfss_simulator nor grism_tso_simulator
- [x] Check that a transmission file with an odd number of rows and/or columns still gives the correct coordinate/location results wherever self.grism_direct_factor is used 

  
Resolves #521


bhilbert4:
- [ ] Add transmission entry to yaml file generator
- [ ] In the near term, transmission files will be retrieved from GRISM_NIRCAM and GRISM_NIRISS. Adapt code for that
In the longer term, the transmission files will most likely become CRDS reference files, and will be retrieved by querying CRDS, as with the other reference files. This work will take longer than the rest though.

bhilbert4:
Currently the direct seed image looks good (still need to test one with odd number of rows/cols), but the dispersed seed image is wrong. Need to update header info/values that are passed to the disperser.

bhilbert4:
Resolves #521

PR #531, opened 2020-08-04
APT/yaml_generator adjustments for moving targets

bhilbert4:
This PR contains small updates such that the `tracking` parameter in the yaml files will be set to `non-sidereal` for observations with non-sidereal targets.


bhilbert4:
Tested with non-sidereal and sidereal targets and everything appears correct. Currently this is implemented only within `read_generic_imaging_template()`. 

- [ ] Add tracking search to other `read` functions

PR #546, opened 2020-08-25
Normalize segment PSFs correctly based on fractional area per segment

mperrin:
Segment PSFs should be normalized not to near 1.0 total intensity, but based on the area of the primary in each segment.

This is basically 1/18 = 0.055, but varies slightly per each segment based on obscurations and projected area of A/B/C segment types. I went ahead and did it the more rigorous way. Varies from 0.051 to 0.057. 

Code is not yet tested at all, but I think and hope this may fix our mystery flux problems for OTE sims.


Skyhawk172:
Looks good to me

mperrin:
Tested, output PSFs are normalized as expected. Looks good and is approved so I will merge.

PR #549, opened 2020-09-01
Allow user to disable file splitting in dark_prep

bhilbert4:
This PR adds a keyword to dark_prep() that allows the user to disable the file-splitting. This is helpful for the creation of linearized dark current files that can be added to the collection of Mirage reference files. Ground testing darks have 108 frames, and these need to be in a single file for Mirage to use them in the dark_prep step.

The default for this new `file_splitting` keyword is True, so this change is transparent as far as most users are concerned.

### DIFF
--- a/mirage/dark/dark_prep.py
+++ b/mirage/dark/dark_prep.py
@@ -48,16 +48,25 @@ INST_LIST = ['nircam', 'niriss', 'fgs']
 
 
 class DarkPrep():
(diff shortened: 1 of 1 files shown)


bhilbert4:
I'm going to ignore the failing documentation, since that isn't specific to these changes, and because I want to try creating some new darks on a different machine with this code, and merge this now.

PR #551, opened 2020-09-02
Dark_prep output adjustments

bhilbert4:
These code changes clean up a few issues with the outputs from dark_prep which were found while using it to create more linearized darks to add to the collection of Mirage reference files.

First, we add the names of the CRDS reference files used when linearizing the dark data. These are saved in the very basic header of the dark_prep object file.

I've also stopped saving the linear dark ramp that is output directly from the pipeline. Originally the idea was that saving this file would allow future runs of Mirage with newly linearized darks. However these outputs are not in the format expected by Mirage (e.g. the superbias+refpix signal needs to be saved in an extension), so the files are not actually all that useful. Instead, everything is repackaged into the dark_prep_object file, which can be used in future Mirage runs. Turning off the pipeline output also cuts the data volume of saved darks in half, which is nice when working with 108-group 3.4GB files.

### DIFF
--- a/mirage/dark/dark_prep.py
+++ b/mirage/dark/dark_prep.py
@@ -495,6 +495,26 @@ class DarkPrep():
             # Don't worry about counting reset frames between integrations
             # We're not concerned with timing here.
 
+    def get_reffile_metadata(self, keyword):
+        """Reetrive the name of the reference file used to calibate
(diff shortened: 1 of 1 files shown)


bhilbert4:
Ignoring documentation failure for the moment.

PR #552, opened 2020-09-04
Add normalization check for gridded PSF library

### LINKED ISSUE #358: Add normalization check on gridded PSF library
Related to #356, it would be nice to have a check on the total signal in a gridded PSF library when it is opened. Mirage assumes that the total signal in a PSF is normalized to the square of the oversampling factor. 

This would have to be a rough check, as the total signal in the PSF image will depend on the size of the image.

bhilbert4:
Resolves #358 

This PR adds a simple normalization check to the gridded PSF library read in by Mirage. This will help to identify any improperly normalized libraries. The check expects that the total signal in a PSF divided by the square of the oversampling factor should be roughly 1.0. 

I currently have the lower and upper limits for a "good" normalization set at 0.8 and 1.0. @mperrin does that make sense? I'm wondering if having a lower limit makes sense at all.

Also, the libraries for ITM images will never fail this test, as the function that reads in the ITM library file currently manually normalizes the total signal to 1.0.


Skyhawk172:
This seems to be causing issues with some NIRISS and FGS PSF grids that were recently generated using the Photutils normalization method (as opposed to strictly normalizing to 1.0).  For example, for some NIRISS PSF grids, the total signal in the PSF is a fraction of a percent above 1.0 (e.g. 1.0011332952694405) and the code catches it and halts because it exceeds 1.0.   It appears the lower limit affects FGS grids as well.   So the limits should be revised and made consistent with the way WebbPSF normalizes its PSF grids.

mperrin:
See now https://github.com/spacetelescope/webbpsf/issues/487

PR #555, opened 2020-09-14
Slowly-moving non-sidereal sources

bhilbert4:
This PR contains a couple of small fixes to address the case of non-sidereal targets that are moving slowly (<1 "/hour). Due to the way that Mirage subsamples the output frame and PSF, if a target was moving slowly enough that it didn't move from one sub-pixel to another over the course of a frame, Mirage was not adding the source at all. This PR fixes that issue. At the same time, it addresses a source scaling error that was affecting all non-sidereal cases, but was especially visible for slow-moving sources. The sources were being scaled too bright.


bhilbert4:
Merging this now. Further updates, such as allowing the use of an ephemeris file, will be handled in a different PR.

PR #572, opened 2020-10-21
WFSC: Partial support for OTE Coarse Phasing sims using DHS

mperrin:
This PR adds just enough basic support for the DHS grisms (GDHS0, GDHS60) to enable running a MIRAGE sim without errors, for a program using the OTE WFSC Coarse Phasing simulation.  

Note, WebbPSF does not actually have any sort of model for the DHS. This is intended only for use with an externally-generated DHS simulation (i.e. from Ball's ITM tool), supplied as an "extended object" catalog.

In particular, we don't actually use the PSFs for anything in this mode - we just drop in the ITM image, without any further PSF convolution - but we have to at least load enough of a PSF that MIRAGE does not error out. This requires adding the DHS pupil values in a few places, and adding a workaround to just load F150W2+Clear PSFs (since there's no way to make a F150W2+GDHS PSF with WebbPSF) and photometric calibration; the PSF is completely inappropriate in this case, but it's enough to let MIRAGE run. 

FYI @obi-wan76 @Skyhawk172


bhilbert4:
I'm in favor of keeping the rotation even if the input position is in units of pixels. I'm thinking of a case of someone with a little stamp image of a galaxy who wants it to be at some specific orientation that's different than what's in the stamp image file. What if instead we allow for a value of None in the position angle column, in which case no rotation is done?

bhilbert4:
I think the file parsing should be ok. There are lots of other fields that can be set to None, and those lines aren't read in in any special way.

bhilbert4 (approved):
Looks good to me! There is another call to rotate_extended_image() on line 1468. Could you also add the same condition to that call?

mperrin:
>There is another call to rotate_extended_image() on line 1468. Could you also add the same condition to that call?

Slight variation: Rather than adding a second copy of the check, I moved that check for None into `rotate_extended_image()` itself, so that it applies for any call to that function. 

I also added a brief note to the documentation that setting pos_angle = None means no rotation.

mperrin:
Hi @bhilbert4 - this isn't particularly urgent, but I think this PR is ready to go. Tests passed, you approved it before with one requested change and I have implemented that. Should I go ahead and merge myself, or would you prefer to? Thanks!

bhilbert4:
Done and done.

PR #579, opened 2020-11-05
Add option for downloading a single dark for each detector

### LINKED ISSUE #562: Option to fetch fewer darks for simple testing
Feature request: provide an option in ``downloader`` to fetch only a limited set of dark reference files rather than forcing the user to download  hundreds of GB of reference files just to get a test up and running!

Thanks!

bhilbert4:
This PR adds an option to the reference file downloader to download only a single dark current file per detector. This is intended for those who want a quick start of sorts, to test out Mirage, without downloading all of the reference files. This is done via the added `single_dark` keyword in the downloader script.

Resolves #562


gbrammer:
Thanks, @bhilbert4 !

PR #590, opened 2020-11-23
Update yaml_generator to properly handle non-sidereal sources

### LINKED ISSUE #588: Update yaml generator to work with non-sidereal targets
Another update related to what was done in #564. The yaml_generator still does not work with non-sidereal targets. 

Need to make sure it works with them, most notably in getting the pointing RA and Dec correct for multiple exposures (i.e. at different times) and also multiple detectors. This may be through the use of an ephemeris file or not. Either way, we'll have to read in the source catalogs in order to get velocities or ephemeris file names (although the latter may be available through the APT xml file).

bhilbert4:
This PR allows the yaml generator to calculate and use the correct pointings for non-sidereal observations, based on an ephemeris file or manually input velocity values in the non-sidereal source catalog.

Note that I made a bunch of updates to apt_inputs.py as well, and then accidentally did `git checkout` instead of `git add` on the file, so I lost those changes and need to redo them.

Resolves #588


bhilbert4:
Tests show that this seems to be working well. It has revealed a bug in the pointing values that are calculated using ra_dec_update(), but the bug appears to affect all modes (not just non-sidereal) and is not caused by these changes, but instead is already present in the master branch. I'll address that in a separate issue and PR.

bhilbert4:
Nevermind about the ra_dec_update bug. Turns out APT was using a different version of the PRD compared to what pysiaf was using. This was causing sources to be shifted from their expected locations.

PR #595, opened 2020-12-11
WFSC: Enable SAM for full-pupil images

Skyhawk172:
This PR enables boresight offsets for full-pupil images, as opposed to single-segment images.  Similarly to the single-segment case, it relies on header keyword information to shift the all the sources.

This will support Coarse MIMF and Global Alignment activities, in particular.

I just want to point out that this change should not impact anything other than WFSC-related activities, BUT all MIRAGE simulations will be performing the checks for (1) the existence of the properly-named input image file and, if it exists, (2) the existence of the BSOFF_V2 and BSOFF_V3 header keywords.

### DIFF
--- a/mirage/seed_image/catalog_seed_image.py
+++ b/mirage/seed_image/catalog_seed_image.py
@@ -2146,9 +2146,18 @@ class Catalog_seed():
         if self.runStep['pointsource'] is True:
             if not self.expand_catalog_for_segments:
 
+                # CHECK IN INPUT IF THERE'S A BORESIGHT OFFSET TO BE APPLIED:
+                offset_vector = None
+                infile = glob.glob(os.path.join(self.params['simSignals']['psfpath'], "{}_{}_{}*.fits".format(self.params['Inst']['instrument'].lower(), self.detector.lower(), self.psf_filter.lower())))[0]
+                if infile:
+                    header = fits.getheader(infile)                    
+                    if ('BSOFF_V2' in header) and ('BSOFF_V3' in header):
+                        offset_vector = header['BSOFF_V2']*60., header['BSOFF_V3']*60. #convert to arcseconds
+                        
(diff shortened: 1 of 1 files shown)


bhilbert4:
I'm wondering if there's a way to skip the check for non wavefront sensing observations? Would it be possible to do this only for certain exposure types? The way it is now, it will fail for some filters when running with the default PSF library files. That's because the standard filename is "<instrument>_<detector>_<pupil>_<filter>*.fits". So for example, observations using F444W, F405N, F466N, F470N will all use a PSF library file that is "nircam_nrca5_f444w*fits"

Skyhawk172:
I think the second IF statement where we check for the BSOFF_V2/BSOFF_V3 keywords would filter out those other observations that are not wavefront sensing. Currently, the only extra cost is that we need to open the file and check the header every time.

bhilbert4:
Ok, I see. So in the worst case, for a normal imaging exposure, it'll open the PSF library file and not find those keywords. One thought that occurs to me is dealing with failure modes. If the glob comes back with an empty list, then Mirage will crash because it will try to take the 0th element of the empty list. So in that case it will fail with 'IndexError: list index out of range'.

But if you were to break up the current glob statement into two statements (1. call glob, 2. check list length and grab 0th element) you could have it come back with a more descriptive error (e.g. 'No matching PSF libraries found for <instrument>, <detector>, <filter>') It depends on the PSF libraries you are using though. For the standard library collection, I don't think this error will ever be seen, since all filters are covered.

Also, for that logger line you commented out, the logger expects a single string, rather than the tuple that it's being given now. I must have missed that at some point in a recent update. Could you change that line to be:
self.logger.info('    Position angle = {}'.format(position_angle))

Skyhawk172:
Okay, I'm on it. Thanks!

PR #597, opened 2020-12-11
[WIP]: Add optical ghosts to NIRISS simulations

### LINKED ISSUE #592: Add optical ghosts to NIRISS data
@mtakahiro has come up with a method to add optical ghosts that correspond to input point sources to NIRISS simulations. The ghosts' locations and brightnesses depend on the filter being used (and possibly on location on the detector). 

Incorporating this into Mirage shouldn't be too difficult. Based on his notebook, the algorithm would be something like:

Read in the point source catalog and calculate locations on the detector (in x,y), as well as count rate. Provide this information to a new function that calculates the ghost location and count rate. Select the stamp image to use for the ghosts based on filter (and maybe location). Update (or create) and extended source catalog that contains the ghost entries, including the appropriate stamp image file. Since extended sources are handled after point source, it should be possible to do this update on the fly.

For WFSS data, the ghosts should have the same SED as the sources that cause them. In this case, there are several options:

- If the SED is derived from one or more catalog magnitudes through various filters, then we need to calculate the ghost's magnitude in these filters as well (based on the real source's magnitudes).
- If the user provides an hdf5 file with SEDs, then that catalog file will need to be updated as well, copying the SED from the original source
- If the user provides a dictionary containing SEDs, then the dictionary will have to be updated in the same way as the hdf5 file above.

bhilbert4:
This PR uses knowledge from ground testing and code from @mtakahiro to give users the option to add optical ghosts when running NIRISS simulations. 

For each source in the input source catalogs, the new code will calculate the location and brightness of an associated ghost. These ghosts are then treated as extended sources. Mirage makes (and saves) separate extended source catalogs for the ghosts from each input catalog. All extended sources are then combined when creating the extended source seed image.

The stamp image to use for the ghost will be added to the Mirage reference files. There is also an ascii file that describes the location of ghosts. This file is added to the config directory of the repository.

The current implementation adds 2 entries to the input yaml files, in the simSignals block:
add_ghosts: True            
PSFConvolveGhosts: False  

add_ghosts allows ghost creation to be toggled on/off. (Ghosts will only be added to NIRISS sims at the moment)
PSFConvolveGhosts dictates whether the stamp images of the ghosts are convolved with the instrument PSF before adding to the seed image.

- [x] Create ghost catalog from point source inputs
- [x] Create ghost catalog from galaxy source inputs
- [x] Create ghost catalog from extended source inputs
- [x] Create ghosts as extended objects
- [x] Differentiate between convolving normal extended sources with the instrument PSF and convolving the ghost images with the PSF (via input yaml keyword)
- [x] Add ghost creation to moving target section of code
- [x] Update yaml generator to add the new keywords and default values
- [x] Add tests for new code
- [x] Update documentation
- [x] Modify example notebook to show the ghost creation
- [x] Add ghost stamp image to Box and update reference file downloader

Resolves #592


bhilbert4:
Working on the notebook has shown that I need to make sure that ghost sources have the same SED as input sources. This means potentially dealing with the hdf5 file.

bhilbert4:
@mtakahiro has tested this and is happy with the results. I'll merge this and we can handle any issues that pop up in separate PRs.

PR #601, opened 2020-12-22
Fix for py39 and collections.abc

nden:
@bhilbert4 This PR provides a fix for Python 3.9 and collections.abc.

```
DeprecationWarning: Using or importing the ABCs from 'collections' 
instead of from 'collections.abc' is deprecated since Python 3.3, 
and in 3.9 it will stop working
```


bhilbert4:
Looks great! Thanks for the update. I hadn't seen this deprecation warning when running Mirage recently. I must not have been running the right cases.

PR #602, opened 2021-01-11
Minor fix to add the SubpixelDitherType element for WFSC dithers

Skyhawk172:
It looks like APT 2020.5 does not export the "SubpixelDitherType" XML element anymore for the case of WFSC dithers, which is needed by MIRAGE.  This PR simply creates the needed element if the PrimaryDitherType is "WFSC".


bhilbert4:
Tests are failing here because for some templates PrimaryDitherType is not present in observation_dict. Could you put your 2-line fix within this if statement?

`if "PrimaryDitherType" in observation_dict.keys():`

PR #606, opened 2021-01-20
Minor fix when logging jitter

Skyhawk172:
Minor PR to fix a logging error I've encountered. Similar to an earlier one we saw recently.


bhilbert4:
Looks good to me! Good catch.

Issue #612, opened 2021-01-27
Add xml support for coronagraphic template


bhilbert4:
This would be helpful for GTO preparations. I think the best way to go at the moment would be to get it working such that filters/read patterns/pointing information is correct, and so that the correct number of files are produced. The part that will not be correct will be the gridded PSF library for occulted observations. Updating this would have to be left to the user after creating the yaml files.

mperrin:
FYI @lpueyo @juliengirard  

My thinking is that we should get the MIRAGE parsing for APT to also work with coronagraphic images, and once we have that working, it should be straightforward to export the parsed information for use with pancake or pyNRC simulations too. This saves reinventing the wheel of having multiple APT-parsing implementations. And hopefully it will be relatively straightforward to extend the existing MIRAGE parsing for NIRCam templates to also support coronagraphy.

PR #615, opened 2021-01-28
APT parsing for NIRCam coronagraphy programs (and MIRI coronagraphy, too...)

mperrin:
Addresses #612 to add XML parsing support for NIRCam coronagraphy. 

First steps / work in progress, based on adapting from the NIRISS AMI parsing code (as suggested by @bhilbert4 since that one already deals with TA exposures). Not yet fully working but starting to get there. Needs work on the dither expansion now I think (?).

I'm testing this on my GTO program 1412; 1193 and 1195 are also good test cases.

### DIFF
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -461,8 +461,7 @@ class AptInput:
                     mode = input_dictionary['Mode'][index]
                     template = input_dictionary['APTTemplate'][index]
                     if (sub == 'FULL'):
-
-                        if mode in ['imaging', 'ts_imaging', 'wfss']:
+                        if mode in ['imaging', 'ts_imaging', 'wfss', 'coron']:
                             # This block should catch full-frame observations
                             # in either imaging (including TS imaging) or
                             # wfss mode
--- a/mirage/apt/read_apt_xml.py
+++ b/mirage/apt/read_apt_xml.py
@@ -1817,9 +1817,6 @@ class ReadAPTXML():
         """
         instrument = 'NIRCam'
 
-        long_filter = 'N/A'
-        long_pupil = 'N/A'
-
         if verbose:
             print(f"Reading template {template_name}")
 
@@ -1836,16 +1833,8 @@ class ReadAPTXML():
 
         parallel_instrument = False
 
-        DitherPatternType = None
         dither_key_name = 'DitherPattern'
 
-        # number of dithers defaults to 1
-        number_of_dithers = 1
-        number_of_subpixel_positions = 1
-
-        number_of_primary_dithers = 1
-        number_of_subpixel_dithers = 1
-        number_of_astrometric_dithers = 0
 
         # Check the target type in order to decide whether the tracking should be
         # sidereal or non-sidereal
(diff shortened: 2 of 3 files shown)


mperrin:
@bhilbert4 I got this started but would definitely appreciate any time or guidance you could spare for this. The APT parsing outputs are a little nonintuitive for me  and I'm sort of handwaving my way through adapting this. 

I've gotten as far as being able to run `yam = yaml_generator.SimInput(xml_file, pointing_file)` without error for a NIRCam coronagraphy program, but then the `yam.create_inputs()` stage errors out: 
```
AssertionError: Inconsistent table size from XML file (14) and pointing file (56). 
Something was not processed correctly in apt_inputs.
```

I suspect this is because I'm not expanding the dithers properly yet, or some such. I may poke at this a bit more but I expect you might be able to help steer me in the right direction?   (This isn't immediately urgent, I just wanted to see how far I could get after we talked on Slack yesterday)


FYI @juliengirard @lpueyo

bhilbert4:
No problem. I can have a look at it. Expansion for dithers is done later, after the xml is read in. You need only one entry per exposure setup, if that makes sense. I often see this error when I have forgotten to take into account assumptions that aren't explicitly in the xml file. For example, in the AMI case, there are actually 4 TA exposures, but there's no indication of this in xml. Usually by looking at the entries in the pointing file you can see what is missing.

mperrin:
Also still to do is to sort out the filter + subarray parts of this. This template breaks the mold compared to most of the NIRCam ones because it only returns data from one SCA, and it's sometimes LW and sometimes SW, depending on filter choice.

mperrin:
ooking at this more now and adding some debug print statements I see it's all confused in the parsing of the science exposures, and missing any entries for any of those...

mperrin:
Bonus: now includes APT parsing for MIRI Coronagraphy and MIRI imaging, enough to handle mixed programs containing those along with NIRCam imaging. (Doesn't output YAML for those, naturally, but internally does parse them enough to get the exposure information, enough to correctly interpret .pointing files containing rows for multiple instruments)

Tested on GTO programs 1411, 1412, 1193, 1194. Seems to be working pretty well. 

(Note, should probably test whether any of the small changes I made in the generic template parsing code might cause problems for other templates or other use cases? I would think not, but caution suggests verifying.)

mperrin:
*To Do:*

- [ ] The 'don't write unused detectors' part of the logic for NIRCam Coron should be made smarter for SW exposures, to only use A2 or A4 depending on coron mask, and skip the other 3 A detectors. Right now it's only omitting A5.  For LW coronagraphy it's working right (saving the A5 yaml and omitting all 4 of the SW ones). This needs some updates in yaml_generator.py around line 825. 
- [ ] Test and double check these changes don't have any inadvertent unintended consequences for parsing other templates? 
- [ ] Go over the outputs YAMLs with a fine-toothed comb for correctness of the values. 
- [ ] Test actually running MIRAGE image simulations on these; so far it's only tested through YAML generation.

mperrin:
I just realized this is still open and I had it at draft status. I've moved it to 'ready for review' though I have not done any work on it recently. I think it is indeed ready enough to merge now, and may be useful to do so alongside the other templates work in #637, perhaps.

PR #619, opened 2021-01-29
Allow gaia results to be strings or bytes

bhilbert4:
This PR solves an issue that has come up when upgrading to astropy version 4.2. Some of the results returned by the GAIA catalog search that were previously returned as bytes are now being returned as strings, and were causing `twomass_crossmatch` to crash. This change updates the function to support either returned data type.

@Skyhawk172


Skyhawk172:
I've tested this PR with some queries using both Astropy v4.0.0 and v4.2 and I'm getting identical results. So this seems to be working as intended.

PR #620, opened 2021-02-01
Add environment files for python 3.7 and 3.8

bhilbert4:
This PR add environment files for python 3.7 and 3.8, and changes the name of the file for python 3.6 to include the python version. The new files install almost everything via pip. Note that astroconda currently does not support python 3.8, so pip is necessary.


bhilbert4:
This is now ready to merge. The new set of environment files can then be used in #634 to create environments.

PR #621, opened 2021-02-02
Grism-related config file location

bhilbert4:
This PR updates the location where the grism-related config files are located. With this change, Mirage will look in:
$MIRAGE_DATA/<inst_name>/GRISM_<inst_name>/current/ for the files. Previously it was looking one directory higher, which was causing confusion since that directory may also have V1, V2, etc subdirectories. 

With this change, when new grism conf files are released, users must copy them into the 'current' subdirectory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,7 @@ For any questions about the `mirage` project or its software or documentation, p
 
 ## Current Development Team
 - Bryan Hilbert [@bhilbert4](https://github.com/bhilbert4)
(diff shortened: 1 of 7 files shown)


bhilbert4:
- [ ] New location for WFSS-related config files: "current" subdir
- [ ] Update wfss_simulator to use pre-computed background images in all cases
- [ ] Updates in tandem with NIRCam_Gsim updates for working with backgrounds
- [ ] Update grism_tso_simulator.py with the background use updates
- [ ] Any updates to notebooks needed?
- [ ] Any updates to documentation needed?

PR #639, opened 2021-03-11
Add FOCUSPOS keyword for FGS and NIS

Skyhawk172:
This header keyword is required for FGS and NIRISS by the WFSC team analysis software.  The "DEFAULT" value is just a placeholder, but in flight will be populated with a float indicating the position of the focus mechanism.

### DIFF
--- a/mirage/ramp_generator/obs_generator.py
+++ b/mirage/ramp_generator/obs_generator.py
@@ -2835,7 +2835,7 @@ class Observation():
             outModel.meta.instrument.channel = channel
 
         if self.instrument.upper() in ['NIRISS', 'FGS']:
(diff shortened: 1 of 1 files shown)


bhilbert4:
You should add a similar statement to the save_DMS() function, as this is the function that is normally used when saving the uncal files.

bhilbert4:
Did you test this? If the value is in flight is going to be a number, I wonder if having a string as the placeholder value will cause the datamodel to crash?

Skyhawk172:
Ah, right. I stopped short of testing the assignment and you're right, the DataModel is not liking it. I'll assign a float. Thanks

bhilbert4:
Not sure why the tests failed before. But they are all passing now. Merging.

Skyhawk172:
I think it was a server connection issue.

PR #647, opened 2021-03-17
Read NIRISS Ext Cal template from APT xml file

### LINKED ISSUE #644: yaml_generator is failing for NIRISS external calibration with target acq
Program 1094.
All observations are NIRISS external calibration. Observations 2 and 4 have TA exposures (which are listed as AMITA in the pointing file??). There are 4 TA entries in the pointing file for each of the two observations. These are being skipped when the xml is read in, resulting in a discrepancy in the number of entries.

bhilbert4:
This PR updates read_apt_xml.py to improve support for the NIRISS External Calibration template. With this update, the reading of external calibration observations is pulled out of read_generic_imaging_template() and instead done with a new function read_niriss_external_calibration_template(). Moving to a custom function made sense because of the wide range of options in the external calibration template, including TA exposures, and several dither pattern types which use inconsistent keywords to capture dither information.

Resolves #644

### DIFF
--- a/mirage/apt/read_apt_xml.py
+++ b/mirage/apt/read_apt_xml.py
@@ -2735,7 +2735,7 @@ class ReadAPTXML():
         # Determine what template is used for the parallel observation
         template = obs.find(self.apt + 'FirstCoordinatedTemplate')[0]
         template_name = etree.QName(template).localname
-        if template_name in ['NircamImaging', 'NircamEngineeringImaging', 'NirissExternalCalibration', 'NirissImaging',
+        if template_name in ['NircamImaging', 'NircamEngineeringImaging', 'NirissImaging',
                              'NirspecImaging', 'MiriMRS', 'FgsExternalCalibration']:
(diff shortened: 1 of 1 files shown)


bhilbert4:
One remaining question: when using PARALLEL dither type, the user can select a number of dither positions, but the APT pointing file seems to show only a single pointing. Right now there is a line in the code to force a single pointing in this case. Waiting to hear back if this is expected behavior. If so, we can make the code a little cleaner. If not we can leave it as-is and modify the code after APT is updated.

bhilbert4:
The PARALLEL dither pattern must be new. The NIRISS team was unaware of it. So for now we'll keep the number of dithers hardwired to 1, in agreement with what APT is doing.

bhilbert4:
Still need to remove NIRISSExternalCalibration from the list of parallel templates to be read by read_generic_imaging_template. Currently it looks like NIRISSExternalCalibration cannot be used as part of a parallel observation.

PR #649, opened 2021-03-24
Use ALLWISE by default. Add keyword to allow WISE all sky search

### LINKED ISSUE #638: Update WISE catalog that is queried
Currently when Mirage queries the WISE catalog for sources, it is using the WISE All-Sky catalog. However, there is a more recent version of this catalog, called the ALLWISE source catalog. While the more recent catalog has more total sources and presumably the best locations, Mirage was querying the older catalog because several NIRISS calibration targets were missing from ALLWISE. 

As a compromise, switch to querying ALLWISE by default, but allow the user to override that and query the older (WISE All-Sky catalog) if they choose. Both are available through astroquery, so the update will hopefully be relatively straightforward.

bhilbert4:
Resolves #638 

This PR updates the WISE source catalog query that Mirage performs to use the ALLWISE source catalog by default. This is the more recent WISE catalog and preferable in most situations. In some cases, for very bright sources, the older WISE All Sky catalog is better to use. Therefore, we've added the 'wise_catalog' keyword to get_all_catalogs(). With this, if a user specifies wise_catalog='wise_all_sky', then Mirage will revert to searching the older catalog. If the keyword is omitted, or given a value of 'allwise', then the new catalog is queried. The new catalog contains significantly more sources than the old one.


bhilbert4:
@mperrin @Skyhawk172 @obi-wan76 Also note this update. I'm not sure how important it is to your sims, but in my tests, the ALLWISE catalog was returning more sources than the WISE All Sky catalog.

PR #650, opened 2021-03-24
Look for correct WLP8 catalog colname in TSO notebook

### LINKED ISSUE #635: Source Catalogs have no magnitude specifically called nircam_f182m_magnitude
I was running the MIRAGE notebook today to make the accompanying SW imaging data that comes with the grism spectroscopy. I came across this error running the notebook here:
`https://github.com/spacetelescope/mirage/blob/master/examples/NIRCam_TSO_examples.ipynb`

It happens at this step
```python
img_tso = ImgSim()
img_tso.paramfile = img_tso_sw_yaml
img_tso.create()
```

> ValueError: WARNING: Catalog tso_imaging_source.cat has no magnitude column for NIRCAM specifically called nircam_f182m_magnitude, nor a generic 'magnitude' column. Unable to proceed.

I can make it go away by explicitly creating a column in `ptsrcs.cat` and `tso_imaging_source.cat` that has the header `nircam_f182m_magnitude` with a copy of the magnitude from `nircam_f182m_clear_magnitude`

bhilbert4:
Resolves #635 

This PR makes some small changes to the case where a weak lens is used in a source catalog.

- In cases where a weak lens is used, if the corresponding catalog column name nircam\_\<filter>\_wlp8\_magnitude or nircam\_\<filter>\_wlm8\_magnitude is not present, then an exception is raised. Previously Mirage would fall back to look for nircam\_\<filter>\_magnitude, however the +/-8 weak lenses have a significant effect on throughput and should not be ignored. This was also leading to confusing errors in the case where nircam\_\<filter>\_clear\_magnitude was present, but was never searched for.
- A bug in the NIRCam TSO notebook was fixed, to create nircam\_f182m\_wlp8\_magnitude columns in the source catalogs, where previously the column names were nircam\_f182m\_clear\_magnitude

### DIFF
--- a/examples/NIRCam_TSO_examples.ipynb
+++ b/examples/NIRCam_TSO_examples.ipynb
@@ -786,7 +786,7 @@
    "source": [
     "# Add source magnitudes\n",
     "tsimg_cat.add_magnitude_column([object_f182m_mag], magnitude_system='vegamag',\n",
-    "                               instrument='nircam', filter_name='f182m')\n",
+    "                               instrument='nircam', filter_name='f182m/wlp8')\n",
     "tsimg_cat.add_magnitude_column([object_f210m_mag], magnitude_system='vegamag',\n",
     "                               instrument='nircam', filter_name='f210m')\n",
     "tsimg_cat.add_magnitude_column([object_f470n_mag], magnitude_system='vegamag',\n",
@@ -877,7 +877,7 @@
    "source": [
     "# Add source magnitudes\n",
     "bkgd_cat.add_magnitude_column(bkgd_sources_f182m_mag, magnitude_system='vegamag',\n",
-    "                               instrument='nircam', filter_name='f182m')\n",
+    "                               instrument='nircam', filter_name='f182m/wlp8')\n",
     "bkgd_cat.add_magnitude_column(bkgd_sources_f444w_mag, magnitude_system='vegamag',\n",
     "                               instrument='nircam', filter_name='f444w')\n",
     "bkgd_cat.add_magnitude_column(bkgd_sources_f322w2_mag, magnitude_system='vegamag',\n",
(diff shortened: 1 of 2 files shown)


mperrin:
Hey Bryan, I'm a little surprised to see the comment that WLP/M8 have a significant effect on throughput, more so than WLP4... that seems exactly opposite my expectation (the WLP4 lens has a filter coating on it but the WLP/M8 I thought just had AR coatings.) Now maybe I just don't have an accurate mental model of these lenses but I'm curious about this. I guess I need to go re-check what the throughput data for those looks like. 

Is some part of this just specific to the use case for TSOs in which these get paired with the F182M filter?

We're going to have to double check whether this change would be disruptive to our OTE simulations - I think we originally implemented the "treat these as clear" fallback option because some of the target catalogs for the OTE sims were not populating those fields. If this is going to change we need to make sure our OTE sim catalogs will all have the right column names. FYI @obi-wan76 and @Skyhawk172

bhilbert4:
[Here](https://jwst-docs.stsci.edu/near-infrared-camera/nircam-instrumentation/nircam-weak-lenses#NIRCamWeakLenses-Transmission) is the figure that inspired my comment about the weak lens transmissions. 
The original motivation for this was that the TSO example notebook, where the observation uses F182M/WLP8, accidentally made catalogs with f182m_clear_magnitude columns, which was then causing Mirage to crash. So it was a trivial fix to make the catalogs with the correct magnitude column names.

bhilbert4:
But as I was working on keeping the fall-back option that we had put in place before, I saw that transmission curve and figured that the +/-8 transmission was too far from 1.0 to gloss over. We can return to that mode if you'd like. I would just add a much stronger warning in the log in the case where it does fall back to using the filter_CLEAR_magnitude column.

mperrin:
It's certainly more complete to include the transmission of the lenses in all cases, for sure. Agreed that's preferable. I just want to make sure all our sim setups are updated if need be so we don't stumble over this as an unexpected problem later.

bhilbert4:
Sorry, I don't want to create more work for you all.

PR #655, opened 2021-04-13
Pass the segmentation threshold value through

eas342:
Made sure the yaml_generator passes the segmentation_threshold through if a custom value for `segmap_flux_limit` or `segmap_flux_limit_units` are passed.


bhilbert4 (approved):
Good catch! I must not have tested with non-default values before.

PR #667, opened 2021-04-21
Log the absence of a filter from the ghost gap file only once

### LINKED ISSUE #666: Excesive log entries when a filter has no GAP value.
I am running a simulation of the astrometric field with the current version of Mirage installed today.  The first image in the simulation is for the F277W filter for which there is no ghost axis point defined.  This has resulted in 4.6 million lines in the log file, almost all of which are noting that the F277W filter has not GAP value and ghosts are not going to be used.  (By the way, I had not intended to have the ghost option on anyway.  I had assumed it would be off by default rather than on, because one really needs to have ghost images for each point source to do it properly, and I have a large number of sources to deal with.)

I would think it more reasonable to report this error once in the log file when a point source is processed for the source catalogue, and then not to repeat it N times more for however many stars are in the list.

Also, when I read the documentation page for the ghosts in NIRISS it sounded like all filters can use this option.  I see that the ghosts file contains 5 filters (F090W, F115W, F140M, F150W, and F200W) but lacks the other seven filters.  This probably should be made clear to users.

bhilbert4:
Resolves #666 

This PR adjusts the logging in catalog_seed_generator.py such that in cases where ghosts are requested, if the filter used does not support ghosts, only one log entry is made saying so.

### DIFF
--- a/docs/ghosts.rst
+++ b/docs/ghosts.rst
@@ -30,7 +30,7 @@ If **add_ghosts** is True, Mirage will calculate, for each point source in your 
 
 .. tip::
 
-    If you wish to run Mirage with a "gap summary" file other than that in the config directory, you can either replace the file in the config directory with your own, or update the value of NIRISS_GHOST_GAP_FILE in the constants.py file of the repository to point to your file. This may be useful for testing new ghost position calibration results.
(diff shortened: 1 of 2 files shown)


bhilbert4:
To do:

- [ ] Update documentation to list the filters that support ghosts

PR #670, opened 2021-04-26
Enable use of IRAC mosaic as seed image

bhilbert4:
This PR makes some changes to the code that creates simulated data from an input mosaic image. Previously this mode had only been tested using an HST mosaic. Testing with an IRAC mosaic revealed some issues that needed to be fixed. 

This is still not a very well tested part of Mirage compared to catalog_seed_image.py.


bhilbert4:
I'm going to merge this now so that the user who inspired the work can test it out. The updated notebook worked in my tests yesterday.

PR #671, opened 2021-04-26
Small fixes for `soss_simulator` 

hover2pi:
This PR fixes `soss_simulator` ability to make simulations with either a paramfile OR keyword arguments. If arguments are given, a paramfile is written so `dark_prep` doesn't get confused.


hover2pi:
@bhilbert4 I FINALLY got this working for long integrations without running into that broadcasting error in the `obs_generator` module. Would you mind taking a look at this when you get a chance? Thanks!

mirage/dark/dark_prep.py
```diff
@@ -652,16 +652,17 @@ def linearize_dark(self, darkobj, save_refs=True):
...
             self.params = params
         else:
             self.read_parameter_file()
 
-        # Get the log caught up on what's already happened
-        self.logger.info('\n\nRunning dark_prep..\n')
-        self.logger.info('Reading parameter file: {}\n'.format(self.paramfile))
-        self.logger.info('Original log file name: ./{}'.format(STANDARD_LOGFILE_NAME))
+            # Get the log caught up on what's already happened
+            self.logger.info('Reading parameter file: {}\n'.format(self.paramfile))
```
bhilbert4, line 664: If the user provides `params`, will `self.paramfile` be defined here?

mirage/ramp_generator/obs_generator.py
```diff
@@ -197,10 +197,10 @@ def add_crs_and_noise(self, seed, num_integrations=None):
                 ngroups = self.params['Readout']['ngroup']
             # TSO and moving targets
             elif seeddim == 4:
-                if num_integrations != seed.shape[0]:
-                    raise ValueError(('The number of integrations reported for the dark ({}) does not match that '
-                                      'implied by the size of the 1st dimension of the seed image ({}). Not sure '
-                                      'which to use.'.format(num_integrations, seed.shape[0])))
+                # if num_integrations != seed.shape[0]:
```
bhilbert4, line 200: Did you mean to leave this commented out?

mirage/ramp_generator/obs_generator.py
```diff
@@ -2829,7 +2829,8 @@ def save_DMS(self, ramp, zeroframe, filename, mod='1b', err_ext=None,
         with warnings.catch_warnings():
             warnings.simplefilter("ignore")
             current_time = datetime.datetime.utcnow()
-            start_time_string = self.params['Output']['date_obs'] + 'T' + self.params['Output']['time_obs']
+            # start_time_string = str(self.params['Output']['date_obs']) + 'T' + str(self.params['Output']['time_obs'])
+            start_time_string = '2022-08-27T12:49:00.000'
```
bhilbert4, line 2833: Did you mean to leave this in?

mirage/ramp_generator/obs_generator.py
```diff
@@ -2918,8 +2919,8 @@ def save_DMS(self, ramp, zeroframe, filename, mod='1b', err_ext=None,
         outModel.meta.observation.time = self.params['Output']['time_obs']
 
         # Create INT_TIMES table, to be saved in INT_TIMES extension
-        int_times = self.int_times_table(self.ramptime, self.params['Output']['date_obs'], self.params['Output']['time_obs'],
-                                         outModel.data.shape[0])
+        int_times = self.int_times_table(self.ramptime, '2022-08-27', '12:49:00.000', outModel.data.shape[0])
```
bhilbert4, line 2922: Did you mean to leave this in?

bhilbert4 (commented):
Looks good to me. Just a few small comments.

mirage/ramp_generator/obs_generator.py
hover2pi: This `ValueError` gets thrown for anything over 213 integrations if I leave it in but the mismatch seems to be corrected somewhere along the line if I comment it out, i.e. any number of integrations work if this `raise` statement is not here! I have NO idea why.

mirage/ramp_generator/obs_generator.py
hover2pi: No! This needs correcting. `self.params['Output']['date_obs']` is a `astropy.time.Time` object and I was having trouble making it into an appropriately formatted string. I'll fix it!

mirage/ramp_generator/obs_generator.py
hover2pi: Same issue as above. I just need to make sure the `date_obs` is a string.

mirage/dark/dark_prep.py
hover2pi: Actually, I'm realizing that I don't need to pass this dict to the dark_prep object anymore since `soss_simulator.SossSim` writes a paramfile before this is called. I'll remove the if/else from here.

mirage/ramp_generator/obs_generator.py
bhilbert4: Very strange. I've run NIRCam TSO exposures with more than 213 integrations and have never run across this. Is this a case where the seed image is being split into multiple files? Maybe that's why the number of integrations pulled from the dark file header isn't matching the size of the seed image? If file splitting is happening, then Mirage works with one exposure segment at a time, so NINT in the dark will be less than the total number of integrations in the exposure. Do you know if SOSS observations will undergo file splitting? Mirage is set up to split observations with > 160 full frame images (i.e. groups x ints). But I recently learned that DMS is only applying this rule to some observing modes. So Mirage is different than DMS since it will split e.g. NIRCam imaging data as well, but with a limit of 160 frames, which is hard to reach in imaging mode, fixing this has been low on the priority list.

hover2pi: Yes, that's what happening. I was running a 420 integration simulation and it split it into 213 and 207 int chunks but then breaks when 213 (the first chunk) doesn't match 420. There will be some pretty long SOSS simulations for sure. Should I leave this commented out or does it break something else?

mirage/ramp_generator/obs_generator.py
hover2pi: Ok I figured out this was because I needed to put the date and time strings in quotes in the yaml file. It's fixed now.

mirage/dark/dark_prep.py
hover2pi: Removed!

mirage/ramp_generator/obs_generator.py
hover2pi: Yes it does!

hover2pi:
I found out planetary transits weren't being injected into the raw simulations. This FINAL commit fixes that. If things look good on your end, feel free to merge it! Thanks @bhilbert4 !

hover2pi:
Haha, one more commit there. We found an error during our commissioning rehearsal today.

bhilbert4:
Looks good to me. Do you want to add in the missing yaml file from #676 and then we can merge?

hover2pi:
@bhilbert4 The build for this PR passed after my previous commit but when I resolved the merge conflicts with master, the tests all failed with that `batman` version. I fixed another bug too which was injecting a flat transit spectrum into the simulations given planetary data. 

Do you know how to reconcile these? Thanks!

bhilbert4:
That's strange how the log entries talking about a newer conda version are mixed with entries talking about cloning batman. I haven't seen that before. Maybe some sort of CI hiccup? Upgrading to batman 2.4.8 resolved the problems I had been seeing with installation.

bhilbert4:
Looks like the python 3.7 test is working with the urllib3 package for python 3.8?

hover2pi:
The 3.7 build failed but the log is ambiguous. Can you rerun it  @bhilbert4?

hover2pi:
OK, ready to merge! Thanks for your help @bhilbert4 !

PR #686, opened 2021-06-01
New gain and fluxcal

bhilbert4:
This PR updates the flux calibration table file for NIRCam to use new values that are based on the bootstrapped gain values from CV3 testing. These gain values are ~20% lower than those used previously, resulting in flux calibration values that are ~20% lower. 

This PR should be merged as closely as possible to the time when the updated gain and photom reference files with these data are delivered to CRDS. In this way, consistent flux cal values will be used for data creation and data reduction.


bhilbert4:
Gain files and photom files have been delivered to CRDS. All tests are passing here, so I'm going to merge this branch in order to keep Mirage in sync with CRDS.

PR #700, opened 2021-06-23
Fixes bug where simulating F277W with yaml file tries to simulate order 2

hover2pi:
This is a very small change but fixes SOSS simulations of F277W so it is not looking for a nonexistent order 2.


bhilbert4:
Looks good to me!

PR #713, opened 2021-08-17
WIP: Some dtype fixes to get poisson noise working properly for SOSS simulations

hover2pi:
This PR fixes a bug that was not allowing pixels to saturate and were breaking the poisson noise modeling for large simulations in SOSS mode. The issue was primarily array dtypes butting heads so forcing them all to float64 seems to fix it.

Also adds a keyword arg to replace the dark current step with all zeroes for debugging purposes.

Also moves all `message` method calls to `logging` calls instead so they are written to the log rather than just printed in the terminal.


hover2pi:
This looks ready to go if you had time for a quick review @bhilbert4 . Thanks!

bhilbert4 (approved):
Looks good to me! Sorry about the delay. I thought it was still in progress.

PR #720, opened 2021-08-31
BlackBody1D to BlackBody

### LINKED ISSUE #715: Update astropy Blackbody1D to Blackbody
@hover2pi 
Blackbody1D has been deprecated since astropy 4.0. It looks like it is gone in astropy 4.3 (which it looks like github actions is using). We'll need to update the code in the soss_simulator to use Blackbody instead. Also note that astropy.modeling.blackbody has been removed. The soss_simulator currently imports the definition of FLAM from there, so that will need to be updated as well. I couldn't find a replacement for it in astropy. It is present in synphot, or it could be added directly to the soss_simulator or constants.py. (FLAM = u.erg/u.s/u.cm**2/u.AA )

I tried making these updates, but the flux conversion still failed, using the code below:

```
        bb = BlackBody(temperature=teff * q.K)
        wav = np.linspace(0.5, 2.9, 1000) * q.um
        flux = bb(wav).to(FLAM, q.spectral_density(wav)) * 1E-8 * scale
```

It throws an error:
`UnitConversionError: 'erg / (cm2 Hz s sr)' and 'erg / (Angstrom cm2 s)' (spectral flux density wav) are not convertible`

I tried playing around with the `equivalencies` keyword, but wasn't able to get it to work.

bhilbert4:
Resolves #715 

This PR updates the SOSS simulator to use astropy's BlackBody model rather than BlackBody1D. The latter was removed from astropy in version 4.3.1. 

setup.py, requirements.txt, and the environment files have also been updated to specify astropy 4.3.1.


hover2pi (approved):
Fantastic. Thanks Bryan!

PR #725, opened 2021-09-13
Renormalize gridded PSFs with a total signal over 1

bhilbert4:
This PR contains a small update during the normalization checks that are performed on gridded PSF libraries. In some cases, the total normalized signal in a distorted PSF can be just > 1. This is due to the model WebbPSF uses when creating and distorting the PSF. See https://github.com/spacetelescope/webbpsf/issues/487

With this PR, rather than raising an exception, when a PSF with a signal > 1 is found, it is renormalized to be 1.0. Previously, the normalization check examined only the first PSF in the PSF grid in order to determine the normalized signal level. With this PR, Mirage now checks all PSFs in the grid, as there are cases where some PSFs have total signals under 1.0 while others are over 1.0.

### DIFF
--- a/mirage/psf/psf_selection.py
+++ b/mirage/psf/psf_selection.py
@@ -88,7 +88,7 @@ def check_normalization(lib, lower_limit=0.80, upper_limit=1.0, renorm_psfs_abov
             if total_signal > upper_limit:
                 if renorm_psfs_above_1 and total_signal <= absolute_upper_limit:
                     lib.data[i, :, :] = lib.data[i, :, :] / total_signal
-                elif:
+                else:
                     # We will end up here if the total signal is above 1,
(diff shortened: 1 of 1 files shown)


bhilbert4:
I've added back in a check for incorrectly normalized PSFs. Any PSF with a normalized total signal that is above some absolute upper limit will be considered bad, and will still raise an exception. For the moment, I've set the absolute upper limit to be 1.5. @mperrin what do you think? Is that high enough to safely avoid catching the PSFs that have a total signal that is slightly over 1.0 due to the distortion added by WebbPSF as described in the webbpsf issue above?

bhilbert4:
@mperrin do these changes look good to you?

hover2pi (approved):
Seems to work as intended.

PR #726, opened 2021-09-13
Catalog crossmatch inconsistencies

### LINKED ISSUE #708: Inconsistent source catalog entries
From @obi-wan76 :

I’ve been looking at some discrepancies in the MIRAGE catalog that you may be able to provide some clarification. I’ll try to summarized the problem with an example:
 
For our program Fine Phasing 3 (PID 1160) we are using a target star with the following parameters:
TYC-8902-1241-1
k = 8.918
ra = '06 25 25.0836'
dec = '-63 56 41.59'
 
When running create_catalog.get_all_catalogs with a large radius (7’ à 420”) I get the target from the catalog to be:
2
96.354515
-63.944885
9.66733
With nircam_f212n_clear_magnitude = 9.66733. The coordinates are consistent with the target star but, to me,  a SW F212N magnitude of 9.67 seem too faint for a K = 8.918 star.  So, I ran create_catalog.get_all_catalogs with a different search radius of 10”, so to isolate the target star,  from which I got a single source:
1
96.354515
-63.944885
8.9304695
 
The same coordinates as before but with a different SW F212N magnitude of 8.93, which seems more consistent for a K=8.918 star. However, this 8.93 mag source doesn’t show up in the larger catalog (7’, above).
 
If I increase the search radius to 1’ then I see both matching sources (9.66 & 8.93 mag):
96.354515
-63.944885
9.66733
96.361626
-63.952843
8.9304695
 
However, this time they swapped the coordinates. At some point, if I keep increasing the search radius the 8.93 mag source dropped out from the catalog all together.
 
So, I guess the question is, what happened to the 8.93 mag source? That source seems to be a more reasonable match to the target star with a K = 8.918.
 
Something similar happened with another of our programs where the “best” match source disappeared in the larger catalog (or as we increased the search radius). My concern is that during our WF team practices we have our K mag requirements for target selection and SBR/SN values but it seems that for some of these sources we may be simulating a target fainter than we should expect.

bhilbert4:
This PR fixes a bug in the Gaia/2MASS/WISE catalog query and generation code. An indexing error was causing the incorrect magnitude values to go into the interpolation function in some cases. This was resulting in different magnitudes for a given object depending on the width of the box on the sky used for the catalog search.

Resolves #708 

Initial test results:

Search center:
ra = 96.354515
dec = -63.9448861111

At these coords, we should find a source with kmag = 8.918

10" box_width (results are unchanged, and look reasonable):

Master branch:
          x_or_RA                         y_or_Dec          f200w  f444w f356w f070w f090w 
96.3545150756836 -63.94488525390625  8.97    8.94    8.87    10.88 10.29

Fix:
96.3545150756836 -63.94488525390625 8.97     8.94    8.87    10.88 10.29

60" box width:

Master branch: 
Note the F200W mag for the first source, which is the source from the 10" search above,
as well as the F200W mag for the 4th source, which is much brighter than in the other filters
96.3545150756836 -63.94488525390625  9.75     8.94    8.89    10.88   10.47
96.35475158691406 -63.9473495483398 19.70    19.65  19.66    19.95  19.81
96.3580093383789 -63.94111633300781   16.30   16.22   16.23    17.79  17.34
96.36162567138672 -63.9528427124023   8.97     13.48   13.47    14.75  13.14
96.335693359375 -63.941307067871094 15.82    15.76   15.75     16.80  16.80
96.37287139892578 -63.9367027282714  17.58     17.49   17.56     17.58  17.58

Fix: 
Note how the first source magnitudes now matches that from the 10" search. The fourth source
now has magnitudes across all filters that are more consistent and reasonable.
96.3545150756836 -63.94488525390625   8.97     8.94    8.87     10.88  10.29
96.35475158691406 -63.9473495483398  19.70   19.65   19.66    19.95  19.81
96.3580093383789 -63.94111633300781   16.30    16.22  16.23     17.79   17.34
96.36162567138672 -63.9528427124023   13.50    13.48  13.47    14.75   14.34
96.335693359375 -63.941307067871094   15.82   15.76   15.75    16.80   16.80
96.37287139892578 -63.9367027282714   17.58     17.49   17.56    17.58   17.58

In general, I think a lot of the catalog creation functions could probably be streamlined in order to avoid indexing errors like this. But given that the results with this fix in place look correct, a more involved code revision can be saved for later.

### DIFF
--- a/mirage/catalogs/create_catalog.py
+++ b/mirage/catalogs/create_catalog.py
@@ -643,13 +643,34 @@ def get_all_catalogs(ra, dec, box_width, besancon_catalog_file=None, instrument=
     gaia_cat, gaia_mag_cols, gaia_2mass, gaia_2mass_crossref, gaia_wise, \
         gaia_wise_crossref = query_GAIA_ptsrc_catalog(outra, outdec, box_width)
     twomass_cat, twomass_cols = query_2MASS_ptsrc_catalog(outra, outdec, box_width)
+    wise_cat, wise_cols = query_WISE_ptsrc_catalog(outra, outdec, box_width, wise_catalog=wise_catalog)
+
+
+    print(gaia_cat.colnames)
+    print(gaia_2mass.colnames)
+    print(gaia_2mass_crossref.colnames)
+    print(gaia_wise.colnames)
+    print(gaia_wise_crossref.colnames)
+
+
+    print('GAIA: ', outra, outdec, box_width)
+    print(gaia_cat['ra', 'dec', 'designation', 'ra_error', 'dec_error', 'phot_g_mean_mag', 'phot_bp_mean_mag', 'phot_rp_mean_mag'])
+    print(gaia_2mass['ra', 'dec', 'designation'])
+    print(gaia_2mass_crossref['ra', 'dec', 'designation', 'designation_2'])
+    print(gaia_wise['ra', 'dec', 'designation'])
+    print(gaia_wise_crossref['ra', 'dec', 'designation'])
+    print('\n\n')
 
     print('2MASS: ', outra, outdec, box_width)
-    print(twomass_cat)
+    print(twomass_cat['ra', 'dec', 'designation', 'err_maj', 'err_min', 'j_m', 'h_m', 'k_m'])
     print('')
     print(twomass_cols)
+    print('\n\n')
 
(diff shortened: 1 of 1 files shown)


bhilbert4:
- [ ]  Still need to remove lots of print statements from testing

PR #730, opened 2021-09-23
Fix bug in NIRISS filter list for imaging mode when creating cats

bhilbert4:
This PR fixes a problem in the for_proposals() within create_catalog.py. For an input observation that uses NIRISS in imaging mode, the filter/pupil values being read in from the APT xml file were being placed in the wrong entry of the exposure dictionary. This was then causing for_proposal() to find an empty list of NIRISS filters to interpolate source magnitudes to. 

These updates populate the correct keys in the exposure dictionary. I've also increased the area covered by the created catalogs in the case of a single target in the xml file. The code was originally designed with multiple targets in mind. With a single target, the size of the area on the sky to query for was too small to cover all of the detectors. Now, if there is a single target, the catalogs output by for_proposal() will cover 6' x 6', which is large enough to cover both NIRCam modules.

### DIFF
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -1116,7 +1116,7 @@ def get_filters(pointing_info):
             short_filters = np.array(pointing_info['FilterWheel'])[good]
             short_pupils = np.array(pointing_info['PupilWheel'])[good]
 
-            short_filter_only = np.where(short_pupils == 'CLEAR')[0]
+            short_filter_only = np.where(((short_pupils == 'CLEAR') | (short_pupils == 'CLEARP')))[0]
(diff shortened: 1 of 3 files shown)


bhilbert4:
I tested the changes using proposal 1963, which is where the problem was first noticed. Catalogs are now successfully created, and result in good seed images.

bhilbert4:
@hover2pi can you give this a quick look? I know it's kind of throwing you into the deep end in terms of parts of the code base that you probably haven't looked at before, but at least it's a small change. :)

hover2pi (approved):
That all looks reasonable and simple enough!

bhilbert4:
Thanks!

PR #740, opened 2021-10-22
Update docs to correct moving_target typos

bhilbert4:
This PR corrects some outdated mode names in the documentation. "moving_target" is not an accepted mode. Those references have been removed.


hover2pi (approved):
Ok!

PR #743, opened 2021-10-26
Extended nonsidereal sources

### LINKED ISSUE #736: Allow non-sidereal source catalog to use extended sources
We need to add a filename column to the non-sidereal source catalog. Currently the catalog does allow an object to be classified as extended, but then there is no filename column to add the name of the stamp file. Adding the filename column should be harmless in the cases where the object is a point source or galaxy. Will need to confirm with testing.

Also, when the ephemeris file is provided, the RA, Dec , and velocities in the catalog appear as sets of empty quotes. (The default values must be translated to empty strings when the file is written.) It might be better to have them set to None, as long as that works downstream.

This oversight was discovered when attempting to create simulations of program 1373 (Jupiter).

We have a goal of the end of the month to get at least some simulations made, so this is pretty high priority.

bhilbert4:
Resolves #736 

This PR makes updates to tracked, non-sidereal source observations such that extended sources and galaxy sources can be used as the non-sidereal source. Development was done using JWST program 1373 (Jupiter). Updates were also made to the APT xml reader in order to correctly process the NIRSpec IFU and MIRI MRS observations which are a part of that program.

With the changes at the moment, program 1373 can be successfully simulated.


bhilbert4:
Tests are all passing and my testing shows things working as expected. I'm going to merge this now so that the changes can be further tested by external collaborators.

PR #744, opened 2021-10-27
Fix NRCB5_MASKLWB subarray name

bhilbert4:
This PR corrects the DMS subarray name that is associated with the NRCB5_MASKLWB aperture name in the PRD. The subarray size is 320x320 pixels, not the 640x640 pixels that the associated SW channel subarrays are.


hover2pi (approved):
Nice.

PR #749, opened 2021-11-10
Add moving_target_position extension to output files

bhilbert4:
For simulations of non-sidereal targets, this PR adds the moving_target_position extension to the output uncal files. It also populates the MT_RA and MT_DEC header keywords, which are needed when running non-sidereal target data through the calibration pipeline.


bhilbert4:
Note that two of the column names in the moving target position table are going to change with the next release of the jwst package. mt_detector_x and mt_detector_y will change to mt_sci_x and mt_sci_y. Let's keep the old names for now, and update after the next release of jwst. I'll create a new issue for this.

PR #751, opened 2021-11-23
Updated NIRISS ghost gap file

bhilbert4:
This PR adds some new entries to the config file describing where to place optical ghosts for NIRISS simulations. The new entries cover new filters.

- For FW filters, GAPs from CV1RR are available (Martel TR).
- Still missing F356W and F444M. According to the TR, they should be similar to F277W and F480M, respectively; thus, GAPs for the two filters are substituted accordingly.
- F480M seems to have two GAPs. For convenience, GAP1 from the table will be used for now.
- Flux ratios were not reported form the FW filters and are set to 1.


bhilbert4:
Note that I had to switch the filter and pupil values for the new entries in order to make them consistent with the previous entries. I ran a test using F444W and the ghosts were successfully added to the scene.

mtakahiro (approved):
tested and all results look good!

PR #752, opened 2021-11-23
Add APERTURE and PPS_APER keywords

bhilbert4:
This PR adds code to allow Mirage to populate the APERTURE and PPS_APER header keywords in the final output files. The value for PPS_APER comes from the "Aperture Name" column of the APT pointing file, which gets the information from the PRD. The APERTURE keyword is derived from this same value using a mapping within obs_generator.py

Still need to test this. Imaging mode seemed fine. Grism TSO sims were crashing for some unrelated reason.


bhilbert4:
Grism TSO crashes were stemming from multiprocessing. Nor tested a grism TSO case, which was successful.

bhilbert4:
@hover2pi do you have time for a quick review?

hover2pi (approved):
👍

PR #753, opened 2021-11-24
Download Niriss ghost file from niriss_ghost repository

bhilbert4:
This PR changes where Mirage gets the niriss ghost gap file from. Rather than being included in the Mirage repository, the file is now downloaded from the [niriss_ghost](https://github.com/spacetelescope/niriss_ghost) repository and placed in Mirage's config subdirectory. With this, we now treat the niriss_ghost repository as containing the master copy of this file. 

At the moment, Mirage is set up to download the file only if it does not already exist in the config subdirectory. But then how would a user know that there is an updated version of the file in the niriss_ghost repo and delete their local copy of the file? Perhaps, since this is a tiny ascii file, Mirage should always download a new copy of the file whenever it creates a NIRISS simulation that will include ghosts.


bhilbert4:
Updated to download the niriss ghost gap file each time Mirage runs. This will ensure that the latest version is always being used.

bhilbert4:
@mtakahiro should I bother merging this PR? This will have Mirage download the gap file each time it runs. Or should I abandon this and wait for the update that will use the niriss_ghost repo instead? The only reason I can think of to merge this one would be if there are people currently making NIRISS simulations that use the new filters in the updated gap file, and they want to get ghosts in now. I'm happy to do either.

mtakahiro:
@bhilbert4 yes please go ahead. I'm not sure how long the other option will take. Your solution actually meets our requirements at least for now.

bhilbert4:
Great! Would you tick the box as the official reviewer? You can see in constants.py that I've hardcoded the URL to the gap file, and then use download_file() to download it early in the seed image creation process.

mtakahiro (approved):
all changes look good. Tested on my machined too. Approved!

mtakahiro (approved):

PR #759, opened 2022-01-12
Version 2.2.1 release to master

bhilbert4:
Trivial changes. Just making a release.


bhilbert4:
Test results were inconsistent across python versions. Failed once in 3.9 but passed on all others. Another time it failed only in 3.7. I lowered the bar for matching the expectation value from being an exact match to matching within 1e-14 arcsec. We can revert later if this is actually an issue.

Issue #757, opened 2021-12-03
Update column names in moving target position table


bhilbert4:
Per https://github.com/spacetelescope/jwst/pull/6485 the mt_detector_x and mt_detector_y columns in the moving target position table are going to be renamed to mt_sci_x and mt_sci_y in the next release of the jwst package (1.3.4). Once that is released, we'll need to go into mirage and updated the column names in ramp_generator/moving_target_position_table.py

PR #761, opened 2022-01-15
#757 updated columns in moving target position table

ctava:


bhilbert4 (approved):
Thanks @ctava! Looks great.

PR #762, opened 2022-02-01
Fix failing simulations with NIRISS raw darks

bhilbert4:
This PR tweaks the code for reading in dark current files to fix an issue seen when using a raw NIRISS dark. The raw NIRISS darks in the reference file collection contain no zeroframe extension. This was causing zeroframe to be set to None. With this fix, as long as the dark current file uses a RAPID readout pattern, Mirage will grab the initial group of the dark file and use it as the zeroframe. 

The larger problem that was being seen with these raw darks was that the metadata was not being read in properly. Updates to the way the metadata are read in have fixed the problem.


bhilbert4:
pysiaf 0.14.0 was giving errors when working with some of the new NIRCam apertures. So for now we pin it to 0.13.0.

bhilbert4:
@hover2pi do you have a few minutes to look this over at some point?

hover2pi (approved):
Cut and dry. I love it.

PR #767, opened 2022-02-11
Fix incorrect galaxy position angles

bhilbert4:
This PR fixes a bug in the orientations (position angles) of galaxies in cases where PAV3 is non-zero.

- [x] Still need to make sure the same bug does not apply to extended sources.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,11 @@ Moving Targets
 Updated the MOVING_TARGET_POSITION table from detector -> sci. (#757)
 
 
+Galaxy and Extended Source Position Angles
+------------------------------------------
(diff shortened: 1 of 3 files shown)


bhilbert4:
Extended sources look incorrect also. Need to fix.

bhilbert4:
Both galaxy and extended sources now appear to be rotated correctly for given position angles and telescope PAV3.

bhilbert4:
@hover2pi do you have a few minutes to give this a look? The important part of the change is the now-simpler equation for translating a galaxy's position angle relative to North into the angle between the semi-major axis and the detector x axis. I also had to use different equations for galaxies vs extended sources because Mirage assumes that the stamp image of an extended source is oriented North up.

hover2pi (approved):
Good to go!

bhilbert4:
Thanks!

PR #772, opened 2022-03-02
Import irsa from new location

### LINKED ISSUE #673: Methods to find v3pa on a date reference self
When trying to run `default_obs_v3pa_on_date` I got an error regarding the following line because of the reference to `self.xml_skipped_observations`: 
https://github.com/spacetelescope/mirage/blob/a915341ca2fc25b5a4afe0ad558ebd5d609cd48d/mirage/yaml/yaml_generator.py#L2357

I noticed that the same issues exists for `all_obs_v3pa_on_date`:
https://github.com/spacetelescope/mirage/blob/a915341ca2fc25b5a4afe0ad558ebd5d609cd48d/mirage/yaml/yaml_generator.py#L2392

You wouldn't reach this error if a `pointing_table` is passed in.

bhilbert4:
Resolves #673

Import Irsa from the new location within astroquery

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 (Unreleased)
 ============
 
+Catalogs
+--------
+
+Update the import location of Irsa from astroquery.irsa to the new location of astroquery.ipac.irsa (#772)
+
+
 Dark Current
 ------------
 
--- a/environment_python_3.7.yml
+++ b/environment_python_3.7.yml
@@ -10,7 +10,7 @@ dependencies:
 - pip:
     - asdf==2.7.2
     - astropy==4.3.1
-    - astroquery==0.4.1
+    - astroquery==0.4.5
     - bokeh>=2.2.3
     - batman-package==2.4.8
     - crds==10.3.8
--- a/environment_python_3.8.yml
+++ b/environment_python_3.8.yml
@@ -8,7 +8,7 @@ dependencies:
 - pip:
     - asdf==2.7.2
     - astropy==4.3.1
-    - astroquery==0.4.1
+    - astroquery==0.4.5
     - batman-package==2.4.8
     - bokeh>=2.2.3
     - crds==10.3.8
--- a/environment_python_3.9.yml
+++ b/environment_python_3.9.yml
@@ -8,7 +8,7 @@ dependencies:
 - pip:
     - asdf==2.7.2
     - astropy==4.3.1
-    - astroquery==0.4.1
+    - astroquery==0.4.5
     - batman-package==2.4.8
     - bokeh>=2.2.3
     - crds==10.3.8
--- a/mirage/catalogs/create_catalog.py
+++ b/mirage/catalogs/create_catalog.py
@@ -20,7 +20,7 @@ from astropy.io import ascii
 from astropy.table import Table, join
 import astropy.units as u
 from astroquery.gaia import Gaia
-from astroquery.irsa import Irsa
+from astroquery.ipac.irsa import Irsa
 from pysiaf.utils.projection import deproject_from_tangent_plane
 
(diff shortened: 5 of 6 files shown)


bhilbert4:
Ugh. It looks like the table column names returned by the GAIA query have changed. It looks like this is not at the astroquery level, but values returned by GAIA itself. In the returned GAIA table, as well as the gaia_twomass table, 'description' became 'DESCRIPTION'. And in the gaia_twomass_crossref table, 'designation_2' became 'DESIGNATION'. 

Confusingly, in the first two cases, 'DESIGNATION' gives the source name in the GAIA catalog, but in the gaia_twomass_crossref table 'DESIGNATION' refers to the name of the source in 2MASS, while the unchanged 'designation' column gives the GAIA source name.

So, before/after column names:

gaia_cat: (searching gaiadr2.gaia_source)
    BEFORE: ['solution_id', 'designation', 'source_id', 'random_index',..... 
    AFTER:    ['solution_id', 'DESIGNATION', 'source_id', 'random_index',....

gaia_twomass: (SELECT ra,dec,ph_qual,j_m,h_m,ks_m,designation FROM gaiadr1.tmass_original_valid)
    BEFORE: ['ra', 'dec', 'ph_qual', 'j_m', 'h_m', 'ks_m', 'designation']
    AFTER:    ['ra', 'dec', 'ph_qual', 'j_m', 'h_m', 'ks_m', 'DESIGNATION']

gaia_twomass_crossmatch:  (SELECT field.ra,field.dec,field.designation,tmass.designation)
    BEFORE: ['ra', 'dec', 'designation', 'designation_2']
    AFTER:    ['ra', 'dec', 'designation', 'DESIGNATION']

bhilbert4:
I don't understand though how tests haven't been failing because of this before now.

bhilbert4:
This update breaks backwards compatibility. If a user doesn't have astroquery 0.4.5 then the new import statement will fail. I think the table column name fixes are not tied to the astroquery version though, since that is info provided by GAIA to astroquery during the query.

bhilbert4:
@KevinVolkSTScI @jotaylor  have you used Mirage recently to create any source catalogs? This update to the GAIA query results took me by surprise, and I can't find any documentation about it anywhere. See the comments above for details.

jotaylor:
> @KevinVolkSTScI @jotaylor have you used Mirage recently to create any source catalogs? This update to the GAIA query results took me by surprise, and I can't find any documentation about it anywhere. See the comments above for details.

No, I haven't either Bryan. If you want I can ask around NIRISS?

bhilbert4:
@jotaylor I'm curious, if someone tried to create a catalog now, without updating any part of Mirage, whether it would fail because of these updated column names. I expect that it will, unless there's something special/weird with my version if Mirage. Do you have a previously-run command around that you could re-run? Or, if you have a cloned copy of Mirage, you could go into the tests directory and run `pytest test_catalog_generation.py -k test_get_all_catalogs`.  Really, I'm just paranoid about this since I wasn't able to find any description of it in the GAIA API documentaiton.

jotaylor:
@bhilbert4  I'm not 100% sure how to interpret the results, but it will probably mean something to you. I already had a cloned copy of mirage and ran the command you said. I'm using mirage version `2.2.1.dev31+gea1b8f1` and astroquery version `0.4.3`. 
I've attached the output.
[test_03Mar22.txt](https://github.com/spacetelescope/mirage/files/8180536/test_03Mar22.txt)

bhilbert4:
@jotaylor excellent, thanks! That's exactly the error I was getting before. The returned table is missing the 'designation' column (which has been renamed 'DESIGNATION'), which is causing the code to fail.

bhilbert4:
Comparison of the output table with these new changes to that made with an old version of the software (located in the test_data directory) shows good agreement. So I think this is ready to merge.

PR #779, opened 2022-03-30
Have WFSS simulations optionally use pickle files of sources

bhilbert4:
These are changes implemented by @NorPirzkal 

In this code, catalog_seed_image() saves individual sources and their associated segmentation maps in pickle files, which the disperser then picks up and uses to do the dispersion. By working on individual sources this way, overlapping sources are all handled correctly. 

The pickle files are optional. If they do not exist, then the behavior is the same as it has been.

The last note I saw about this on slack, Nor said it is only implemented for point sources and galaxies. 
- [x] Check if this is the case, and implement for all sources if necessary

@NorPirzkal anything to add? This is the code you posted to slack on Feb 15.


bhilbert4:
Code changes were actually made by @NorPirzkal, so I can serve as reviewer.

bhilbert4:
These updates look good to me. Merging.

PR #790, opened 2022-04-27
Remove pixel area map references

### LINKED ISSUE #786: Remove mention of PAM from documentation
We switched from multiplying the seed image by the pixel area map to applying the flat field to the raw ramp. But it looks like there is still a mention of the pixel area map on read the docs.

bhilbert4:
This PR contains code to remove the pixel area map from Mirage. Mirage has not used the PAM for quite a while, in favor of using the flat field to introduce pixel-by-pixel effects (including effects from varying pixel area across the detector). So we remove PAM references here in order to prevent confusion.

Resolves #786


bhilbert4:
Ran a test case where I created some new yaml files and ran Mirage, and everything completed successfully. Also ran Mirage using an old yaml file, and that completed successfully as well.

hover2pi (approved):
Looks good!

PR #791, opened 2022-04-27
Remove pipeline config files

### LINKED ISSUE #742: Remove pipeline cfg files
The jwst pipeline has moved away from the old cfg files. We should remove these in Mirage as well.

bhilbert4:
Resolves #742 

This PR removes the pipeline configuration files (*.cfg) from Mirage. The JWST pipeline has switched to using parameter reference files, and these cfg files are no longer supported.

After some debate, I decided to remove the files completely and not add support for the parameter reference files. This is because the only place Mirage uses them is to linearize a raw dark current file. In this case, the superbias and reference pixel signal is subtracted, and the data are linearized. The simulated sources are added, the ramp is unlinearized, and the superbias and reference pixel signal is added back in. So really, the exact details of how the reference pixel signal is subtracted (that seems like the only step where someone would want to deviate from the defaults) don't matter, as the signal will be added back in at the end.

### DIFF
--- a/docs/dark_preparation.rst
+++ b/docs/dark_preparation.rst
@@ -8,7 +8,7 @@ After creating the seed image `mirage` moves on to preparing the input dark curr
 Adjust Number of Frames/Groups and Integrations
 -----------------------------------------------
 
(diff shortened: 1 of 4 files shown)


bhilbert4:
@hover2pi, when you have a free moment, I've got 2 PRs that I finished a while ago but never got around to asking you to review. This one and the other are largely removing unused files/features, so hopefully should be quick to review. Thanks!

hover2pi (commented):
Adios, archivos!

PR #805, opened 2022-10-04
Update galaxy orientation/position within stamp images

bhilbert4:
This PR makes tweaks to galaxy orientations within stamp images to correctly account for rotation


bhilbert4:
@hover2pi this is the other tweak based on user feedback. I had switched x and y in one of the position angle calculations, and the equation that I had been using wasn't quite right for non-zero telescope PAV3. With this fix, the results look good, so this is ready for review as well if you have a few minutes.

hover2pi (approved):
No objections here!

PR #808, opened 2022-11-10
Dependency update nov2022

bhilbert4:
Update dependencies for the 3.8 and 3.9 environments.


bhilbert4:
One test was failing because the new version of pysiaf shifted the distortions by enough that our "truth" values from the previous pysiaf version were incorrect. I removed the truth values, so that future changes in the distortion don't get caught as errors. This was for a test where we created a siaf instance for an instrument/aperture and calculated the x,y values for a given RA, Dec and compared to a truth value stored in the file. Given that distortions in pysiaf will always be tweaked, it doesn't make a whole lot of sense to have that as a test. Now, it tests only that the round trip from RA, Dec -> x,y -> RA, Dec returns the original RA, Dec values, which should always be true.

The other failing test was some problem with the CRDS server.

bhilbert4:
Merging w/o review so I can create a release.

PR #809, opened 2022-11-15
Update filter info with results from commissioning

bhilbert4:
Update the filter throughput curves that Mirage uses with those derived from commissioning. Along with this, update the Vegamag/ABmag/STmag/PHOTFLAM/PHOTPLAM values in the zeropoints file.


bhilbert4:
Trying to understand a failing test. 
Count rates that Mirage calculates using synphot and combining filter throughput curves and rescaled spectra are not agreeing with count rates calculated by the magnitude_to_countrate function. For Vegamags the difference is small (couple percent). For ABmags, the difference is a factor of 4-5.

Updated definition for ABmags from the updated nircam abs flux cal and zeropoints page:
ZPAB is the zeropoint:  ZPAB = –6.10 – 2.5 log10(PIXAR_SR[sr/pix]). For example, ZPAB = 28.0 for PIXAR_SR = 2.29e-14 (corresponding to pixel size 0.0312").
compare with the equation mirage uses in magnitude_to_countrate:
-2.5 * np.log10(photfnu * count_rate / 1.) - 48.599934378

When I plug in the new photfnu value and a count rate of 1.0 (for B-mod F322W2), I get an ABmag zeropoint of 26.470, which agrees with the zeropoint in the new zeropoint file. When I use the jdox eqn and a pixel area of 0.062"**2, I get an ABmag zeropoint of 26.510. So those seem to agree well. Translating between the new Fnu and Flambda values using the pivot wavelength also agree. 

This makes me think that the problem must be coming from either synphot (doing something strange with the new filter throughput files?), or Mirage's spectrum rescaling?

I compared the difference in the integral of the throughput curves. For F322W2 the integral increased by 10% from old to new curve. And the countrate calculated by synphot has also increased by 10%.

So that means the problem must lie in Mirage's rescaling?

bhilbert4:
I think I resolved the issue. Somehow the photfnu and photflam calculations were giving bad numbers. I'm not sure what I was doing earlier, but the numbers out of the calculation are different now. Note that a lot of confusion also came about because the AB zeropoint values were wrong independent of the photfnu issue. In this case, I had been simply using the AB zeropoint values that are on JDox. However, those zeropoints are defined as the AB magnitudes where the flux is 1 MJy/sr, while the definition that has been (and is still) used in Mirage is that the zeropoint corresponds to the AB magnitude where the signal is 1 DN/sec. This is all academic though, since Mirage only uses the photfnu and photflam values (as well as the vegamag zeropoint) in order to calculate the signal rate of a source in the simulated data.

bhilbert4:
@hover2pi can you give this a look-over? Lots of changed files, but the only code updates are some tests where truth values have been updated. I'm not sure how the environment updates were added to this PR. Those have already been merged as part of #808.

hover2pi (approved):
Fantastico!

bhilbert4:
Thanks!

Issue #813, opened 2022-11-29
jwst_gtvt failure?


bhilbert4:
It seems like something has changed in jwst_gtvt? This came from @dthatte yesterday. I tried running the jwst_gtvt function that the mirage function wraps around, and it crashed. The traceback for that is at the bottom of this ticket.

The notebook first crashes at
yaml_generator.default_obs_v3pa_on_date

I then comment out that line and manually provide pav3_obs3.

IndexError                                Traceback (most recent call last)
Cell In [17], line 10
      7 datatype = 'raw'
      9 obs3 = 3
---> 10 pav3_obs3 = yaml_generator.default_obs_v3pa_on_date(pointing_name, obs3, date = dates) 
     

File ~/anaconda3/envs/2022nov27/lib/python3.10/site-packages/mirage/yaml/yaml_generator.py:2304, in default_obs_v3pa_on_date(pointing_filename, obs_num, date, verbose, pointing_table)
   2280 """Find the nominal/default V3PA for an observation on a given date.
   2281 
   2282 Note, this relies on the JWST Generalized Target Visibility Tool (GTVT), and as such it
   (...)
   2300     V3PA in decimal degrees. Provide this to the `roll_angle` argument to mirage.yaml_generator()
   2301 """
   2303 if pointing_table is None:
-> 2304     pointing_table = apt_inputs.AptInput().get_pointing_info(pointing_filename, 0)
   2305 for i in range(len(pointing_table['obs_num'])):
   2306     if pointing_table['obs_num'][i] == f"{obs_num:03d}":

File ~/anaconda3/envs/2022nov27/lib/python3.10/site-packages/mirage/apt/apt_inputs.py:870, in AptInput.get_pointing_info(self, file, propid, skipped_obs_from_xml, verbose)
    856         self.logger.info('Skipping observation {} ({})'.format(obsnum, obslabel))
    858 try:
    859     # Skip the line at the beginning of each
    860     # visit that gives info on the target,
   (...)
    866     # The exception to this rule is TA images.
    867     # These have Exp values of 0. Sigh.
--> 870     if ((int(elements[1]) > 0) & ('NRC' in elements[4]
    871                                      or 'NIS' in elements[4]
    872                                      or 'FGS' in elements[4]
    873                                      or 'NRS' in elements[4]
    874                                      or 'MIR' in elements[4])
    875         ) or (('TA' in elements[4]) & ('NRC' in elements[4]
    876                                        or 'NIS' in elements[4])):
[...]

PR #815, opened 2022-12-05
Handle new Glossary in pointing file

bhilbert4:
Looks like the latest version of APT now adds a Glossary section to the pointing file. The current function for reading in the pointing file was finding the glossary section but not successfully ignoring it.

This PR updates the code to explicitly stop reading the pointing file when it reaches the Glossary line. It also protects against lines that are too short to contain pointing information.

### DIFF
--- a/mirage/apt/apt_inputs.py
+++ b/mirage/apt/apt_inputs.py
@@ -833,6 +833,10 @@ class AptInput:
                 elif (len(line) > 1):
                     elements = line.split()
 
+                    # If we hit the Glossary at the end, we can stop
+                    if elements[0] == 'Glossary':
+                        break
+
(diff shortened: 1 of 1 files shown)


bhilbert4:
With this fix in place, I was able to successfully generate the yaml files for the data in #813.

bhilbert4:
Fixed a couple tests that were failing for unrelated reasons. The new version of jwst_backgrounds uses updated data, and the calculated background rates have changed enough that the tests were failing. So I updated the expected values based on what the new version returns.

hover2pi (approved):
I believe it was Thoreau who said, "Our code is frittered away by detail. Simplify, simplify."

PR #835, opened 2023-12-29
WebbPSF oversampling can be int or array

bhilbert4:
Small change that allows the oversampling factor in the WebbPSF gridded library to be an integer or an array. Local tests have shown this to be an integer, as expected. But in help desk ticket INC0196396, the oversampling factor was a numpy array with one value for each of the x and y dimensions. 

This PR updates catalog_seed_image.py to be able to accept either an integer or an array. Note that for this simple fix, when the oversampling is an array, Mirage assumes that the oversampling is the same in both dimensions. With a few more modifications, it appears that Mirage could support different factors in the two dimensions, but it's not clear that this is needed at the moment.


mfixstsci (approved):
Thanks @bhilbert4 looks good!

PR #837, opened 2024-07-28
Fix errors in source location and signal level for moving targets

bhilbert4:
This was more major surgery than anticipated.

These changes fix a jumpiness at the sub-pixel level, of moving targets as they progress across pixels.
It also fixes a problem in the source brightnesses of the moving targets.

With these changes, Mirage now evaluates the PSF gridded library at every frame and subframe source location. Previously, a single PSF from the library was constructed, and Mirage would subsample the array, shift it, and return to nominal pixel scale.


bhilbert4:
At the moment, the fix seems to work, and has been tested only for point sources. More work is needed for extended targets and galaxies/ellipsoids.

bhilbert4:
Extended targets and ellipsoid work will be pushed to a separate PR. I've added NotImplementedErrors here for those cases.

bhilbert4:
This is working well for my test cases (moving point sources only). Ready for review.

mirage/utils/utils.py
```diff
@@ -424,6 +424,22 @@ def expand_environment_variable(variable_name, offline=False):
     return variable_directory
 
 
+def flatten_nested_list(nested_list):
```
mfixstsci, line 427: Any reason you could convert the nested list to a numpy array and use `.ravel()`?

mirage/config/NIRCam_zeropoints.list
```diff
@@ -1,18 +1,18 @@
 Filter Pupil Module Detector VEGAMAG ABMAG STMAG PHOTFLAM PHOTFNU Pivot_wave
-F070W CLEAR A NRCA1 26.16 26.45614182774024 27.001201986651644 5.748379794034092e-20 9.49663e-31 0.704
-F090W CLEAR A NRCA1 26.28 26.809122393581703 27.88994321167786 2.5354144673386703e-20 6.86084e-31 0.901
-F115W CLEAR A NRCA1 26.02 26.82114778745016 28.439373694749563 1.5285398688609284e-20 6.785270000000001e-31 1.154
-F140M CLEAR A NRCA1 25.11 26.236886799687156 28.280919201856925 1.7687178878418194e-20 1.162175e-30 1.404
-F150W2 CLEAR A NRCA1 27.23 28.495253771445945 30.917332884113748 1.5598868435783303e-21 1.45186e-31 1.671
-F150W CLEAR A NRCA1 25.89 27.11953817864898 29.308638503066177 6.863898418154277e-21 5.15479e-31 1.501
+F070W CLEAR A NRCA1 26.18 26.41421553992525 26.959275698836656 5.974698153409092e-20 9.870520000000001e-31 0.704
```
mfixstsci, line 2: This looks fine, I am curious do we keep this anywhere online? A rest API would be useful for something like this.

mirage/seed_image/catalog_seed_image.py
```diff
@@ -1426,17 +1429,16 @@ def movingTargetInputs(self, filename, input_type, MT_tracking=False,
         # Determine the name of the column to use for source magnitudes
         mag_column = self.select_magnitude_column(mtlist, filename)
 
-        # If any ephemeris file will be used, get the calendar dates associated
-        # with each frame
-        if 'ephemeris_file' in mtlist.colnames or non_sidereal_ra_interp_function is not None:
-            ob_time = '{}T{}'.format(self.params['Output']['date_obs'], self.params['Output']['time_obs'])
+        # Get the calendar dates associated with each frame
+        #if 'ephemeris_file' in mtlist.colnames or non_sidereal_ra_interp_function is not None:
```
mfixstsci, line 1433: Remove these lines if you aren't using them anymore.

mfixstsci (approved):
There is a lot here, I will bounce it back to you approved @bhilbert4 with a couple of comments. From what I am seeing there are improvements to the API and the documentation and I see updates to match changing APIs from dependencies. So with that, I would say over all this looks good, assuming you have done the necessary testing!

mirage/config/NIRCam_zeropoints.list
bhilbert4: I don't think this is kept anywhere else online. It's related to the pipeline's flux calibration, but it uses different units.

mirage/utils/utils.py
bhilbert4: There's no guarantee that all the sub-lists within the list have the same length, so converting to a numpy array doesn't work.
e.g. the nested list could look like: [[1,2,,3],[4,5],[6],[7,8,9]]

PR #839, opened 2024-09-18
Update environments 

bhilbert4:
Update dependency versions to support a 3.11 environment


mfixstsci (approved):
LGTM. Thanks for the investigation on the GTVT. I will need to look into the NaN row, I think it's because there is a line the use the previous day to project forward. So, I guess the trouble is, there isn't a day before the first row.

PR #848, opened 2025-06-06
Tweaks to support dependency changes

bhilbert4:
Just a couple of small changes to support changes in table column names, as well as a couple of numpy.float uses that were missed previously.


bhilbert4:
Also removed all calls to pkg_resources, which the warnings say may disappear completely later this year.

bhilbert4:
A bit of scope creep here. Dependency versions have been updated, including moving to support of python 3.11 and 3.12. pkg_resources has been removed (which accounts for most of the changed files here). Complications due to the batman package's latest version not supporting Intel-based Macs and also not working in python 3.12 resulted in new environment files as work-arounds. @mfixstsci would you have time at some point to give this a quick review?

mfixstsci:
@bhilbert4 I can give this a review no problem! I will get back to you later today.

mfixstsci (approved):
@bhilbert4 Looks good to me! Was thinking it might be easier to potentially create a constant for the `package path` and then import it so you only need to really make the repetitive change I was seeing once.

mfixstsci (approved):
Nice one, looks good @bhilbert4

mfixstsci (approved):
These look good too!

bhilbert4:
Thanks for the review @mfixstsci!!

PR #858, opened 2026-05-06
add build workflow that also builds frozen environment files and publishes to PyPI on a GitHub release

zacharyburnett:
also removes the conda environment files and `requirements.txt` in favor of generating them on each build and attaching them to each GitHub release as assets

PR #860, opened 2026-05-26
Remove requirements.txt

bhilbert4:
Remove requirements.txt, as mentioned in #858


bhilbert4:
@zacharyburnett is there anything else I should do here? Looks like tests are failing because the IPAC site is struggling today.

zacharyburnett:
that should be it!

